PR #11, opened 2022-03-04
Fix #7 and #9: set up FIMS testing frameworks

### LINKED ISSUE #7: Set up testing system on the FIMS repository
@k-doering-NOAA and I can work on 

- [x] setting up the testing system (e.g., include the files and folders needed) on the FIMS repository
- [x] getting the testing system running on GitHub actions
- [x] adding [C++ files](https://github.com/Bai-Li-NOAA/TestingExamples/blob/main/.gitignore) to the current .gitignore file
- [x] modifying the GH Actions workflow to save artifacts if a job fails

_Originally posted by @Bai-Li-NOAA in https://github.com/NOAA-FIMS/FIMS/issues/3#issuecomment-1049300604_

Bai-Li-NOAA:
Following changes have been made in this pull request:
- set up FIMS testing frameworks by adding tests folder (both R testthat and C++ gtest will be there)
- updated DESCRIPTION and add testthat section
- updated .gitignore file by adding more C++ files and CMake files
- updated .buildignore file to avoid running gtest when doing R CMD check
- added a GH Actions workflow to run google tests

Let me know if you would like me to change anything. Thanks!


.github/workflows/run-googletest.yml
```diff
@@ -0,0 +1,33 @@
+# Run C++ tests using GoogleTest
+name: run-googletest
+# on specifies the build triggers. See more info at https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows
+on:
+# The default build trigger is to run the action on every push and pull request, for any branch
+  push:
+  pull_request:
+  # To run the default repository branch weekly on sunday, uncomment the following 2 lines
+  #schedule:
+    #- cron: '0 0 * * 0'
```
k-doering-NOAA, line 10: It may be worth adding a CRON trigger, if FIMS has dependencies that we want to make sure are still compatible (e.g., tmb). This could be done later, as needed.

.github/workflows/run-googletest.yml
```diff
@@ -0,0 +1,33 @@
...
+      run: |
+        cmake -S . -B build -G Ninja
+        
+    - name: Build
+      run: |  
+        cmake --build build
+        
+    - name: Test
+      run: |
+        ctest --test-dir build
```
k-doering-NOAA, line 33: Will the job fail if the Test step have failing tests? We may also want to consider saving artifacts (perhaps conditionally if a job fails), if there is something useful to save (e.g., a detailed summary of which tests failed and why).

CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
...
+  CXX
+) # CXX is the language name
+
+# GoogleTest requires at least C++11
+set(CMAKE_CXX_STANDARD 11)
+
+include(FetchContent)
+FetchContent_Declare(
+  googletest
+  URL https://github.com/google/googletest/archive/refs/tags/release-1.11.0.zip
```
k-doering-NOAA, line 13: would we need to change this from time to time? I wonder if there is a way to just use the latest release? Or maybe fixing to a particular release is intentional?

k-doering-NOAA (approved):
Thanks for putting this together, @Bai-Li-NOAA! Looks good.

I was also able to run the tests locally by cloning the FIMS repo, checking out the feature branch and running from the command line:

```
cmake -S . -B build -G Ninja
cmake --build build
ctest --test-dir build
```

k-doering-NOAA:
It looks like someone with write access to this repo will also have to approve this PR before it can be merged, btw! (I only have read access).

.github/workflows/run-googletest.yml
Bai-Li-NOAA: Yes, the job will fail if there are failing tests. Good idea for saving artifacts. I added this suggestion as a new task in issue #7.

CMakeLists.txt
Bai-Li-NOAA: We will change it, but not very often. We will stick with particular releases when using CMake, Google Test, and Google Benchmark. When there is a need to use newer versions of dependencies, we will conduct regression testing before switching to new versions.

inst/include/interface/place_holder0.hpp
```diff
@@ -0,0 +1,49 @@
...
+ typedef typename std::vector<Type> parameter_vector;
+};
+
+template<typename T>
+T exp(const T& x){
+  return std::exp(x);
+}
+
+template <class T>
+const T log(const T& x){return std::log(x);}
```
ChristineStawitz-NOAA, line 45: Do we need to define this twice (lines 18 - 24 and lines 39 - 45 are identical to my eye)?

inst/include/interface/place_holder1.hpp
```diff
@@ -0,0 +1,17 @@
...
+template <class Type>
+/**
+ * Probability density function of the lognormal distribution.
+ *
+ * @param x A double value.
+ * @param meanlog Mean of the distribution on the log scale.
+ * @param sdlog Standard deviation of distribution on the log scale.
+ * @returns A density value.
+ */
+Type dlognorm(Type x, Type meanlog, Type sdlog)
```
ChristineStawitz-NOAA, line 12: is it possible to remove this and instead use the `dlognorm` defined in TMB?

tests/gtest/test_place_holder2.cpp
```diff
@@ -0,0 +1,126 @@
+#include "gtest/gtest.h"
+#include "place_holder2.hpp"
+
+// # R code that generates true values for the test
```
ChristineStawitz-NOAA, line 4: Can we remove the commented out code for now? If we want to include, should do so in a separate file.

ChristineStawitz-NOAA (approved):
Hi Bai - all of my comments on this are on the cpp side. I just spoke to @Andrea-Havron-NOAA about using a simple logistic function to hook up the `Rcpp` and `TMB` side. I think to avoid duplication we could try removing the logistic growth function here and apply your test case framework to the simpler logistic function @Andrea-Havron-NOAA is using? Or we could merge as-is and let her work on that in a separate pull request. I will approve for merging and leave it up to you.

inst/include/interface/place_holder0.hpp
Bai-Li-NOAA: Good question, I think the lines 18-24 and lines 39-45 are used for different libraries (TMB v.s. std). @Andrea-Havron-NOAA perhaps has a better explanation than me. I copied her FIMS-software [Rcpp code here](https://github.com/Andrea-Havron-NOAA/FIMS-software/blob/master/src/Rcpp/Common.hpp).

inst/include/interface/place_holder1.hpp
Bai-Li-NOAA: I am not sure if TMB has dlognorm function, I know there is a dnorm.hpp. Maybe the name of place_holder is confusing? I plan to remove all testing place_holder.hpp and place_holder.cpp files after the modeling team sets up one example of a module.

tests/gtest/test_place_holder2.cpp
Bai-Li-NOAA: Those commented out code are for generating one-time true values for C++ unit tests. I agree that for complicated tests we need to put the R script somewhere else, do you have a suggestion on where to put the R script? For simple C++ unit tests, commented out R code is part of the testing documentation. The benefits of including the simple R code as comments in the same C++ testing file is that you can remove or move C++ tests without worrying about where the R code is.

Bai-Li-NOAA:
Hi @ChristineStawitz-NOAA , Totally agree with the logistic function plan. All the testing place_holder files are independent and all of them can be removed once the modeling team sets up one module with some tests. The current testing place_holder files are used for quickly setting up the CMake system and GH actions workflow. @Andrea-Havron-NOAA can set up an independent logistic function without using any of the testing place_holder files. I think it is better to merge the pull request now and let @Andrea-Havron-NOAA work on a new branch (to develop the logistic function) simultaneously with @msupernaw (to modify the CMakeLists.txt).

PR #38, opened 2022-03-22
fix: move pr template

k-doering-NOAA:
The PR template is not automatically being generated when opening a PR. In the Stock Synthesis repo, I see that our PR template is working and is in a slightly different location, so I'm wondering if moving the PR template will solve the issue. We will have to merge this into main to test it out, though, as PR templates in branches are not used, only the one in main.


Andrea-Havron-NOAA (approved):
approving in order to test out

Issue #39, opened 2022-03-22
github actions: investigate build triggers


k-doering-NOAA:
**Is your feature request related to a problem? Please describe.**
We may be generating github action runs more often than necessary. For example, if a github actions runs on every push, we probably do not also need it to run when a PR is opened, because the same code is being tested twice.

**Describe the solution you'd like**
Limit github action runs to only when they are necessary.

**Describe alternatives you've considered**
We could just leave the build triggers as is, but the extra runs are unnecessary and may also cause confusion.

**Describe a reference describing the statistical validity of this approach**
n/a

**Describe if this is needed for a management application**
n/a

**Additional context**
I'm happy to investigate this, unless @Bai-Li-NOAA would like to!

ChristineStawitz-NOAA:
I think this is closed, but let's use the discussion on wednesday to reopen any needed issues on gh actions

PR #68, opened 2022-04-04
gha, #39: rm duplicative pull request triggers

k-doering-NOAA:
*What is the feature?*
See #39. Removes duplicative Github Actions triggers. In particular, if running the workflow on every push to every branch, there is no need to run the workflow upon opening a pull request.

*How have you implemented the solution?*
Remove the pull request build trigger

*Does it impact any other area of the project?*
No

*Does this change impact the model input or output?*
No

*How to test this change*
Note that there should not be the same jobs running on pull request and push for this pull request.


Bai-Li-NOAA (approved):
Looks great!

PR #71, opened 2022-04-05
chore: add relative path to the build folder

Andrea-Havron-NOAA:
*What is the feature?*
The cmake build folder is being tracked by git

*How have you implemented the solution?*
add relative path to the build folder in .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,7 +45,7 @@ vignettes/*.pdf
 *.dll
 
 # CMake build files
-/build
+/tests/gtest/build
 
 # C++ ignore files
 # Prerequisites


Bai-Li-NOAA:
Thanks @Andrea-Havron-NOAA for bringing this up. The generated build folder is supposed to be in the main directory after running `cmake -S . -B build -G Ninja` under FIMS folder. See test log folder structure [here](https://github.com/NOAA-FIMS/FIMS/blob/0831a3196a990c8c840b6ddce1a7bfc7feec7211/.github/workflows/run-googletest.yml#L40). I wonder if this happened because the cmake command was ran inside of tests folder locally. Happy to take a look at the issue through a code share session.

Andrea-Havron-NOAA:
Thanks for explaining this issue, @Bai-Li-NOAA. I will close out this PR.

Issue #103, opened 2022-05-04
[FEATURE] Double logistic selectivity


JaneSullivan-NOAA:
**Is your feature request related to a problem? Please describe.**
Implement double logistic selectivity as described in Li et al. 2020 model comparison project

**Describe the solution you'd like**

Meeting date:  May 25 8-10 am AKT

Workflow:
- [x] Requirements gathering:  track down equation in Li et al. operating model. Add this equation to collaborative workflow document
- [x] Implementation:  create new branch, add new files and modify FIMS math as needed
- [x] Documentation:  generate new doxygen and supporting code documentation; make sure these are rendering correctly by building in powershell (`cmake -S . -B build -G Ninja` and `cmake --build build --parallel 16` and opening FIMS/build/html/index.html)
- [x] Write tests:  Follow the simple logistic example
- [x] Test the tests?  Maybe pull in Bai for help with this! `ctest --test-dir build`
- [x] Integration:  make sure all the proper `#include` statements are used
- [x] Internal review:  Check equations, documentation, formatting, then submit PR

**Describe alternatives you've considered**
Our directive is to reproduce the model comparison project.

**Describe a reference describing the statistical validity of this approach**
Our subgroup may determine that there are more statistically robust or flexible methods to implementing selectivity. We will compile these findings in a separate Issue. 

**Describe if this is needed for a management application**
N/A

**Additional context**
N/A

ChristineStawitz-NOAA:
Removing triage_needed label since this is approved

JaneSullivan-NOAA:
@Bai-Li-NOAA it's not clear to me which parameterization of the double-logistic selectivity you used in the [comparison paper](https://spo.nmfs.noaa.gov/sites/default/files/pdf-content/fish-bull/11923li.pdf).  Do you mind helping us figure out which equation to implement (or, you could just "AMAK style!"). Thanks for your help!

PR #135, opened 2022-05-25
Add double logistic selectivity submodule #103 feat dbl logistic slx

JaneSullivan-NOAA:
Contacts: Jane Sullivan, Chris Legault, Jon Brodziak

# What is the feature?
Added double logistic selectivity function

Following the logistic selectivity example, we added the new double logistic equation to fims_math.hpp and created a new child class DoubleLogisticSelectivity which inherits from SelectivityBase. We modified files as needed to code proper #include statements and also added C++ tests.

No, it just adds new functionality

Will need new to add input/output functionality to include double logistic selectivity.

# How to test this change
We added new tests to the tests/gtest folder, all are passing.
# Developer pre-PR Checklist

# Please do these steps locally for big changes.  
We did all the following these steps locally. 


inst/include/common/fims_math.hpp
```diff
@@ -86,6 +86,24 @@ inline const T logistic(const T &median, const T &slope, const T &x) {
   return (1.0) / (1.0 + exp(-1.0 * slope * (x - median)));
 }
 
+/**
+ * @brief The general double logistic function
+ *
+ * \f$ \frac{1.0}{ 1.0 + exp(-1.0 * slope_{1} (x - median_{1}))} \left(1-\frac{1.0}{ 1.0 + exp(-1.0 * slope_{2} (x - median_{2}))}  \right)\f$
+ *
+ * @param median1 the median (inflection point) of the ascending limb of the double double logistic function
```
ChristineStawitz-NOAA, line 94: nit: I'm wondering if it would be more descriptive to name these args something like slope_asc and slope_desc instead of slope_1 and slope_2 so it's more clear to the user which is which (I get confused easily when numbering params).

inst/include/common/fims_math.hpp
```diff
@@ -86,6 +86,24 @@ inline const T logistic(const T &median, const T &slope, const T &x) {
...
+ * @param slope1 the slope of the aascending limb of e double hlogistic function
```
ChristineStawitz-NOAA, line 95: extra "a" in ascending

JonBrodziak: Issue was discussed and subgroup had strong preference but noted that
(i) if you use 1 and 2 then there is an implied ordering of smaller to larger
for the inflection point values, so median1 < median2 or median_asc < median_des
(ii) if one thought of adding general logistic function at some point with more
than 2 limbs then using a numbers would be a simpler labeling approach as in point (i)
(iii) it seems like adding a comment would help, something like, 
"input median1 < median2"

inst/include/common/fims_math.hpp
```diff
@@ -86,6 +86,24 @@ inline const T logistic(const T &median, const T &slope, const T &x) {
   return (1.0) / (1.0 + exp(-1.0 * slope * (x - median)));
 }
 
+/**
+ * @brief The general double logistic function
+ *
+ * \f$ \frac{1.0}{ 1.0 + exp(-1.0 * slope_{asc} (x - median_{asc}))} \left(1-\frac{1.0}{ 1.0 + exp(-1.0 * slope_{desc} (x - median_{desc}))}  \right)\f$
+ *
+ * @param median_asc the median (inflection point) of the ascending limb of the double logistic function
+ * @param slope_asc the slope of the ascending limb of e double logistic function
```
MOshima-PIFSC, line 95: small typo, change e to the ("of e double logistic function")

ChristineStawitz-NOAA (approved):

MOshima-PIFSC (commented):
Sorry I was a little slow with this but everything looks good!

Issue #137, opened 2022-05-25
[BUG] repeats of fims::exp() and fims::log()


JaneSullivan-NOAA:
The exp() and log() functions are repeated twice in [fims_math](https://github.com/NOAA-FIMS/FIMS/blob/9b23fa8ee695fdddd079935704a37437ac4435f1/inst/include/common/fims_math.hpp#L36). Probably just a typo... happy to fix it if you want to assign it to me.

ChristineStawitz-NOAA:
I don't think this is a bug actually..on line 21 the comment points out that the `IFDEF STD_LIB` and `IFDEF TMB_MODEL` are preprocessor directives that wrap each of the `exp` and `log` functions..so the first set of functions only gets run if the standard library is loaded and call `std:exp` while the second set only run if the model is being run within TMB and call the TMB exp function I believe. But, I am going to use this branch to do some other FIMS math cleanup

PR #138, opened 2022-05-25
update comment on fims_math

ChristineStawitz-NOAA:
This just updates the fims_math.hpp header.

I don't think we should get rid of the doubled exp and log functions but adding Andrea to confirm

### DIFF
--- a/inst/include/common/fims_math.hpp
+++ b/inst/include/common/fims_math.hpp
@@ -1,18 +1,9 @@
 /*
  * File:   fims_math.hpp
  *
- * Author: Matthew Supernaw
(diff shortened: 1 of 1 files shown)


JaneSullivan-NOAA (approved):

JaneSullivan-NOAA:
These changes and your feedback about the duplicate exp and log make sense to me, thanks!

Andrea-Havron-NOAA:
Christine's explanation in #137 is correct, fims::exp() will call std::exp() if the STD_LIB is defined and TMB's exp() if TMB_MODEL is defined.

PR #142, opened 2022-05-31
update calc coverage workflow to only run daily

ChristineStawitz-NOAA:
This changes the calc_coverage workflow to only run once a day so it doesn't give us failures whenever we push or commit


k-doering-NOAA:
@ChristineStawitz-NOAA , is this being merged into the right branch? Just wanted to check since it isn't main.

But you are right, running on every push and pull request seems excessive. Perhaps it should be running only on push or pull request to the main branch? The cron only runs on the main branch, so that also seems fine.

I did add a manual trigger, as it is nice to have an option to run manually sometimes.

ChristineStawitz-NOAA:
oops yes, changed to main!

PR #145, opened 2022-06-01
New maturity module with logistic function

JaneSullivan-NOAA:
# Contacts
Subgroup: @JonBrodziak, @cmlegault, @JaneSullivan-NOAA 

# What is the feature?
Added a new maturity module with the maturity base class and logistic maturity submodule. 

We followed the selectivity module structure.

No but we did make a note that we did not include the necessary `#include` statements because we were not sure which files to edit (e.g. information.hpp or model_object.hpp). This may need to be done in the future.

No, but our subgroup identified the need to add functionality to input maturity as a vector of proportions mature at age.

# How to test this change
We added gtests, passed github actions, and also checked that our changes did not disrupt any R tests

# Developer pre-PR Checklist

### DIFF
--- a/inst/include/population_dynamics/maturity/functors/logistic.hpp
+++ b/inst/include/population_dynamics/maturity/functors/logistic.hpp
@@ -1,0 +1,47 @@
+/*
+ * This File is part of the NOAA, National Marine Fisheries Service
+ * Fisheries Integrated Modeling System project.
+ * Refer to the LICENSE file for reuse information.
+ *
+ * The purpose of this file is to define the LogisticMaturity class, which 
+ * inherits from the MaturityBase class
+ */
+#ifndef POPULATION_DYNAMICS_MATURITY_LOGISTIC_HPP
+#define POPULATION_DYNAMICS_MATURITY_LOGISTIC_HPP
+
+//#include "../../../interface/interface.hpp"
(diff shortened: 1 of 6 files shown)


ChristineStawitz-NOAA (approved):
2. One potential issue that arises with the double logistic curve is that it can peak at a value lower than one, depending on the parameter values. This isn't necessarily a problem, but it does mean that we need to consider that when interpreting F. If we want selectivity curves to peak at one, the curve could be scaled to its max value. Identifiability of parameters can then become an issue, but that can be remedied by priors.

3. Will there be a mechanism to ensure that median_desc > median_asc? This could be accomplished by parameter bounding, but we would ideally want the lower bound of median_desc to depend on the value of median_asc, which may fluctuate during the optimization procedure.  One easy way to ensure the inequality holds is to set median_desc = median_asc + C, where C is the estimated parameter (along with median_asc) with a lower bound greater than zero. 

4. Why are two of the parameters referred to as medians in the naming? These are location parameters that define where the selectivity curve is equal to 0.5. They aren't likely to be median values of either the independent variable (e.g., age or size) or dependent variable (selectivity values). Unless I'm overlooking something, I would prefer names like "loc50_asc" and "loc50_desc" for the location parameters.

PR #168, opened 2022-06-09
107 feature implement empirical weight at age module and Rcpp interface

ChristineStawitz-NOAA:
Thanks for opening a pull request (PR) to FIMS! Please describe the changes under the headings below. Please also go through the checklist below to ensure the appropriate steps have been taken during development. For smaller changes, feel free to skip steps that aren't relevant, but at least describe the feature under the "what is the feature".

# What is the feature?
This implements an EWAA module for FIMS and exposes it in R using an Rcpp class. The EWAA structure is created as a map where the lookup key is ages and it returns weight. In a future module, we can expand this to a map of maps such that different growth patterns can be mapped across fleets, subpopulations, time periods, etc.

This is a new module under population_dynamics/growth and an associated Rcpp interface under interface/Rcpp. It is a bit different from other models in that it uses an empirical input rather than estimating parameters.

It affects compilation and test cases

Yes, this should now allow the model to take ewaa as input

# How to test this change
tests/
  compilation_tests/test_growth_module.cpp
  gtest/test_growth.cpp
testthat/FIMS_Rcpp_test.R

# Developer pre-PR Checklist


tests/gtest/test_growth.cpp
```diff
@@ -0,0 +1,59 @@
...
+      // not sure why this dummy test is here
+      EXPECT_EQ(1.0, 1.0);
+      
+      // create a new ewaa singleton class
+      fims::EWAAgrowth<double> ewaa1;
+      // set the ewaa values using an initializer list
+      // std::pair is a class template that provides a way to store two heterogeneous objects as a single unit
+      ewaa1.ewaa = 
+      std::map<double,double> {std::pair<double, double> (0.0,0.0),
+      std::pair<double, double> (1.0, 0.005306555),
```
cmlegault, line 24: The ewaa value should start 0.0005 instead of 0.005 as it does currently to match the comment on line 12. Currently growth decreases from age 1 to age 2. Since all the values were copy/pasted in the tests, everything passes even though the growth doesn't make sense biologically. Would be nice to fix the tests, but not really necessary.

cmlegault (approved):
Nice job with the EWAA module. It was easy to follow the logic of what was being done. The tests are all passing. I could not check the docker, and do not have any expertise with dockers, so am passing this aspect without any real review. I found one minor issue with the test values used. It is not necessary to change the tests, but it might help eliminate future confusion to do so.

Issue #209, opened 2022-08-02
[Bug]: Need to add FimsFrame documentation


ChristineStawitz-NOAA:
### Describe the bug

We need to update the `DESCRIPTION` and `NAMESPACE` fields of the package to work with FimsFrame

### To Reproduce

Run `devtools::check`
See the following warnings:
`Consider adding
  importFrom("methods", "new", "slot", "slotNames")
  importFrom("utils", "head")
to your NAMESPACE file (and ensure that your DESCRIPTION Imports field
contains 'methods').`

### Expected behavior

no warnings

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

k-doering-NOAA:
Looking at these warnings again, it looks like calling methods:: or utils:: in front of the function calls is one solution (the other one would be to use @import or @importFrom within the documentation for each functino that uses one of these function calls.

PR #214, opened 2022-08-11
fix, #209: changes recommended by r cmd check

k-doering-NOAA:
# What is the feature?
This reduces the number of notes/warnings found when running r cmd check (some noted in #209).

Most of these fixes had to do with making sure R could find the correct package that the function came from. For example, `slots()` is part of the {methods} package, so `methods::slots()` needed to be use in the codebase (an alternative is using `@importFrom methods slots` at the top of the function in which it is used.

N/a

No
# How to test this change
Look at the summary from the r cmd check github action. There should only be 3 notes now.

# Developer pre-PR Checklist

- [ ] Run [cmake build and ctest locally](https://noaa-fims.github.io/collaborative_workflow/testing.html#c-unit-testing-and-benchmarking) and make sure the C++ tests pass
- [ ] Run `styler::style_pkg()` locally and push changes to remote feature branch. If there are many changes, please do this in a separate commit.


k-doering-NOAA:
I think the run-doxygen was already failing on the main branch, hence why it is failing here. I also didn't touch c++ code in this PR, so I think it can be safely ignored.

kellijohnson-NOAA:
@k-doering-NOAA how come the r-cmd check is passing even though `utils` wasn't added DESCRIPTION?

k-doering-NOAA:
Is there a note about adding `utils` to DESCRIPTION @kellijohnson-NOAA ? I'm not seeing it...

R CMD Check only fails on errors, not warnings.

k-doering-NOAA:
Sorry, I see what you mean. I think because `utils` is a base package, you may not need to add it? I see this Stack overflow answer: https://stackoverflow.com/questions/39439006/package-dependencies-for-r-base-packages

But, if you think we should add it, I'm happy to do so!

kellijohnson-NOAA:
Thanks @k-doering-NOAA for finding the explanation on stack exchange. Now I understand why it isn't failing.

Andrea-Havron-NOAA:
I recently had a conversation with @JaneSullivan-NOAA and @jimianelli about using importsFrom vs. package::fn. Using importsFrom is preferred as it improves readability of code and is [recommended](https://r-pkgs.org/dependencies.html#import-r) if you are using a function more than once. It's better, though, to use package::fn if two packages use the same function name (eg. stats::filter and dplyr::filter).

As roxygen is just being used to write namespace and it's the namespace that determines which functions/packages are imported, the importFrom only needs to be written once for a function. So we would only need to write `importFrom methods new` above FIMSFrame and then that function is available everywhere, we wouldn't need to repeat `importFrom methods new` above FIMSFrameAge.

 I would recommend moving all methods functions to importFrom statements to clearn up the code. Also add library(methods) to the top of test files to avoid using methods:: each time a function is called.

k-doering-NOAA:
Thanks @Andrea-Havron-NOAA ! I'm happy to change that. I have tended to use the other style due to personal preferences of the people I have worked with, thinking it makes it more clear to see where the function comes from, but I can adapt.

kellijohnson-NOAA:
I too would have to adapt because I :heart: [Karl Broman's approach](https://kbroman.org/pkg_primer/pages/depends.html) with the explicit use of package::function. I like this notation over the use of importFrom because (1) your code doesn't have to change over time if you start using a function more, (2) new users know exactly where functions are coming from without having to look them up, and (3) you do not need to know if another package that you are importing has a function named the same.

If we are going to use importFrom, I suggest that we make a single file that declares all importFrom calls rather than intersperse them throughout the code, e.g.,
```
#' @importFrom pkg fun1 fun2
#' @importFrom pkg2 fun3
#' @importFrom pkg3 fun4
NULL
#> NULL
```
as suggested by [roxygen2](https://roxygen2.r-lib.org/articles/namespace.html).

Andrea-Havron-NOAA:
I recognize this might be my own preference and I am open to using :: instead of importFrom if that is preferred (although I still would argue against using :: for ggplot as code quickly becomes hard to read).

I did come across this advice in the [Google style guide](https://google.github.io/styleguide/Rguide.html#qualifying-namespaces):

> Functions from default R packages, including datasets, utils, grDevices, graphics, stats and methods. If needed, you can @import the full package.

So we can list utils and methods under the Imports field in DESCRIPTION, then we can use all functions without :: or importFrom.

k-doering-NOAA:
Thanks for looking to the google style guide, Andrea!
I think following the google guide make sense in this case, since that is the style guide we are following for FIMS- I can change my PR to match the advice.

k-doering-NOAA:
I took a look back at my PR and I think it is (accidentally) following the google style guide.

I'm a bit confused about the language regarding the exception regarding base packages, @Andrea-Havron-NOAA ; I think it is just saying that if you need to use many functions within these base packages, it's ok to use `@import methods` in the roxygen documentation on the top of the function rather than `methods::`? However, I'm not sure I've interpreted correctly.

On the main branch, methods is in `DESCRIPTION` and that we still get warnings from r-cmd-check if we don't also be explicit about the package name in the function (either through `@importFrom` or `methods::`; see https://github.com/NOAA-FIMS/FIMS/runs/7568287743?check_suite_focus=true#step:9:91 ). If that is the case, maybe I should leave the PR as is?

Andrea-Havron-NOAA:
Yes @k-doering-NOAA, that was my interpretation as well. And you are correct, listing methods under Imports in DESCRIPTION only loads the package - it doesn't attach it, sorry for the incorrect advice above! Per the Google style guide recommendation,  adding `@import methods` to roxygen doumentation will make all functions available from the package. Given we will be setting up a lot of S4 objects, I anticipate methods will be called a lot.

k-doering-NOAA:
@Andrea-Havron-NOAA I changed this PR to match the style outlined in #217 (it would not be too difficult to change it back if needed). Feel free to review again when you have time!

k-doering-NOAA:
@Andrea-Havron-NOAA , I changed the PR to use `ggplot2::` instead of `@import ggplot2`

Andrea-Havron-NOAA (approved):

PR #357, opened 2023-05-16
Document how to fix fatal error of file too big

kellijohnson-NOAA:
# What is the feature?

Provide formal documentation for how to fix the error that a file is too big.

Edited the README file on the main page.

@Andrea-Havron-NOAA please feel free to make any changes that you want to this branch during your "review" before you merge it in.


Andrea-Havron-NOAA:
@kellijohnson-NOAA, I've added more details on this error and how to compile in debugger mode. Can you review to make sure the descriptions I've added are easy to understand?

README.md
```diff
@@ -28,6 +28,26 @@ remotes::install_github("NOAA-FIMS/FIMS")
 library(FIMS)
 ```
 
+#### Fixing Fatal Error
```
kellijohnson-NOAA, line 31: @Andrea-Havron-NOAA do you think we should move this to the "Windows" section below line 22 given it is a windows-specific problem?

README.md
```diff
@@ -28,6 +28,26 @@ remotes::install_github("NOAA-FIMS/FIMS")
...
+
+Windows users can expect to see some derivative of the following error message in their R session if they have not yet set some flags using {withr}.
+```
+Fatal error: can't write <xxx> bytes to section .text of FIMS.o: 'file too big
+```
+You can easily fix this by running
```
kellijohnson-NOAA, line 37: we should add "the following line in your local R session. Note that this call will need to be repeated each time you open a new session." Or, at least that is what it sounded like needed to happen when you were giving instructions to Cole.

README.md
```diff
@@ -28,6 +28,26 @@ remotes::install_github("NOAA-FIMS/FIMS")
...
+```
+withr::local_options(pkg.build_extra_flags = FALSE)
+```
+This fix removes the debugger flag `-O0 -g` from being automatically inserted for certain devtools calls (eg. `devtools::load_all()`). Windows developers wanting to compile FIMS with the debugger turned on will need to run the above script in addition to manually modifying the [Makevars.win](https://github.com/NOAA-FIMS/FIMS/blob/doc-install/src/Makevars.win) file in the src directory to the following:
```
kellijohnson-NOAA, line 41: * "eg." should be "e.g.,"
* add "call to PKG_CXXFLAGS in the" after "modifying the"

kellijohnson-NOAA (commented):
I added some comments. Two main things are
1. explaining that the withr call needs to constantly be done each time you open a new R session
2. removing the 2 in front of "PKG_CXXFLAGS ="

src/.gitignore
```diff
@@ -1,3 +1,4 @@
 *.o
 *.so
 *.dll
+*.win
```
kellijohnson-NOAA, line 4: I do not think this actually works because the file has already been indexed. I think you have to use
```
git update-index --skip-worktree src/Makevars.win
```
and then eventually if you want users to be able to make changes to it, then use
```
git update-index --no-skip-worktree path/to/file src/Makevars.win
```
See [automationpanda](https://automationpanda.com/2018/09/19/ignoring-files-with-git/#:~:text=Like%20skip%2Dworktree%2C%20it%20makes,will%20not%20change%20the%20file.) as well as many stack overflow conversations about skip-worktree and assume-unchanged, though automationpanda better explains why you use one or the other.

README.md
Andrea-Havron-NOAA: Agreed, it fits better under the Windows section

src/.gitignore
Andrea-Havron-NOAA: I think I will remove the *.win from the .gitignore for now. Bai and I are discussing ways this workflow can be automated so developers won't need to change the Makevars.win file.

Andrea-Havron-NOAA (approved):

PR #358, opened 2023-05-16
Modifies the PR template

kellijohnson-NOAA:
# What is the feature?
This PR modifies the PR template to make it more visually pleasing after submitting. Using html comments, which are hidden in the resulting email and submitted PR allows users to know what they need to do without cluttering up the actual submitted PR. It is often unclear what we can and cannot delete in the current form. Using comments the instructions automatically go away instead of relying on the user to delete them as I did here in this PR.

I also modified some of the sections, for example, I combined the "Does it impact any other area of the project with Does this change impact the model input or output?" because technically that is another area of the project. I suggest to users to use sub headings for each area that is affected.

I used html comments in the .md file.

PRs will be less verbose.

NO

# How to test this change
I have tested using commented .md files for PR in other repositories.

# Developer pre-PR Checklist
I did nothing in the pre-PR checklist.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -38,7 +38,7 @@ changes.
 ## Input
 Text describing changes to the input.
 
-## Ouput
+## Output
 Text describing changes to the output.
 -->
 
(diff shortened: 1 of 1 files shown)


k-doering-NOAA (approved):
@kellijohnson-NOAA thank you for taking the time to do this! This is long overdue, and the template is greatly improved :)

PR #370, opened 2023-06-07
Merge in architecture vignette to main

ChristineStawitz-NOAA:
# What is the feature?

A vignette describing the FIMS architecture

no

# How to test this change

# Developer pre-PR checklist


vignettes/debug.txt
k-doering-NOAA, line 1: rm this file?

k-doering-NOAA, line 1: rm this file?

vignettes/fims-path-maturity.Rmd
```diff
@@ -0,0 +1,549 @@
...
+
+```{r, include = FALSE}
+knitr::opts_chunk$set(
+  collapse = TRUE,
+  comment = "#>"
+)
+```
+This vignette describes the hierarchical structure of FIMS by describing the linkages, or *path* through the FIMS model using maturity as an example. The vignette walks through how input from R is carried into the core FIMS calculations. 
+
+## Setting up a module in R and linking to the Rcpp interface
```
k-doering-NOAA, line 18: I got a little lost between code that you have to write yourself (in R in a script) to create the maturity module and then the code that is already written as part of FIMS. I suggest putting everything that a user has to write in a single section up top (including how to clear FIMS), and then adding a clear heading to signal transition to talking about the code that is within FIMS and the user does not have to write.

vignettes/fims-path-maturity.Rmd
```diff
@@ -0,0 +1,549 @@
...
+
+A namespace in C++ is similar to what a library is in R.
+
+The core of FIMS is within the `fims` namespace and is a convenient way to differentiate between the part of the C++ code base that is portable, i.e. independent of statistical platform, and the part of the codebase that depends on platforms outside of base C++ (e.g. `{Rcpp}`, R, `{TMB}`). Any code written within:
+```{Rcpp, eval = FALSE}
+namespace fims{
+  
+}
+```
+is considered to be a part of the `fims` namespace. Classes written within the `fims` namespace can be accessed using `fims::`.
```
k-doering-NOAA, line 135: I think fims:: is accessed using C++ code, not R code?

vignettes/fims-path-maturity.Rmd
```diff
@@ -0,0 +1,549 @@
...
+
+```{Rcpp, eval = FALSE}
+ for (population_iterator it = this->populations.begin();
+             it != this->populations.end(); ++it) {
+          std::shared_ptr<fims::Population<T> > p = (*it).second;
+          ...
+```
+In this code chunk, `(*it)` refers to a single population in the `populations` map. The second element of the map is the pointer, so `p = (*it).second` means we are setting the pointer to a single population to equal the pointer to `populations` in the map. 
+
+Within this population loop, we then link the `maturity` pointer in `population` to equal the `maturity` pointer in information, passing on the information that we want to use the `LogisticMaturity` class:
```
k-doering-NOAA, line 404: I think the chunk of code needs a comment that shows which is the maturity pointer in information and which is the maturity pointer in population, I wasn't sure.

vignettes/fims-path-maturity.Rmd
```diff
@@ -0,0 +1,549 @@
...
+```
+
+Here, (*it) is referring to the `maturity_models` pointer.
+
+```{r image-path-maturity-8, echo = FALSE, message=FALSE, out.width = '100%'}
+knitr::include_graphics("figures/fims-path-maturity-8.png")
+```
+
+
+## Conceptualizing this path with R
```
k-doering-NOAA, line 427: I like the idea, but this didn't work for me (I felt like I spent all that time trying to figure out the path in C++ and then had to start over to understand it in R), but I also don't always find analogies helpful. It might still be helpful for others!

vignettes/fims-path-maturity.Rmd
```diff
@@ -0,0 +1,549 @@
...
+                 maturity = "MaturityBase"),
+  "Pop 3" = list(id = 3,
+                 maturity = "MaturityBase")) 
+maturity_modules <- list(
+  "Logistic Maturity" = list(id = 1, median = 10, slope = 0.2), 
+  "Some Other Maturity" = list(id = 2, median = 8, slope = 0.3)
+)
+```
+
+We can think of population_modules and maturity_modules above as initiated C++ classes stored in our computer's memory. This R code chunk above is comparable to the following from FIMS, which sets up modules in R:
```
k-doering-NOAA, line 450: should initiated be "instantiated"?

vignettes/fims-path-maturity.Rmd
```diff
@@ -0,0 +1,549 @@
...
+}
+
+sapply(population_modules, function(x) x[2]$maturity)
+```
+
+In R, the above code sets up an intermediate value, p. Once we update the maturity from p to equal the correct maturity module, `p$pointer$maturity <- information$maturity[[newit]][[2]]`, the above code passes this back to population_modules, `population_modules[[it]][2]$maturity <- p$pointer$maturity`. In this context, `population_modules` is the population class held in memory that is being used to run calculations.
+
+The deeper level of abstraction happening in C++ is that in FIMS, _p is the population module_ stored in memory. An equivalent line of code to `population_modules[[it]][2]$maturity <- p$pointer$maturity` isn't needed in FIMS because when we update p, we are also updating the population class held in memory that is being used to run calculations.
+
+## Clearing FIMS
```
k-doering-NOAA, line 542: put this up with the other R code that the user has to write so it is all in one place.

vignettes/fims-path-maturity.Rmd
```diff
@@ -0,0 +1,549 @@
...
+  %\VignetteEncoding{UTF-8}
+---
+
+```{r, include = FALSE}
+knitr::opts_chunk$set(
+  collapse = TRUE,
+  comment = "#>"
+)
+```
+This vignette describes the hierarchical structure of FIMS by describing the linkages, or *path* through the FIMS model using maturity as an example. The vignette walks through how input from R is carried into the core FIMS calculations.
```
k-doering-NOAA, line 16: is "path" a technical concept in programming? I assumed so, but wasn't sure. Maybe if it is, a link would be helpful?

vignettes/fims-path-maturity.Rmd
```diff
@@ -0,0 +1,549 @@
...
+ */
+d0->maturity_models[lm0->id] = lm0;
+```
+
+
+The `add_to_fims_tmb()` function repeats code four times to track the estimated value of parameters along with their first, second, and third derivatives. 
+
+
+```{r image-path-maturity-2, echo = FALSE, message=FALSE, out.width = '100%'}
+knitr::include_graphics("figures/fims-path-maturity-2.png")
```
k-doering-NOAA, line 221: Thank you for making all these graphics! The visuals are so helpful.

vignettes/fims-path-maturity.Rmd
k-doering-NOAA, line 16: Is this vignette intended for developers only? Or would we imagine future users who are not also developers would need to understand this as well? Perhaps a sentence highlighting who the audience is would be helpful.

k-doering-NOAA (approved):
@Andrea-Havron-NOAA thanks for putting this together! This looks like a TON of work and I'm so grateful to have it.

I pushed two commits (one with minor formatting changes, one removing some unneeded library files and adding them to .gitignore), and included some suggestions. Nothing critical, though, so up to you if you want to make changes before merging to main or at a future time.

One other organizational idea: it could be helpful to have a "big picture" section before diving in the details, Perhaps with the fully built image of the FIMS path. I felt myself wondering how long we were going to keep building the up the image, and knowing up front how it looked would have helped my brain to know how far along we were in the vignette!

Bai-Li-NOAA (approved):

vignettes/fims-path-maturity.Rmd
```diff
@@ -0,0 +1,549 @@
...
+
+```{Rcpp, eval = FALSE}
+// file: inst/include/population_dynamics/population/population.hpp
+// maturity is a shared pointer to MaturityBase
+int maturity_id = -999; /*!< id of maturity model object*/
+std::shared_ptr<fims::MaturityBase<Type>>
+      maturity; /*!< shared pointer to maturity module */
+```
+
+The `maturity_id` was set from R using the `SetMaturity()` method from the `PopulationInterface` class: `fims$population$SetMaturity(fims$maturity$get_id())`
```
timjmiller, line 278: I noticed that this fims$population$SetMaturity(fims$maturity$get_id()) is not technically what "we" do earlier in the vignette on line 44. Maybe it does the same thing though.

vignettes/fims-path-maturity.Rmd
```diff
@@ -0,0 +1,549 @@
...
+
+Notice above that we set up the shared pointer to `MaturityBase` (parent class), not the specific maturity class we are using in this example, `LogisticMaturity` (child class). `MaturityBase` has an `evaluate` method with input arguments that match the inputs of each child class:
```
timjmiller, line 280: Not sure if this is right. At the beginning of the vignette 
maturity <- new(fims$LogisticMaturity)
makes maturity specifically logistic.

vignettes/fims-path-maturity.Rmd
```diff
@@ -0,0 +1,549 @@
...
+
+In R, the above code sets up an intermediate value, p. Once we update the maturity from p to equal the correct maturity module, `p$pointer$maturity <- information$maturity[[newit]][[2]]`, the above code passes this back to population_modules, `population_modules[[it]][2]$maturity <- p$pointer$maturity`. In this context, `population_modules` is the population class held in memory that is being used to run calculations.
+
+The deeper level of abstraction happening in C++ is that in FIMS, _p is the population module_ stored in memory. An equivalent line of code to `population_modules[[it]][2]$maturity <- p$pointer$maturity` isn't needed in FIMS because when we update p, we are also updating the population class held in memory that is being used to run calculations.
+
+## Clearing FIMS
+
+Because FIMS sets up objects stored in memory, simply running `rm(fims)`, won't free this memory back up. There is a function that can be run to clear out memory:
+```{r, eval = FALSE}
+fims$clear()
```
timjmiller, line 546: when I run this line fims (and other things) still exists in my session. perhaps this is expected, but maybe some more description on why that is expected and ok would be good.

vignettes/fims-path-maturity.Rmd
Andrea-Havron-NOAA: Updated to: C++ classes written within the `fims` namespace can be accessed within the C++ code base using `fims::`

vignettes/debug.txt
Andrea-Havron-NOAA: removed the file

vignettes/fims-path-maturity.Rmd
Andrea-Havron-NOAA: Thanks for the feedback! I added the following to teh intro of this section to help clarify it is optional and not required to understand the FIMS architecture!

This is an optional exercise for those who benefit from thinking through the structure in C++ code by seeing its corollary in R.

vignettes/fims-path-maturity.Rmd
Andrea-Havron-NOAA: yes, thanks!

vignettes/fims-path-maturity.Rmd
Andrea-Havron-NOAA: I added: _The vignette is tailored to FIMS developers or others interested in understanding the FIMS core architecture._

vignettes/fims-path-maturity.Rmd
Andrea-Havron-NOAA: fixed - this should be the same as line 44

vignettes/fims-path-maturity.Rmd
Andrea-Havron-NOAA: Agreed, this was not worded well. I updated to,

_Notice above that population by default declares a pointer of type `MaturityBase`_

vignettes/fims-path-maturity.Rmd
Andrea-Havron-NOAA: I added a comment that rm(fims), rm(population) and rm(maturity) also need to be run in R to remove these objects from the R session

vignettes/fims-path-maturity.Rmd
Andrea-Havron-NOAA: I attempted to color code the code chunks but think this might be too much. I  would prefer to show you the colored version and get your feedback before pushing up that change!

vignettes/fims-path-maturity.Rmd
timjmiller: maybe clear could be defined to do the rm commands on appropriate objects as well?

vignettes/fims-path-maturity.Rmd
Andrea-Havron-NOAA: This is a somewhat self created term on my part, but I have heard it before within the context of cyclomatic complexity, which is defined as the  number of independent paths through a program. I can't seem to locate any helpful links.

vignettes/fims-path-maturity.Rmd
Andrea-Havron-NOAA: clear is specifically a C++ function, so we wouldn't be able to call any R functions from inside it. We could create a wrapper than calls fims$clear and rm.

kellijohnson-NOAA: How would we know what R objects to clear though because I don't think it is a requirement that users name things `fims`, `population`, and `maturity`. It would be much cleaner to develop things in an environment and then just remove the whole environment.

vignettes/fims-path-maturity.Rmd
```diff
@@ -0,0 +1,614 @@
...
+   * @brief Calculates the maturity.
+   * @param x The independent variable in the maturity function (e.g., logistic
+   * maturity at age or size).
+   */
+  virtual const T evaluate(const T& x) = 0;
+};
+```
+
+The evaluate function is a [virtual C++ function](https://www.geeksforgeeks.org/virtual-function-cpp/), which means the function can be overwritten by functions of the same name from a child class. This structure keeps the maturity module in population generic. We don't need any conditional statements to loop over all possible maturity formulations within population. We only need to add a new child maturity class with an Rcpp interface and we can automatically call it from population. There is a trade-off here. We are creating a nested hierarchical structure that makes the code base harder to read. In exchange, we're creating code with a lower [cyclomatic complexity](https://en.wikipedia.org/wiki/Cyclomatic_complexity), that is there will only ever be one independent path from the user to the maturity->evaluate() call in population, regardless of how many maturity functions are added to FIMS. Code with lower cyclomatic complexity is easier to test, maintain, and extend.
```
Andrea-Havron-NOAA, line 372: @nathanvaughan-NOAA, @k-doering-NOAA, @Bai-Li-NOAA, does this paragraph capture what we discussed about virtual functions?

k-doering-NOAA: This looks great! I just checked out the link and found the example and text there really effective. Thanks, Andrea!

Andrea-Havron-NOAA:
@msupernaw, @Bai-Li-NOAA, @ChristineStawitz-NOAA,  the fims demo is working but reporting NANs in the sdreport standard errors.

Bai-Li-NOAA:
> the fims demo is working but reporting NANs in the sdreport standard errors.

I have fixed the input values for survey index and M vector. The demo is reporting standard errors now. See changes made [here](https://github.com/NOAA-FIMS/FIMS/pull/370/commits/943d68906043151d6ab6c83e86bd900cbd2a04ce).

I think we need to create two more issues to really fix the problems.
1) update data_mile1 to use biomass index for survey, instead of abundance index
2) add error checking for size of input vector (in R side?). The current model runs without printing any warnings/errors when the size of M input is wrong.

Andrea-Havron-NOAA:
@Bai-Li-NOAA, thanks for fixing this! I agree with your comments on additional issues. Should these be filed for this vignette or as issues for the next phase of R interface development? Especially wrt error checking, I think there are already plans to improve the error checking within the S4 framework.

Bai-Li-NOAA:
@Andrea-Havron-NOAA, Separate issues for the next phase of R interface development probably work!

DESCRIPTION
```diff
@@ -67,7 +67,11 @@ Roxygen: list(markdown = TRUE)
 RoxygenNote: 7.2.3
 Suggests: 
     covr,
-    testthat (>= 3.0.0)
+    testthat (>= 3.0.0),
+    remotes,
```
kellijohnson-NOAA, line 71: {remotes} will eventually be taken over by {pak} although the data has yet to be announced.

vignettes/fims-demo.Rmd
```diff
@@ -0,0 +1,432 @@
+---
+title: "Introducing the Fisheries Integrated Modeling System (FIMS)"
+output: github_document
+---
```
kellijohnson-NOAA, line 4: There is no engine listed so I get the following error when running `devtools::build_vignette()`
```
Warning message:
In tools::buildVignettes(...) :
  Files named as vignettes but with no recognized vignette engine:
   'vignettes/fims-demo.Rmd'
(Is a VignetteBuilder field missing?)
```
suggested code stolen from fims-path-maturity.Rmd
```
vignette: >
  %\VignetteIndexEntry{FIMS Path: Maturity}
  %\VignetteEngine{knitr::rmarkdown}
  %\VignetteEncoding{UTF-8}
```

vignettes/fims-path-maturity.Rmd
```diff
@@ -0,0 +1,614 @@
...
+```
+
+
+<style type="text/css">
+code span.co{
+  color: #8c8c8c; 
+}
+</style>
+
+This vignette describes the hierarchical structure of FIMS by describing the linkages, or *path* through the FIMS model using maturity as an example. The vignette is tailored to FIMS developers or others interested in understanding the FIMS core architecture. The vignette walks through how input from R is carried into the core FIMS calculations.
```
kellijohnson-NOAA, line 41: I am not sure what is meant by "input from R"

vignettes/fims-path-maturity.Rmd
```diff
@@ -0,0 +1,614 @@
...
+
+## Setting up a module in R
+
+We'll start from R. The first step is to install FIMS, retrieve the FIMS module, create a new maturity module inside FIMS, and populate the maturity module with parameter values. 
+
+```{r, class.source = "rchunk", eval = FALSE}
+# install and load FIMS
+remotes::install_github("NOAA-FIMS/FIMS@integration-test-estimation-R-clear-FIMS")
+library(FIMS)
+# Create a new maturity model
```
kellijohnson-NOAA, line 58: I suggest splitting out this chunk into two separate chunks so users that want to run the vignette interactively do not have to install the package if they already have it installed when they create the `fims` object or initiate a maturity module.

vignettes/fims-path-maturity.Rmd
```diff
@@ -0,0 +1,614 @@
...
+fims <- Rcpp::Module("fims", PACKAGE = "FIMS")
+maturity <- new(fims$LogisticMaturity)
+# Populate the maturity module with parameter values.
+maturity$median$value <- 10
```
kellijohnson-NOAA, line 62: thought: in the future it might be good to not name parameters the same name as major functions in R.

vignettes/fims-path-maturity.Rmd
```diff
@@ -0,0 +1,614 @@
...
+rm(fims)
+rm(maturity)
+rm(population)
+```
+Also restarting or closing out your R session will free up memory.
+
+## The Rcpp Interface
+
+The fields specific to the maturity module that are accessible from R are defined in the 
+[rcpp_interface.hpp](https://github.com/NOAA-FIMS/FIMS/blob/main/inst/include/interface/rcpp/rcpp_interface.hpp) file under [inst/include/interface/rcpp](https://github.com/NOAA-FIMS/FIMS/tree/main/inst/include/interface/rcpp):
```
kellijohnson-NOAA, line 93: suggestion: change "under <path>" to "in the <path> directory" or something along those lines because in the rendered vignette it is not clear upon first read that the path to the rcpp directory is a folder path. Perhaps I should also just pay better attention when I am reading.

vignettes/fims-path-maturity.Rmd
```diff
@@ -0,0 +1,614 @@
...
+```{Rcpp, class.source = "rcppchunk", eval = FALSE}
+Rcpp::class_<LogisticMaturityInterface>("LogisticMaturity")
+      .constructor()
+      .field("median", &LogisticMaturityInterface::median)
+      .field("slope", &LogisticMaturityInterface::slope)
+      .method("get_id", &LogisticMaturityInterface::get_id)
+      .method("evaluate", &LogisticMaturityInterface::evaluate);
+```
+
+The code above shows that there are two maturity module-specific fields, `median` and `slope`, and two maturity module-specific methods (i.e. functions), `get_id` and `evaluate`. You can access the fields and methods from R using the `show(fims)` function.
```
kellijohnson-NOAA, line 104: quibble: All instances of "i.e." and "e.g." should be followed by a comma because we are using American English rather than British English. See [jakubmarian](https://jakubmarian.com/comma-after-i-e-and-e-g/) for a more detailed explanation.

suggestion: Add "(i.e., parameters)" after "fields" to match the 👍text👍 explaining "method".

vignettes/fims-path-maturity.Rmd
```diff
@@ -0,0 +1,614 @@
...
+
+The maturity module's `get_id` method returns a unique ID for this specific module.
+
+The maturity module is called from R using the name in quotes from the `Rcpp::class` function, `LogisticMaturity` and references the maturity rcpp class, `LogisticMaturityInterface` defined in [rcpp_maturity.hpp](https://github.com/NOAA-FIMS/FIMS/blob/main/inst/include/interface/rcpp/rcpp_objects/rcpp_maturity.hpp) under [inst/include/interface/rcpp/rcpp_objects](https://github.com/NOAA-FIMS/FIMS/tree/main/inst/include/interface/rcpp/rcpp_objects).
```
kellijohnson-NOAA, line 108: issue: I do not understand the following text "maturity module is called from R using the name in quotes from the `Rcpp::class` function" and I think users could benefit from some more hand holding.

vignettes/fims-path-maturity.Rmd
```diff
@@ -0,0 +1,614 @@
...
+
+
+Within the rcpp_maturity.hpp file, we can link to the `fims::Information` class in order to register maturity parameters and specify whether or not they are random effects:
+
+```{Rcpp, class.source = "rcppchunk", eval = FALSE}
+//file: rcpp_maturity.hpp
+
+//d0 is a shared pointer that points to fims::Information
```
kellijohnson-NOAA, line 220: suggestion: `d0` is not described in the text but it is used 8 times in the code within the is vignette. I think readers would benefit from a bit of a road map regarding what d0 is and how you chose to use the name lm0 and how the latter name is specific to logistic maturity and if you were to implement a new module you would name that something different, e.g., sl0 for logistic selectivity.

vignettes/fims-path-maturity.Rmd
```diff
@@ -0,0 +1,614 @@
...
+/*
+ the maturity_models pointer from Information that matches the 
+ id of the fims::LogisticMaturity class is equal to the pointer 
+ to fims::LogisticMaturity 
+ */
+d0->maturity_models[lm0->id] = lm0;
+```
+
+
+The `add_to_fims_tmb()` function repeats code four times to track the estimated value of parameters along with their first, second, and third derivatives.
```
kellijohnson-NOAA, line 271: suggestion: can you format "add_to_fims_tmb()" to make it orange or whatever color the Rcpp interface is in the diagram? At first glance, I thought it was an R function exposed to the user, then I looked in the diagram and saw that it was orange. 😮‍💨 and felt kind of dumb.

vignettes/figures/fims-path-maturity-2.png
kellijohnson-NOAA, line 1: question: How come the legend has "fims:Information" instead of "fims::Information" like the box in the figure?
bug? :bug: : should the green and gray boxes be referenced with upper case FIMS within the box and within the legend? I think having them as uppercase points readers to the actual code base rather than the `fims` object that is being created in R.

vignettes/fims-path-maturity.Rmd
kellijohnson-NOAA: +1 This graphic deserves a 🏆 and helped me wrap my brain around a lot of missing pieces. The color coding of the parts and the ⬇️s were really helpful. I wonder though if the color of all the arrows should be orange?

vignettes/figures/fims-path-maturity-2.png
kellijohnson-NOAA: question: should there be a closing bracket in the orange box after `d0`?

vignettes/fims-path-maturity.Rmd
```diff
@@ -0,0 +1,614 @@
...
+}  // namespace fims
+```
+
+```{r image-path-maturity-3, echo = FALSE, message=FALSE, out.width = '100%'}
+knitr::include_graphics("figures/fims-path-maturity-3.png")
+```
+
+## Calling maturity from population
+
+The `Population` class defined in [population.hpp](https://github.com/NOAA-FIMS/FIMS/blob/main/inst/include/population_dynamics/population/population.hpp) under [inst/include/population_dynamics/population](https://github.com/NOAA-FIMS/FIMS/blob/main/inst/include/population_dynamics/population) is where all the biological calculations happen, producing expected values used in likelihood equations and derived values important for management (e.g. SSB).
```
kellijohnson-NOAA, line 320: fix: change "SSB" to "spawning biomass" because SSB is not defined in this vignette.

vignettes/fims-path-maturity.Rmd
```diff
@@ -0,0 +1,614 @@
...
+
+```{r image-path-maturity-7, echo = FALSE, message=FALSE, out.width = '100%'}
+knitr::include_graphics("figures/fims-path-maturity-7.png")
+```
+
+We then loop through the populations and create a new shared pointer, `p`, to reference the individual population of interest.  
+
+```{Rcpp,  class.source = "infochunk", eval = FALSE}
+//file: inst/include/common/information.hpp
+ for (population_iterator it = this->populations.begin();
```
kellijohnson-NOAA, line 464: fix: I think @Andrea-Havron-NOAA mentioned that we are no longer using `it` because it can be negative?

vignettes/fims-path-maturity.Rmd
Andrea-Havron-NOAA: changed to: how user provided input from R is carried into the core FIMS calculations.

vignettes/fims-path-maturity.Rmd
```diff
@@ -0,0 +1,614 @@
...
+      .constructor()
+      .field("median", &LogisticMaturityInterface::median)
+      .field("slope", &LogisticMaturityInterface::slope)
+      .method("get_id", &LogisticMaturityInterface::get_id)
+      .method("evaluate", &LogisticMaturityInterface::evaluate);
+```
+
+The code above shows that there are two maturity module-specific fields, `median` and `slope`, and two maturity module-specific methods (i.e. functions), `get_id` and `evaluate`. You can access the fields and methods from R using the `show(fims)` function.
+
+The maturity module's `get_id` method returns a unique ID for this specific module.
```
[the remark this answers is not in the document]
Andrea-Havron-NOAA: Added: Defined methods are functions that can be called from R.

vignettes/fims-path-maturity.Rmd
Andrea-Havron-NOAA: Added:

When creating a new module in R, users can access the module using the name in quotes from the `Rcpp::class`, `LogisticMaturity`

```{r, class.source = "rchunk", eval = FALSE}
maturity <- new(fims$LogisticMaturity)
```

This module references the maturity rcpp class, `LogisticMaturityInterface` ...

vignettes/fims-path-maturity.Rmd
```diff
@@ -0,0 +1,614 @@
...
+```
+
+All Rcpp interface classes from FIMS define parameters (e.g. `median`, `slope`) using the `Parameter` class defined in [rcpp_interface_base.hpp](https://github.com/NOAA-FIMS/FIMS/blob/main/inst/include/interface/rcpp/rcpp_objects/rcpp_interface_base.hpp) under [inst/include/interface/rcpp/rcpp_objects](https://github.com/NOAA-FIMS/FIMS/blob/main/inst/include/interface/rcpp/rcpp_objects):
+
+```{Rcpp, class.source = "rcppchunk", eval = FALSE}
+class Parameter {
+ public:
+  double value; /**< initial value of the parameter*/
+  double min =
+      std::numeric_limits<double>::min(); /**< min value of the parameter*/
```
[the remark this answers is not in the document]
Andrea-Havron-NOAA: min and max default to -large number and +large number, but it has been recommended to remove these from the interface given that they are not actually used in FIMS. I will update the vignette once this change has been made. 

For now I have added the following comments:
//std::numeric_limits<double>::min() returns a very large negative number
//std::numeric_limits<double>::max() returns a very large positive number

vignettes/fims-path-maturity.Rmd
Andrea-Havron-NOAA: I expanded the paragraph above this code chunk to read:

Within the rcpp_maturity.hpp file, we can link to the `fims::Information` class in order to register maturity parameters and specify whether or not they are random effects. We do this by setting up two pointers in the interface: d0 which points to information, and lm0, which points to the logistic maturity module. 

I am reluctant to go into details on the naming convention as it doesn't always make sense will likely change soon.

vignettes/fims-path-maturity.Rmd
Andrea-Havron-NOAA: We are no longer using `int` in for loops because `int` can be negative. [Issue #339](https://github.com/NOAA-FIMS/FIMS/issues/379) is about the use of iterators (`it`) , we may simplify maps with iterators to vectors that can be referenced by their index directly.

vignettes/fims-path-maturity.Rmd
Andrea-Havron-NOAA: I increased the line size, thanks for the suggestion!

vignettes/figures/fims-path-maturity-2.png
Andrea-Havron-NOAA: fixed the bugs, thanks! I will update `fims` in the R code to `Fims` or `FIMS`  as the fims namespace in C++ is all lower case per C++ best practice.

vignettes/fims-path-maturity.Rmd
```diff
@@ -0,0 +1,614 @@
+---
+title: "FIMS Path: Maturity"
```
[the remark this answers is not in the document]
Andrea-Havron-NOAA: Thanks for the feedback!

vignettes/fims-path-maturity.Rmd
Andrea-Havron-NOAA: Created issue #412

PR #421, opened 2023-07-13
format C++ code with clang format

ChristineStawitz-NOAA:
Auto-generated by [run-clang-format.yml][1]
  [1]: https://github.com/NOAA-FIMS/FIMS/blob/main/.github/workflows/run-clang-format.yml

### DIFF
--- a/inst/include/population_dynamics/population/population.hpp
+++ b/inst/include/population_dynamics/population/population.hpp
@@ -53,7 +53,7 @@ struct Population : public FIMSObject<Type> {
   ParameterVector log_M; /*!< estimated parameter: log Natural Mortality*/
 
   // Transformed values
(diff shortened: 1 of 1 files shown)


Rick-Methot-NOAA:
Note that the SS3 project has a tool to format code.  Perhaps you have already used it in developing this code.

k-doering-NOAA:
> Note that the SS3 project has a tool to format code. Perhaps you have already used it in developing this code.

@Rick-Methot-NOAA yes, it is the same tool. For FIMS, it actually is run with every change to the cpp source code thanks to GitHub actions (and the fact that we are using pure c++ instead of an admb tpl file)

Andrea-Havron-NOAA (approved):

PR #430, opened 2023-08-02
resolved issues causing warnings

MOshima-PIFSC:
# What is the feature?

* changed int to size_t in population.hpp and FIMS.cpp
* updated constructor in data_object.hpp
* changed size_t to int in rcpp.population

# How to test this change

Run devtools::load_all() to verify there are no compiler warnings.

# Developer pre-PR checklist

### DIFF
--- a/inst/include/common/data_object.hpp
+++ b/inst/include/common/data_object.hpp
@@ -55,7 +55,7 @@ struct DataObject : public fims::FIMSObject<Type> {
   /**
    * Constructs a one-dimensional data object.
    */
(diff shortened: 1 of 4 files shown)


Bai-Li-NOAA (approved):
Looks great! Thanks for getting rid of the warnings. 

Do we need to add a GHA workflow (not in this pull request) and run `devtools::load_all()`? The GHA workflow will fail if there are warnings in the output. I am thinking about something like this:
```r
output <- utils::capture.output(devtools::load_all())
base::grep("warning", output, value = TRUE)
```

Issue #242, opened 2022-10-04
[Bug]: fix code coverage reporting


k-doering-NOAA:
### Describe the bug

Reports of code coverage are being uploaded by 2 separate job, 1 or the code coverage of g test and another for the codecoverage of test that. This leads to two different reports being pushed up to code cov, and thus not reliable code coverage values.

originally found in https://github.com/NOAA-FIMS/FIMS/issues/182

### To Reproduce

See https://app.codecov.io/gh/NOAA-FIMS/FIMS

### Expected behavior

A consistent single value for code coverage that includes coverage from c++ and r tests. Alternatively, following only 1 test.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Bai-Li-NOAA:
Thanks @k-doering-NOAA for creating this issue. I'm not sure what the solution would be either. It is hard to say one framework is more important than the other because it would be great if developers know if they need to write more C++ tests and R tests before merging in their code.

I think the C++ test coverage is sent to codecov every time we run run-googletest.yml while R test coverage is sent to codecov 5:30 after we make a pull request to the main branch. That may be why we get two code coverage values. I cannot remember why the R testthat coverage is scheduled at 5:30. Do you think we can modify the two workflows and make them submit the two reports simultaneously when we make a push/pull request to the main branch? It also helps developers figure out whether to write more tests before merging a pull request.

k-doering-NOAA:
That seems like a good goal, for developers to see what their test that/ c++ code coverage is. I think if we are making and sending 2 reports, the stats will still look weird on codecov.io, but maybe what is most important for now is that developers know the code cov calcs for their PR and how those compare to the main branch.

k-doering-NOAA:
Currently the r test that coverage is turned off and we push the gtest coverage reports up to codecov.io. @Andrea-Havron-NOAA suggested that maybe we could use another service besides codecov.io to push one of our reports too, since calculating a joint code coverage is tricky.

Andrea-Havron-NOAA:
For example, [coveralls](https://coveralls.io/) could be used to calculate C++ test coverage and codecov.io for R test coverage

kellijohnson-NOAA:
I found a few articles on how to combine the results of the code coverage from the R tests and the google tests into a single metric. We just have to combine running of the tests into a single action.
### references
* [code coverage :octocat: action](https://github.com/codecov/codecov-action)
[...]

PR #436, opened 2023-08-11
#242 fix codecov calc and upload

k-doering-NOAA:
# What is the feature?
Fix calculating and pushing code coverage reports to codecov.io. Addresses #242 

* Turns on the call-calc-coverage.yml workflow that uses covr::codecov()
* This calculates the coverage for both R and C++ code
* Remove the code coverage step from run-googletest.yml, since it was not doing what it was supposed to.
* Note I'm not sure if the build triggers are what we want or not - right now, it runs on every push and manually, but maybe only running to every push on main and on pull requests to main (and manually) are necessary?

Not directly

# How to test this change
- Verify that only 1 report is uploaded: https://app.codecov.io/github/NOAA-FIMS/FIMS/commit/6715b979ceeb361ed5960df210ac912cf28dd562 
- See that the codecov includes both R and C++ code (yay!): https://app.codecov.io/github/NOAA-FIMS/FIMS/tree/242-fix-codecov-rebased

# Developer pre-PR checklist

### DIFF
--- a/.github/workflows/call-calc-coverage.yml
+++ b/.github/workflows/call-calc-coverage.yml
@@ -4,6 +4,7 @@ name: call-calc_coverage
 # The default is to run the workflow on every push or pull request to every branch.
 on:
   workflow_dispatch:
+  push:
   #schedule:
   #  # * is a special character in YAML so you have to quote this string
   #  - cron:  '30 5 * * *'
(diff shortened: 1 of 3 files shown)


Bai-Li-NOAA (approved):
Looks great! I agree with your suggestions. I think only running to every push on main and on pull requests to main (and manually) are necessary.

k-doering-NOAA:
Thanks @Bai-Li-NOAA ! I added the build triggers, and also excluded some paths when coverage should not be expected to change (e.g.,changes only to the README).

I also realized it might be helpful to have codecov add a comment to our pull requests with information about coverage, so I turned on the option.

PR #445, opened 2023-08-15
Update run-clang-format.yml

msupernaw:
use llvm rather than google

# What is the feature?

* 

# How to test this change

# Developer pre-PR checklist


k-doering-NOAA:
@msupernaw I ran the GitHub action with your change, this is what happens to the code: https://github.com/NOAA-FIMS/FIMS/pull/447/files

Does it fix your issue?

PR #466, opened 2023-09-29
377 refactor call report f from modelhpp instead of fleet and population

Andrea-Havron-NOAA:
# What is the feature?

* Moved reporting from fleet.hpp and population.hpp to model.hpp

* Used nested vectors so reported objects from different fleets or populations do not need to be the same dimension
* Created a new function in inst/include/interface/interface.hpp to collapse the nested vector back to a single vector for ADREPORT

- Output from obj$report() is now formatted as a list with each element of the list being a single fleet or population

# How to test this change

- test-fims-estimation.R

# Developer pre-PR checklist

- [x] Ran [cmake build and ctest locally](https://noaa-fims.github.io/collaborativ/e_workflow/testing.html#c-unit-testing-and-benchmarking) and the C++ tests passed -->

### DIFF
--- a/inst/include/common/model.hpp
+++ b/inst/include/common/model.hpp
@@ -84,6 +84,7 @@ class Model {  // may need singleton
     vector< vector<Type> > ssb(n_pops);
     vector< vector<Type> > biomass(n_pops);
     vector< vector<Type> > rec_dev(n_pops);
+    vector< vector<Type> > recruitment(n_pops);
     vector< vector<Type> > M(n_pops);
     #endif
 
@@ -131,6 +132,7 @@ class Model {  // may need singleton
         naa(pop_idx) = vector<Type>((*it).second->numbers_at_age);
         ssb(pop_idx) = vector<Type>((*it).second->spawning_biomass);
         rec_dev(pop_idx) = vector<Type>((*it).second->recruitment->recruit_deviations);
+        recruitment(pop_idx) = vector<Type>((*it).second->expected_recruitment);
         biomass(pop_idx) = vector<Type>((*it).second->biomass);
         M(pop_idx) = vector<Type>((*it).second->M);
       #endif
@@ -162,14 +164,11 @@ class Model {  // may need singleton
     REPORT_F(index_nll, of);
     REPORT_F(jnll, of);
     REPORT_F(naa, of);
-    ADREPORT_F(naa, of);
     REPORT_F(ssb, of);
-    ADREPORT_F(ssb, of);
     REPORT_F(rec_dev, of);
-    ADREPORT_F(rec_dev, of);
+    REPORT_F(recruitment, of);
     REPORT_F(biomass, of);
     REPORT_F(M, of);
-    ADREPORT_F(M, of);
     REPORT_F(exp_index, of);
     REPORT_F(exp_catch, of);
     REPORT_F(F_mort, of);
--- a/inst/include/common/model.hpp
+++ b/inst/include/common/model.hpp
@@ -26,48 +26,48 @@ namespace fims {
 /**
  * @brief Model class. FIMS objective function.
  */
-template <typename T>
+template <typename Type>
 class Model {  // may need singleton
  public:
-  static std::shared_ptr<Model<T> >
+  static std::shared_ptr<Model<Type> >
       fims_model; /*!< Create a shared fims_model as a pointer to Model*/
-  std::shared_ptr<Information<T> >
+  std::shared_ptr<Information<Type> >
       fims_information; /*!< Create a shared fims_information as a pointer to
                            Information*/
 
 #ifdef TMB_MODEL
-  ::objective_function<T> *of;
+  ::objective_function<Type> *of;
 #endif
 
   // constructor
   virtual ~Model() {}
 
   /**
-   * Returns a single Information object for type T.
+   * Returns a single Information object for type Type.
    *
-   * @return singleton for type T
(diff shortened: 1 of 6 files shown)


inst/include/common/model.hpp
```diff
@@ -86,34 +109,101 @@ class Model {  // may need singleton
...
+    REPORT_F(ssb, of);
+    REPORT_F(rec_dev, of);
+    REPORT_F(recruitment, of);
+    REPORT_F(biomass, of);
+    REPORT_F(M, of);
+    REPORT_F(exp_index, of);
+    REPORT_F(exp_catch, of);
+    REPORT_F(F_mort, of);
+    REPORT_F(cnaa, of);
+    REPORT_F(cwaa, of);
```
Bai-Li-NOAA, line 176: Should we report log_rzero and selectivity parameters (e.g., median and slope) since these parameters are estimated in the estimation test?

tests/testthat/test-fims-estimation.R
```diff
@@ -167,7 +167,7 @@ test_that("deterministic test of fims", {
   sdr <- TMB::sdreport(obj)
   sdr_fixed <- summary(sdr, "fixed")
 
-  report <- obj$report()
+  report <- obj$report(obj$par)
```
Bai-Li-NOAA, line 170: Perhaps include documentation here as a reminder for developers to provide the parameter list to obj$report.

tests/testthat/test-fims-estimation.R
```diff
@@ -389,103 +390,134 @@ test_that("estimation test of fims", {
     control = list(maxit = 1000000, reltol = 1e-15)
   ))
 
-  report <- obj$report()
+  report <- obj$report(obj$env$last.par.best)
```
Bai-Li-NOAA, line 393: Perhaps include documentation here as a reminder for developers to provide `obj$env$last.par.best` to obj$report.

tests/testthat/test-fims-estimation.R
```diff
@@ -261,31 +261,29 @@ test_that("deterministic test of fims", {
   }
 
   # Expected survey index
-  fims_object <- report$expected_index[, 2]
-
-  cwaa <- matrix(report$cwaa[1:(om_input$nyr * om_input$nages), 2], nrow = om_input$nyr, byrow = T)
-  expect_equal(fims_object, apply(cwaa, 1, sum) * om_output$survey_q$survey1)
+  cwaa <- matrix(report$cwaa[[2]][1:(om_input$nyr * om_input$nages)], nrow = om_input$nyr, byrow = T)
+  expect_equal(fims_index[[2]], apply(cwaa, 1, sum) * om_output$survey_q$survey1)
```
Bai-Li-NOAA, line 265: I find the nested vectors from different fleets a bit confusing. Should we consider adding an additional layer of nesting for fleets? This way, `fims_index[[1]][[1]]` and `fims_index[[1]][[2]]` would correspond to indices 1 and 2 from population 1, related to `biomass[[1]]`, while `fims_index[[2]][[1]]` and `fims_index[[2]][[2]]` would correspond to indices 1 and 2 from population 2, related to `biomass[[2]]`. With the current setup, are we planning to continue expanding the fims_index IDs to accommodate more fleets for different populations (e.g., `fims_index[[3]]` and `fims_index[[4]]` for population 2)?

Bai-Li-NOAA (approved):
I'm happy to provide clarification for my questions if needed. Let me know whether these questions should be addressed in this pull request or not.

inst/include/common/model.hpp
ChristineStawitz-NOAA: We can access parameter estimates and standard errors via `summary(sdreport)` - these do need to be reported for testing. log_rzero is sometimes different than true value and it would be good to ID when it's rzero vs reference point calcs - we could report these by calling individual modules.

tests/testthat/test-fims-estimation.R
Andrea-Havron-NOAA: In the current set-up, fims_index[[1]] corresponds to the fleet module with an id == 1 (fishing_fleet in test_fims_estimation.R), that is:

`
fims_index[[1]] == fims_index[[fishing_fleet$get_id()]]
fims_index[[2]] == fims_index[[survey_fleet$get_id()]]
`

I was thinking it would be easier to keep track on the R side if we keep this 1:1 mapping of outputs and modules. We should probably figure out how we are going to connect populations and fleets when there are multiple populations, each with multiple, independent fleets.

The other way of reporting out derived variables from fleet is to report a matrix where the number of columns correspond to the number of fleets. The reason for using the ragged array approach (using nested vectors) is that it would allow the dimensions (e.g. numbers of years) to vary by fleet. The matrix approach will still work. If the dimensions do not match, the matrix dimensions can be set based on the largest dimension and missing values filled in with -999. These are treated as constants and dropped out of standard error calculations. 

I can change the code to use a matrix approach if this is easier to understand. I think an expansion that accounts for multiple populations, each with multiple fleets can be handled in a separate issue.

Bai-Li-NOAA: Thank you for the clarification! I think the ragged array approach works for the current version of FIMS. I agree that addressing the expansion to accommodate multiple populations, each with multiple fleets, can be a separate issue.

inst/include/common/model.hpp
Andrea-Havron-NOAA: Issue #482

PR #475, opened 2023-10-05
add unit tests for index and age comp nll evaluation

ChristineStawitz-NOAA:
# What is the feature?

* Removed age comp nll and index nll unit test since these are testing objects that no longer exist

delete .hpp, .cpp, and .R files associated with these tests

# How to test this change

# Developer pre-PR checklist


Andrea-Havron-NOAA:
I added tests on index_nll and age_comp_nll to the [test-fims-estimation.R](https://github.com/NOAA-FIMS/FIMS/blob/a77db895d063a1f7906a9ea8f86f054e824c02dc/tests/testthat/test-fims-estimation.R#L353)  as part of PR #377. Is this sufficient for unit testing fleet nll evaluations?

Andrea-Havron-NOAA:
After further conversation with @k-doering-NOAA and @Bai-Li-NOAA, we decided unit tests on index_nll and age_comp_nll should still be added to test-fleet-interface.R in addition to the tests on these functions in test-fims-estimation.R

ChristineStawitz-NOAA:
Unfortunately, after investigating for a bit, I don't think there's a straightforward way to unit test this. The expected value vectors for both index and age comps are only prepared when `MakeADFun` is called. We'd need to expose a way to initialize Fleets in Rcpp to call it separately, but I don't think that would be useful other than for testing purposes.

Issue #478, opened 2023-10-18
[Documentation]: doxygen documenation of namespaces


JaneSullivan-NOAA:
### Issue details

[Original issue](https://github.com/NOAA-FIMS/FIMS/issues/393)

[Related collaborative workflow issue](https://github.com/NOAA-FIMS/collaborative_workflow/issues/113)

As of 10/18/2023 we are still trying to figure out how to document namespaces in doxygen. We want something that looks like TMB's doxygen website, where namespaces are defined in the [namespace list](https://kaskr.github.io/adcomp/namespaces.html) and classes are defined as a member of each namespace in the [class list](https://kaskr.github.io/adcomp/annotated.html).

Andrea-Havron-NOAA:
[Example](https://github.com/kaskr/adcomp/blob/92ab18478aa4ec8a24aa8821960f835c48a038a2/TMB/inst/include/cppad/cppad.hpp#L18) from TMB where namespace documentation happens in a single file opposed to at each occurance of a namespace declaration:

```
\namespace CppAD
\brief contains all the variables and functions defined by the CppAD package.
*/
```

PR #479, opened 2023-10-18
Create multiple namespaces in FIMS

k-doering-NOAA:
# What is the feature?

* Address #393 
* Splits the fims namespace into several namespaces

Why use unique namespaces? The use of unique namespaces allows developers to quickly identify the source file for a namespace member class when used in a different part of the code and to understand how classes relate to one another throughout the codebase.

Note: we couldn't figure out the right way to document namespaces thoroughly in doxygen, so we opened a [new issue](https://github.com/NOAA-FIMS/FIMS/issues/478) for the future.

May cause merge conflicts with existing branches, because namespaces edits happen across many cpp files. 
Does not affect i/o on the R side.

# How to test this change

Make sure GHAs pass, especially google test.

# Developer pre-PR checklist


msupernaw (commented):
I feel like nested namespace would be more readable, for example:

fims::poppy rather than fims_popdy. This is more standard c++.

https://google.github.io/styleguide/cppguide.html#Namespace_Names

Bai-Li-NOAA:
Please following Matthew's recommendations when making changes to namespaces in FIMS. I don't have sufficient knowledge regarding namespaces design in C++. Here is a [link](https://learn.microsoft.com/en-us/cpp/cpp/namespaces-cpp?view=msvc-170) containing examples of namespaces that could be useful.

ChristineStawitz-NOAA:
@msupernaw are you willing to approve this without nested namespaces? It doesn't seem like there's a clear consensus of best practices. We should just check that everything is currently within a namespace.

msupernaw (approved):

PR #480, opened 2023-10-18
renamed dimension folded indices to be clearer

peterkuriyama-NOAA:
Co-authored-by: Kathryn Doering <kathryn.doering@noaa.gov>
Co-authored-by: Meg Oshima <MOshima-PIFSC@users.noreply.github.com>
Co-authored-by: Chris Legault <chris.legault@noaa.gov>

# What is the feature?

* Rename dimension folded indices to be clearer

# How have you implemented the solution?

Replaced index_ya and index_ya2 to be i_age_year and i_agem1_yearm1

# Does the PR impact any other area of the project?

No anticipated changes to input and output

# How to test this change

Not applicable, checked with github actions
# Developer pre-PR checklist

- [x] I relied on GitHub actions to :test_tube: things for me while I sat on the :couch_and_lamp:.

PR #485, opened 2023-10-19
Changed <T> to <Type>

peterkuriyama-NOAA:
# What is the feature?

* Changed <T> to <Type>. Also changed "T" to "Type" in declarations and definitions

See above

I don't think so

# How to test this change

Checked with devtools::load_all(); setup_and_run_gtest(); devtools::test()

# Developer pre-PR checklist


k-doering-NOAA:
@peterkuriyama-NOAA thanks for doing this! It looks like our work from earlier this week snuck into this branch (part of PR #480), so do you think it's ok that we cherry pick this branch and open a new PR? Perhaps our group can work on this in our session today, and then merge into main if everything looks good!

PR #489, opened 2023-10-24
Update GHA workflows to enforce FIMS installation

Bai-Li-NOAA:
# What is the feature?

Update GHA workflows to ensure successful checks on the main branch.

- Modify `run-googletest.yml` and `get-gtest-codecov.yml` to enforce FIMS installation.
- add PAT to env variable `GITHUB_PAT` in both run-googletest.yml and get-gtest-codecov.yml

# How to test this change

All GHA checks have passed, see details [here](https://github.com/NOAA-FIMS/FIMS/actions?query=branch%3Afix-GHA)

# Developer pre-PR checklist

### DIFF
--- a/.github/workflows/get-gtest-codecov.yml
+++ b/.github/workflows/get-gtest-codecov.yml
@@ -32,7 +32,7 @@ jobs:
       with:
         extra-packages: any::remotes
     - name: install local version of FIMS
-      run: remotes::install_local(upgrade = "always")
+      run: remotes::install_github("NOAA-FIMS/FIMS", upgrade = "always", force = TRUE)
       shell: Rscript {0}
     - name: setup for gtest
       run: FIMS:::setup_gtest()
--- a/.github/workflows/get-gtest-codecov.yml
+++ b/.github/workflows/get-gtest-codecov.yml
@@ -22,6 +22,8 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
+    env:
+      GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
 
     steps:
     - uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


.github/workflows/get-gtest-codecov.yml
```diff
@@ -32,7 +34,7 @@ jobs:
       with:
         extra-packages: any::remotes
     - name: install local version of FIMS
-      run: remotes::install_local(upgrade = "always")
+      run: remotes::install_github("NOAA-FIMS/FIMS", upgrade = "always", force = TRUE)
```
k-doering-NOAA, line 37: I think this always install the main branch, so the code coverage won't be accurate for branches, which I don't think is what we want?

k-doering-NOAA:
@Bai-Li-NOAA , thanks for working on this!

Hmm, it looks like these workflows now are only testing code on the main branch, though, which I don't think is what we want? Maybe we need to find a way to specify the branch in the install statement? (I also don't understand the differences between install_github and install_local, and why one works and one doesnt...)

I also wonder if getting rid of the install dependencies step would help; I don't think it is needed if using install_github()

I'm happy to look into options with you to get these working again!

k-doering-NOAA:
what about using `install_local(upgrade= "always", force = TRUE)`?

Bai-Li-NOAA:
@k-doering-NOAA I think `install_local(upgrade= "always", force = TRUE)` may solve the issue. I will give it a try and let you know!

.github/workflows/get-gtest-codecov.yml
Bai-Li-NOAA: Thanks @k-doering-NOAA , the issue was resolved using `install_local(upgrade= "always", force = TRUE)`.

k-doering-NOAA: fantastic!!!

k-doering-NOAA (approved):
Looks good now, thanks!

PR #498, opened 2023-10-31
fix: build trigger only on pr being opened

k-doering-NOAA:
# What is the feature?

* Fix build trigger on the pr checklist so it only runs once. It seems to be running when the PR is opened and additional commits are pushed, which leads to duplicate checklist

add `[opened]` so it only runs when the PR is opened rather than on other events 

# How to test this change

# Developer pre-PR checklist


Bai-Li-NOAA (approved):
Looks good!

PR #503, opened 2023-11-01
Name refactoring for tests, growth, and median

nathanvaughan-NOAA:
# What is the feature?

* Refactoring test names to remove _ and match google style recommendations
* Refactoring population to use weight_at_age[] vector instead of growth evaluate function
* Refactoring to replace the median parameter with inflection_point to be more descriptive and avoid overlap with R function names

We have refactored the code base to change names as described.

No this should not impact the rest of the code

# How to test this change

# Developer pre-PR checklist

### DIFF
--- a/inst/include/population_dynamics/population/population.hpp
+++ b/inst/include/population_dynamics/population/population.hpp
@@ -173,6 +173,7 @@ struct Population : public fims_model_object::FIMSObject<Type> {
 
     // Transformation Section
     for (size_t age = 0; age < this->nages; age++) {
+      this->weight_at_age[age] = growth->evaluate(ages[age]);
       for (size_t year = 0; year < this->nyears; year++) {
         size_t i_age_year = age * this->nyears + year;
         this->M[i_age_year] = fims_math::exp(this->log_M[i_age_year]);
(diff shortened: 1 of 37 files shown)


nathanvaughan-NOAA:
We know we are a couple of commits behind main but given the large number of open pull requests we were leaving another rebase until this is ready to be merged into main.

inst/include/population_dynamics/population/population.hpp
```diff
@@ -173,6 +173,7 @@ struct Population : public fims_model_object::FIMSObject<Type> {
 
     // Transformation Section
     for (size_t age = 0; age < this->nages; age++) {
+      this->weight_at_age[age] = growth->evaluate(ages[age]);
```
Bai-Li-NOAA, line 176: question[non-blocking]: do we want to use `this->weight_at_age[ages[age]]` or `this->weight_at_age[age]` throughout the codebase?

Bai-Li-NOAA (commented):
This looks great! Thanks for the name refactoring. I can help with merging these changes into the main branch once the conflicts are resolved.

Bai-Li-NOAA (approved):

inst/include/population_dynamics/population/population.hpp
nathanvaughan-NOAA: We discussed this when we were writing and concluded that this->weight_at_age[age] will always be the expected index value while this->weight_at_age[ages[age]] would crash if at some point we allow no integer ages. At the moment they are both equivalent though.

nathanvaughan-NOAA:
@ChristineStawitz-NOAA Looks like we already found a scenario where the merge block is an issue. This pull request says it cannot be rebased due to conflicts and it will now only let me choose Squash and merge which I don't want to do because these are 4 different issues that would be best kept separate for future interpretability.

ChristineStawitz-NOAA:
@nathanvaughan-NOAA I came to the same conclusion once I tried to approve :) I removed that setting

nathanvaughan-NOAA:
Hahaha sounds good, I'll merge once I fix these merge conflicts :)

PR #571, opened 2024-05-02
Rename testthat files

k-doering-NOAA:
# What is the feature?

* Rename the testthat files as proposed in https://github.com/NOAA-FIMS/collaborative_workflow/issues/63.

This includes: 
1) renaming/combining testthat files that correspond with testing R files match the convention `test-name_of_r_file.R` 
2) renaming test of rcpp interface to the convention `test-rcpp-description.R` (where `description` describes the interface being tested.
3) renaming integration tests (currently 1 file) to `test-integration-description.R` (where `description` describes the integration test).

# How to test this change

Make sure GitHub actions are passing and that filenames are as described

# Developer pre-PR checklist


tests/testthat/test-rcpp-get-parameter-vector.R
kellijohnson-NOAA, line 1: Should this one be renamed to test-rcpp-get_fixed.R?

tests/testthat/test-fimsframe.R
kellijohnson-NOAA, line 1: Should the file name be in the same case as the function name, i.e., test-FIMSFrame?

kellijohnson-NOAA (changes requested):
I made two comments about the file names but other than that everything looks good to merge this into main tomorrow.

kellijohnson-NOAA:
@k-doering-NOAA I made two comments on this PR that I am not entirely sure if they are valid or not. Once those are addressed we can merge it in.

tests/testthat/test-fimsframe.R
k-doering-NOAA: I think it should be the same as the filename that the function FIMSFrame is in, which is `fimsframe.R`. However, do you think we should modify the file names to be more consistent with the function name?

tests/testthat/test-rcpp-get-parameter-vector.R
k-doering-NOAA: Sorry, I can't follow why - but if that name feels more appropriate to what the test does, I think it can be renamed!

k-doering-NOAA:
@kellijohnson-NOAA sorry this took so long - I just replied to your comments. Feel free to decide based on my comments if the changes should be made and merge it in when ready (Or, I'm happy to make the changes if you prefer, just let me know what you think makes sense)!

tests/testthat/test-rcpp-get-parameter-vector.R
kellijohnson-NOAA: I just made the suggestion because it seems like we are testing a specific Rcpp function called get_fixed.

Somewhere in the FIMS code it says the following
```
get_fixed function is an Rcpp function that loops over all Rcpp
modules and returned a vector of parameters being estimated
```

k-doering-NOAA: Good catch! I just changed the name to test-rcpp-get_fixed.R.

kellijohnson-NOAA (approved):

PR #563, opened 2024-04-26
fix Codespace fails to build, add rstudio option

k-doering-NOAA:
# What is the feature?

* Fix codespaces build by including only pkgs that are dependencies
* Add Rstudio as an IDE option. To use, type `rserver` in the command line.more details: https://github.com/rocker-org/devcontainer-features/blob/main/src/rstudio-server/README.md

# How have you implemented the solution?

I think the inclusion of the cli and r lang packages was causing conflicts, which resulted in the codespace failing to build. I removed all packages that weren't dependencies of FIMS (and added some that are, but weren't listed) and was able to successfully build a codespace.

# Does the PR impact any other area of the project?

No

# How to test this change

Try building a codespace on this branch and see if it succeeds.

# Developer pre-PR checklist

- [x] I relied on GitHub actions to :test_tube: things for me while I sat on the :couch_and_lamp:.

PR #580, opened 2024-05-02
M2 r output issue 558 (partial, add skeleton output object/constructor)

k-doering-NOAA:
# What is the feature?

Partially addresses #558 by:
* Adds skeleton for the R output object
* Adds skeleton for the constructor function
* Adds skeleton for tests of the constructor function

Also
* Adds a fix to get codespaces to work and option to use R studio in the codespace

@Huihualee-NOAA and I worked on this today, but will not work again until next week, so it is probably best to rebase in our changes now and make a new branch next week. @kellijohnson-NOAA , let us know if you prefer something different, though.

no

# How to test this change

# Developer pre-PR checklist

### DIFF
--- a/R/output_object.R
+++ b/R/output_object.R
@@ -1,0 +1,8 @@
+#' Create FIMS Output Object
+#'
+#' @return The FIMS output object as a S4 object
+#' @examples
+#' create_fims_output()
+create_fims_output <- function() { 
+
+}
--- a/R/output_object.R
+++ b/R/output_object.R
@@ -1,8 +1,33 @@
 #' Create FIMS Output Object
 #'
-#' @return The FIMS output object as a S4 object
+#' @return The FIMS output S4 object of class FIMSOutput.
 #' @examples
 #' create_fims_output()
-create_fims_output <- function() { 
+create_fims_output <- function() {
 
 }
+
+# This file defines a parent class and its children by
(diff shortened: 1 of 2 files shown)


kellijohnson-NOAA:
This is great @k-doering-NOAA and @Huihualee-NOAA. I am wondering though if we should back out the container commits (there are two of them) and open up a separate PR to main with those changes. It seems like they would be good for those commits to be in main now rather than mixed in this branch. The workflow that I propose to get this done would require the following steps:
1. make a new branch off of this one labeled something like `allow-rstudio`
2. do a hard reset of `allow-rstudio` to only include the first two commits
3. rebase `allow-rstudio` with  main
4. open a pull request using `allow-rstudio` to main
5. Once (4) is in main then do the following
6. perform an interactive rebase on `m2-r-output-issue-558` and remove the first two commits that are now in `allow-rstudio`
7. force push `m2-r-output-issue-558` to `m2-r-output-issue-558`
8. keep working on `m2-r-output-issue-558` because there are no other commits to `m2-r-output` so merging today is not imperative

k-doering-NOAA:
@kellijohnson-NOAA, sure, I can remove the container commits and put them in #563 - I'm not sure how long it will take to get them in main, though, as they have been sitting for a while!

Issue #567, opened 2024-04-30
[Refactor]: R code does not adhere to style guide


kellijohnson-NOAA:
### Refactor request

The R code in the FIMS repository has code that deviates from the style guide recommended in the collaborative workflow even though the code is being styled by a styler. 

### Tasks
- [x] parameter and object names should be snake_case, e.g., `n_years` not `nyears` in fimsframe.R and test-fimsframe.R (@MOshima-PIFSC)
- [x] determine if the usethis package should be suggested or imported (test moving it to suggests in a branch)
- ~~look for additional rules to add to the styler to catch these types of deviations~~
- [x] open an issue in the case-studies repo about removing FIMSFrameAge from case studies
- [x] #564 
- [x] #590 
- [x] function names should align with the guide rather than be a mix of snake_case and camelCase; issue NOAA-FIMS/collaborative_workflow#133 will need to be decided upon before this task can more forward
- [x] roxygen @params should be sentences with subsequent lines indented with two spaces.
- [x] roxygen titles should be less than 80 but preferably 60 characters in sentence case without a full stop
- [x] roxygen even though I advocated for it, it has not officially accepted to use {} around package names
- [x] roxygen paragraph entries should start on a new line, e.g., @description, @source, @format, and @return should not have text on the same line
- [x] roxygen R/data_mile1.R#L6 can be wrapped to be less than 80 characters per line
- [x] #565
- [x] `use_template` and `ui_stop` are being imported in [R/FIMS-package.R](https://github.com/NOAA-FIMS/FIMS/blob/97281c20ba697c66cbbbf5fe4f01241f79183676/R/FIMS-package.R#L8) but they are being called with `::` so there is no need to import them
- [x] use `utils::head()` rather than importing it
- [x] Remove other unused functions should be removed from the code base

### Path forward

@MOshima-PIFSC and @kellijohnson-NOAA will work on this in the m2-r-cleanup branch, where tasks will be completed in sub-feature branches, e.g., m2-r-cleanup-zzz, that are merged into the feature branch daily. Then a pull request will be submitted after style-r is rebased to main. We can work on one file at a time rather than one task at a time, then move onto the tasks that cross multiple files. The timeline for completion is May 17th, 2024.

kellijohnson-NOAA:
[...]

PR #589, opened 2024-05-09
M2 r cleanup format

kellijohnson-NOAA:
# What is the feature?

* Formatted .R files that will be staying
* Reduced code by calling run_gtest() instead of duplicating code
* Part of #567 

## Input
@Bai-Li-NOAA I changed one line of the gtest code to use `paste()` instead of `paste0()`.

## Output
Using `paste()` allows for automatic spaces between the hard-coded command and any additional arguments the user supplies as well as spaces between arguments because a user might pass multiple strings.

# How to test this change

# Developer pre-PR checklist

### DIFF
--- a/R/data_mile1.R
+++ b/R/data_mile1.R
@@ -1,9 +1,9 @@
-#' {FIMS} input data frame for milestone 1
+#' FIMS input data frame for milestone 1
 #'
 #' A dataset containing information necessary to run an age-structured stock
-#' assessment model in {FIMS} for milestone 1. This data was generated using
-#' the `ASSAMC` package written for the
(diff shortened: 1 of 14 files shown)


MOshima-PIFSC (approved):
Changes look good to me

Bai-Li-NOAA:
Agree with Meg, the changes look good to me. I'm not sure whether the [call-doc-and-style-r workflow](https://github.com/NOAA-FIMS/FIMS/blob/main/.github/workflows/call-doc-and-style-r.yml) will overwrite the formatted .R files, but we can open an issue if it automatically style the .R files using a different style.

Bai-Li-NOAA (approved):

kellijohnson-NOAA:
Part of #567

Issue #605, opened 2024-05-28
[Developer Issue]: code coverage not getting calculated for pull requests


iantaylor-NOAA:
### Issue details

The code coverage of Pull Requests is calculated in order to confirm that coverage remains high after the change in the PR. However none of the PRs in the past month have had a successful calculation of the code coverage metric: https://app.codecov.io/gh/NOAA-FIMS/FIMS/pulls.

The message provided by codecov is

> Missing Head Commit
> Unable to compare commits because the head commit of the pull request is not found.
> [Learn more here](https://docs.codecov.com/docs/error-reference#section-missing-head-commit)

where that link says the error can be caused by "not yet uploaded coverage results to Codecov".

I'm confused about the combination of code coverage workflows for gtest and testthat (discussed in #242) so I'm not sure where to look for a solution.
Did something change?

k-doering-NOAA:
Thanks for reporting this @iantaylor-NOAA - I can look into this and see what's going on.

k-doering-NOAA:
It looks like codecov now requires an upload token in order to upload reports. See [this issue on the rlib actions](https://github.com/r-lib/actions/issues/834). What is happening is that the reports being generated for gtest and testthat are not being successfully uploaded to codecov.io due to the lack of an upload token.

I'm working on adding an upload token to the FIMS repo and then referencing it in the 2 workflows.

PR #611, opened 2024-05-29
Fix codecov (addresses #605)

k-doering-NOAA:
Note: this should only be merged after this [ghactions4r PR is merged](https://github.com/nmfs-fish-tools/ghactions4r/pull/124)

# What is the feature?

* Adds an upload token to the coverage workflows.

An upload token was generated and added to the FIMS org as a secret, then the YAMLs for the coverage workflows modified to include the token.

No

# How to test this change

Can test that the workflows run successfully once the ghactions4r PR is merged in.

# Developer pre-PR checklist

### DIFF
--- a/.github/workflows/call-calc-coverage.yml
+++ b/.github/workflows/call-calc-coverage.yml
@@ -22,3 +22,6 @@ on:
 jobs:
   call-workflow:
     uses: nmfs-fish-tools/ghactions4r/.github/workflows/calc-coverage.yml@main
+    secrets:
+      CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
+
--- a/.github/workflows/get-gtest-codecov.yml
+++ b/.github/workflows/get-gtest-codecov.yml
@@ -72,6 +72,7 @@ jobs:
       uses: codecov/codecov-action@v3
       with:
         files: coverage.xml
+        token: ${{ secrets.CODECOV_TOKEN }}
   
 
 


iantaylor-NOAA:
@k-doering-NOAA, great to see the code coverage calculated for a pull request once again!
I don't know why this worked without the ghactions4r branch yet being merged, but it's a good sign.

k-doering-NOAA:
@iantaylor-NOAA I think it worked because the report from `get-gtest-codecov.yml` successfully uploaded! The one for test that will not until the ghactions4r changes are merged in, though, so the stats are not completely accurate (it is only accounting for the gtest coverage and not the test that coverage)

.github/workflows/call-calc-coverage.yml
```diff
@@ -22,3 +22,6 @@ on:
 jobs:
   call-workflow:
     uses: nmfs-fish-tools/ghactions4r/.github/workflows/calc-coverage.yml@main
+    secrets:
```
Bai-Li-NOAA, line 25: Do we want to add brief documentation or a link to instructions on how to add an upload token as a secret? Additionally, maybe include a reminder here to update all workflows that depend on this workflow?

Bai-Li-NOAA (approved):
Looks good to me. Thank you, @k-doering-NOAA !

.github/workflows/call-calc-coverage.yml
kellijohnson-NOAA: I think a reminder about dependent workflows is a good idea. But, my preference would to not add documentation or instructions about uploading a token as a secret. There is plenty of help on the web for that with a quick google search. I am worried about broken future links or outdated instructions.

k-doering-NOAA: I think the code cov link is probably sufficient for how to add a secret: https://docs.codecov.com/docs/adding-the-codecov-token, but I think this can be added to ghactions4r instead of FIMS?

I don't think there are any dependent workflows on this one.

k-doering-NOAA:
> Thanks, @k-doering-NOAA, for the speedy fix! I am a bit confused by the Codecov report above. Do you know why it compares base ([97281c2](https://github.com/NOAA-FIMS/FIMS/commit/97281c20ba697c66cbbbf5fe4f01241f79183676)) to head ([656c262](https://github.com/NOAA-FIMS/FIMS/commit/656c26235f786fec5a768d7234d00fcb87273fd0)) and why the report is 11 commits behind head on main?

Sorry, I thought I had responded to this! My guess is that 11 commits ago was when the codecov report last successfully run on main, so codecov is just using the latest available info it has to make the comparison.

PR #615, opened 2024-05-31
fix rec_nll evaluate bug

Andrea-Havron-NOAA:
# What is the feature?

* Fixes bug related to the recruitment nll evaluation
* The value from the bool, `estimate_log_devs` was not being passed from the interface into the model
* The model bool was `true` by default so the recruitment nll was being evaluated regardless if the recruitment deviations were being estimated.

Passed the bool value from the interface to the model

This fix will change the overall joint negative log-likelihood and the recruitment negative log-likelihood reported from the model

# How to test this change

fims_demo.Rmd should print a value of `0` from the recruitment nll, `report$rec_nll`

# Developer pre-PR checklist


Andrea-Havron-NOAA:
@Bai-Li-NOAA or @ChristineStawitz-NOAA, can either of you review and approve this PR?

ChristineStawitz-NOAA (approved):

Issue #586, opened 2024-05-02
[Refactor]: Edit connections between fleet.hpp and distributions


Andrea-Havron-NOAA:
### Refactor request

- [x] Include new base and children class files in inst/include/common/distributions/distributions.hpp

- [x] Edit fleet functions: [evaluate_age_comp_nll](https://github.com/NOAA-FIMS/FIMS/blob/97281c20ba697c66cbbbf5fe4f01241f79183676/inst/include/population_dynamics/fleet/fleet.hpp#L151) and [evaluate_index_nl](https://github.com/NOAA-FIMS/FIMS/blob/97281c20ba697c66cbbbf5fe4f01241f79183676/inst/include/population_dynamics/fleet/fleet.hpp#L207)l
  *Linkages will change in sprint 2, development should be bare minimum to hook up the new distribution classes and pass tests

- [x] Change names of child distribution classes on lines [154](https://github.com/NOAA-FIMS/FIMS/blob/97281c20ba697c66cbbbf5fe4f01241f79183676/inst/include/population_dynamics/fleet/fleet.hpp#L154) and [211](https://github.com/NOAA-FIMS/FIMS/blob/97281c20ba697c66cbbbf5fe4f01241f79183676/inst/include/population_dynamics/fleet/fleet.hpp#L211) 

- [x] Edit member names and add indices (e.g. dnorm: change x to observed_values[i] and mean to expected_values[i])
- [x] Edit rcpp_tmb_distributions and rcpp_interface
- [x] Delete tmb_distribution.hpp

Andrea-Havron-NOAA:
close with PR #619 PR #624

PR #619, opened 2024-06-04
Connect densities

Andrea-Havron-NOAA:
# What is the feature?

* This PR addresses issue #586 

The fleet.hpp file was edited to depend on the new density classes. The interface and tests were modified to accomodate changes. 

This change does not affect input or output

# How to test this change

test_rcpp_tmb_distributions and test-integration-fims-estimation

# Developer pre-PR checklist


inst/include/distributions/functors/lognormal_lpdf.hpp
```diff
@@ -1,74 +1,94 @@
...
 
-    virtual ~LogNormalLPDF() {}
+        virtual const Type evaluate(const bool &do_log)
+        {
+            this->mu.resize(this->observed_values.size());
+            this->sd.resize(this->observed_values.size());
+            is_na.resize(this->observed_values.size());
+            for (size_t i = 0; i < this->expected_values.size(); i++)
+            {
+                if (this->expected_values.size() == 1)
```
cmlegault, line 39: I do not understand why this if statement is needed, perhaps I am missing something. If only line 45 was used, then wouldn't mu[0] be correctly defined when i=0? Is this if statement trying to fill a vector with a single value? If so, I do not see how it works, but that could be my misunderstanding.

inst/include/distributions/functors/lognormal_lpdf.hpp
```diff
@@ -1,74 +1,94 @@
...
-                this->nll_vec[i] = this->keep.cdf_lower[i] * -log( pnorm(this->observed_value[i], mu[i], sd[i]) );
-                this->nll_vec[i] = this->keep.cdf_upper[i] * -log( 1.0 - pnorm(this->observed_value[i], mu[i], sd[i]) );
-            } */
+            this->nll_vec.resize(this->observed_values.size());
+            for (size_t i = 0; i < this->observed_values.size(); i++)
+            {
+              if(!is_na[i]){
+                #ifdef TMB_MODEL
+                // this->nll_vec[i] = this->keep[i] * -dnorm(this->observed_values[i], mu[i], sd[i], do_log);
+                this->nll_vec[i] = -dnorm(log(this->observed_values[i]), mu[i], sd[i], true) + log(this->observed_values[i]);
```
cmlegault, line 63: Adding the log of observed values to the nll makes complete sense when the observed are fixed data. Where this can run into problems is when the "observed" are actually derived quantities/estimates, such as when stock-recruitment residuals are assumed to follow a lognormal distribution around the stock-recruitment curve. In the latter case, adding the log of the "observed" values gives the minimizer an incentive to estimate lower "observed" values. Jon Deroba and Tim Miller wrote a paper about this: https://doi.org/10.1016/j.fishres.2015.12.002. No need to change anything here, just a reminder to be careful about how the lognormal distribution is used in FIMS.

inst/include/distributions/distributions.hpp
```diff
@@ -15,6 +15,8 @@
 #define FIMS_DISTRIBUTIONS_HPP
 
 #include "functors/density_components_base.hpp"
-#include "functors/tmb_distributions.hpp"
+#include "functors/lognormal_lpdf.hpp"
```
cmlegault, line 18: Not sure where, but somewhere it would be nice to define LPDF and LPMF for developers reading the code. This has no impact on users.

inst/include/distributions/functors/lognormal_lpdf.hpp
msupernaw: @cmlegault your observation is correct, the conditional statement is not needed. Also, is_na will always be false being initialized on line 36 and used at line 60 without being updated.

Cole-Monnahan-NOAA: Yes it's filling a vector with a scalar in a case like a random effect where you have a single mean and SD but a vector of deviations. But it does seem like it should loop over `observed_values.size()` instead of `expected_values.size()` right and combine the two for loops together then right?

Andrea-Havron-NOAA: My logic is that dnorm(observed_values[i], mu[i], sd[i]) needs to work for situations where all three arguments are scalar (e.g. prior), observed_values is a vector but mu is a scalar and sd is a scalar or vector (random effects), and where all three are vectors in the case of data. I think @Cole-Monnahan-NOAA is correct in that the loop should be over observed_values.size().

inst/include/distributions/functors/lognormal_lpdf.hpp
Andrea-Havron-NOAA: Thanks for noting this. We will be tracking in the model whether or not input to the distributions are data. We could add a conditional that only includes the constant term when its a data nll.

Cole-Monnahan-NOAA:
I think given the confusion about the vectors and scalars its important to test the three most common usages. To that end I pushed some new tests in [this commit](https://github.com/NOAA-FIMS/FIMS/commit/1e3cccdeb43acdc2f0cdd4dfa234d9efa34bd059) but they are currently failing. Once they work we should also do the same for the lognormal distribution.

inst/include/distributions/functors/lognormal_lpdf.hpp
```diff
@@ -1,74 +1,93 @@
...
-    LogNormalLPDF() : DistributionComponentBase<Type>() {
+        LogNormalLPDF() : DensityComponentBase<Type>()
+        {
+        }
 
-    }
+        virtual ~LogNormalLPDF() {}
 
-    virtual ~LogNormalLPDF() {}
+        virtual const Type evaluate(const bool &do_log)
```
Cole-Monnahan-NOAA, line 32: I actually don't think we can use do_log=FALSE here b/c later we sum up which won't make sense for negative densities. What is the use case here? Conversion back to natural space, when necessary, should take place outside of this function I think.

inst/include/distributions/distributions.hpp
Cole-Monnahan-NOAA: And also these are not technically lpdf or lpmf since they're negatives. I don't care but something to think about. Many of the downstream uses of the likelihood components will be positive, such as cross validation, and so perhaps defining them as positive and then negating at the end is worth considering.

Andrea-Havron-NOAA:
> I think given the confusion about the vectors and scalars its important to test the three most common usages. To that end I pushed some new tests in [this commit](https://github.com/NOAA-FIMS/FIMS/commit/1e3cccdeb43acdc2f0cdd4dfa234d9efa34bd059) but they are currently failing. Once they work we should also do the same for the lognormal distribution.

Tests pass and I added the same tests for lognormal. The tests that produce dimension mismatch error are currently commented out. These will be implemented in sprint 2 when we can add the dimension checks to CreateModel in Information.

Cole-Monnahan-NOAA (approved):

Cole-Monnahan-NOAA:
@Andrea-Havron-NOAA thanks for getting those working. I'm happy with the PR and have approved it. Thanks!

inst/include/distributions/distributions.hpp
Andrea-Havron-NOAA: I think this is a good point @Cole-Monnahan-NOAA, I will make a note to refactor this in the next sprint

inst/include/distributions/functors/lognormal_lpdf.hpp
Andrea-Havron-NOAA: will add task to next sprint

Andrea-Havron-NOAA:
completed with PR #624

PR #628, opened 2024-06-13
fix(data_mile1): sd of log(x) instead of CV

### LINKED ISSUE #136: What type of uncertainty should be used for each data type?
https://github.com/NOAA-FIMS/FIMS/blob/947a97ffdf57d470a2b9ec3b2c5dc64044f8b8b3/data-raw/data_mile1.R#L47-L51

Above is an example for "landings" where we are using CV; but, I am uncertain if we should be using CV. So, what type of uncertainty is appropriate for the following data types:
* landings
* index
* age - input sample size
* weight-at-age - NA

kellijohnson-NOAA:
# What is the feature?
* For catch and index data the measurement of uncertainty should be the standard deviation (sd) of the logged observations but the operating model outputs a CV.

Close #136

* I added a temporary function in data_mile1.R to calculate the sd the same way that the vignette was doing it and then updated those values in the data frame for landings and index.

* This time series of uncertainty can be fed to the modules but I did not change the vignette.

# How to test this change

# Developer pre-PR checklist


data-raw/data_mile1.R
```diff
@@ -15,6 +15,10 @@
 ###############################################################################
 # Helper functions and load packages
 ###############################################################################
+cv_2_sd <- function(x) {
+  log(sqrt(log(x^2 + 1)))
```
Bai-Li-NOAA, line 19: I think I need help from @Andrea-Havron-NOAA here. The `cv_2_sd()` function is designed to convert the CVs from model comparison project for the `log_obs_error` in FIMS fleet module. To me, standard deviation on the log scale is calculated from the specified CVs using the relationship `sdlog=sqrt(log(CV^2+1))`. Does the additional `log` outside of `sqrt(log(CV^2+1))` imply that we log the standard deviation on the log scale? Do we need to consider renaming the function to `cv_2_logsd()`?

Bai-Li-NOAA (commented):
If there is no need to rename the cv_2_sd function, the changes look good to me. We might need to create a separate issue in the case-study repo to update the code related to cv and sd conversion.

data-raw/data_mile1.R
Andrea-Havron-NOAA: The outside log is the parameter transformation where internally the model takes the exponential to keep the value positive. FIMS will be using internal transformations to bound parameters. Most parameters will need to be transformed so my personal preference would be to have the cv_2_sd return the value without the outer log and then have a separate function that applies the correct transformation based on the parameter.

kellijohnson-NOAA:
Thank you @Bai-Li-NOAA and @Andrea-Havron-NOAA. I will change it to be SE rather than log(SE). Then, at a later date we can augment the SE internally to be in the desired space when setting up the modules.

data-raw/data_mile1.R
```diff
@@ -15,6 +15,10 @@
 ###############################################################################
 # Helper functions and load packages
 ###############################################################################
+cv_2_sd <- function(x) {
+  sqrt(log(x^2 + 1))
+}
+
```
kellijohnson-NOAA, line 21: Now just the sd calculated from the CV rather than the log of the standard deviation. Thanks again @Bai-Li-NOAA for catching this and @Andrea-Havron-NOAA for confirming what we need.

Andrea-Havron-NOAA: looks good!

Andrea-Havron-NOAA (approved):

data-raw/data_mile1.R
Cole-Monnahan-NOAA: But these aren't parameters so I don't think it's necessary to log it internally? It'll just be exponentiated again. I'm not aware of any assessments that will estimate these in each year. At most an additional variance term is estimated on top of the supplied values.

PR #633, opened 2024-06-18
fix: Use native pipe instead of %>%

kellijohnson-NOAA:
# What is the feature?

* We no longer import the magrittr pipe

* Use native pipe


Bai-Li-NOAA (approved):
The changes look good!

PR #640, opened 2024-07-02
Add an R test to run FIMS in parallel using {snowfall}

Bai-Li-NOAA:
# What is the feature?

This PR addresses issues #378 and #449. The R test verifies that we can run FIMS models in parallel using {snowfall}.

- Set up an R testthat fixture to pre-simulate 100 sets of test data using the model comparison project OM.
  - `tests/testthat/fixtures/simulate-integration-test-data.R`
  - `tests/testthat/fixtures/integration_test_data.RData`
- Add a helper function to set up Rcpp modules and run FIMS.
  - `tests/testthat/helper-integration-tests-setup.R`
- Add an R test to run FIMS in serial and parallel, and compare the execution time and results of running the models in serial versus parallel.
  - `tests/testthat/test-parallel-with-snowfall.R`
- Update the DESCRIPTION file to include the packages needed for the test under the Suggests section.

Other notes:
- The issue #378 has mentioned two additional packages for running FIMS in parallel: {processR} and {Rmpi}. We could create a vignette to demonstrate how to run FIMS in parallel using different packages and benchmark them. This could be a new issue to track this task.
- Initially, the R test ran successfully on my local machine but failed on GHA through commit f5d4b6f01f3129055f7b680bbcd40f019c9b75e9 (see [failed GHA log](https://github.com/NOAA-FIMS/FIMS/actions/runs/9762956877/job/26947645590)). The results from parallel runs differed from those of serial runs, even when using the same model, data, and seed. It took me some time to realize that the mismatch was caused by different versions of tools used locally and on GHA. Initially, the test data from the serial run were generated locally and uploaded online as a test fixture. Now, the results from the serial run are simulated on the fly within the test. For FIMS results to be reproducible, similar to [Stan](https://mc-stan.org/docs/reference-manual/reproducibility.html), we need to ensure the following components are identical: FIMS version, R interface, included library versions, operating system version, computer hardware, C++ compiler, and configuration of the FIMS call. @msupernaw suggests narrowing it down to one particular component. I think this requirement list for reproducibility could be documented in a vignette. 

No

# How to test this change

All GHA checks have passed.

# Developer pre-PR checklist

### DIFF
--- a/tests/testthat/helper-integration-tests-setup.R
+++ b/tests/testthat/helper-integration-tests-setup.R
@@ -1,0 +1,186 @@
+# Set-up Rcpp modules and fix parameters to "true" values from the OM
+setup_and_run_FIMS <- function(iter_id,
+                               om_input_list,
+                               om_output_list,
+                               em_input_list,
+                               estimation_mode = TRUE) {
+  # set.seed(seed = 123)
+
+  # Load operating model data
+  om_input <- om_input_list[[iter_id]]
+  om_output <- om_output_list[[iter_id]]
+  em_input <- em_input_list[[iter_id]]
+
+  clear()
+  # Recruitment
+  # create new module in the recruitment class (specifically Beverton-Holt,
+  # when there are other options, this would be where the option would be chosen)
+  recruitment <- new(BevertonHoltRecruitment)
+
+  # NOTE: in first set of parameters below (for recruitment),
+  # $is_random_effect (default is FALSE) and $estimated (default is FALSE)
+  # are defined even if they match the defaults in order to provide an example
+  # of how that is done. Other sections of the code below leave defaults in
+  # place as appropriate.
+
+  # set up logR_sd
+  # logR_sd is NOT logged. It needs to enter the model logged b/c the exp() is
+  # taken before the likelihood calculation
+  recruitment$log_sigma_recruit$value <- log(om_input$logR_sd)
+  recruitment$log_sigma_recruit$is_random_effect <- FALSE
+  recruitment$log_sigma_recruit$estimated <- FALSE
+  # set up log_rzero (equilibrium recruitment)
+  recruitment$log_rzero$value <- log(om_input$R0)
+  recruitment$log_rzero$is_random_effect <- FALSE
+  recruitment$log_rzero$estimated <- TRUE
+  # set up logit_steep
+  recruitment$logit_steep$value <- -log(1.0 - om_input$h) + log(om_input$h - 0.2)
+  recruitment$logit_steep$is_random_effect <- FALSE
+  recruitment$logit_steep$estimated <- FALSE
+  # turn on estimation of deviations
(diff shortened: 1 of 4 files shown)


tests/testthat/helper-integration-tests-setup.R
```diff
@@ -0,0 +1,186 @@
...
+  population$nyears <- om_input$nyr
+  population$SetMaturity(maturity$get_id())
+  population$SetGrowth(ewaa_growth$get_id())
+  population$SetRecruitment(recruitment$get_id())
+
+  # Set-up TMB
+  CreateTMBModel()
+  # Create parameter list from Rcpp modules
+  parameters <- list(p = get_fixed())
+  obj <- TMB::MakeADFun(data = list(), parameters, DLL = "FIMS", silent=TRUE)
```
Cole-Monnahan-NOAA, line 159: I added `silent=TRUE` here to declutter the console. It prevents printing the mgc at each iteration.

tests/testthat/helper-integration-tests-setup.R
```diff
@@ -0,0 +1,186 @@
...
+  if (estimation_mode == TRUE){
+
+    opt <- with(obj, optim(par, fn, gr,
+                           method = "BFGS",
+                           control = list(maxit = 1000000, reltol = 1e-15)
+    ))
+  }
+  # Call report using MLE parameter values, or
+  # the initial values if optimization is skipped
+  report <- obj$report(obj$env$last.par.best)
```
Cole-Monnahan-NOAA, line 170: I simplified this a bit and clarified in the comments what parameter vector is being used.

Specifically  `obj$par` and `obj$env$last.par.best` are the same if optimization is skipped. Also note that since there are no random effects you can pass `obj$par` to report but generally cannot do that. `obj$report()` requires the full vector.

tests/testthat/test-parallel-with-snowfall.R
```diff
@@ -0,0 +1,70 @@
...
+    estimation_mode = TRUE)
+}
+end_time_serial <- Sys.time()
+estimation_time_serial <- end_time_serial - start_time_serial
+
+test_that("Run FIMS in parallel using {snowfall}", {
+
+  core_num <- parallel::detectCores() - 1
+  snowfall::sfInit(parallel = TRUE, cpus = core_num)
+  snowfall::sfLibrary(FIMS)
```
Cole-Monnahan-NOAA, line 34: I'm not sure why it ran without loading FIMS in the child processes. Had to add it to run locally, feel free to remove if you want.

Bai-Li-NOAA: According to the guidelines from Advanced R, we should avoid including `library()` calls in the test files. Could you try running `devtools::install()` first and then execute the tests (e.g., devtools::test()) without using library(FIMS) to see if it works locally? Thank you!

PR #641, opened 2024-07-02
Fix FIMSFrame

### LINKED ISSUE #639: [Bug]: `type.convert()` is reformatting date to character without leading zeros in data_mile1
### Describe the bug

Here
https://github.com/NOAA-FIMS/FIMS/blob/0296e41c9834991a62cd237d2387dca687bb6931/data-raw/data_mile1.R#L152
in data-raw, the script that creates data_mile1 the dates are being reformatted as character strings rather than date objects. Thus, when `n_years()` attempts to calculate the number of years in the data it thinks that "9" is the maximum year rather than "30". This line of code has not changed for two years and used to work so I am not sure what is going on.

I am wondering if we can just remove the call to `type.convert()` and just have `rbind()`?

### To Reproduce

```
source(textConnection(readLines("data-raw/data_mile1.R")[1:151]))
type.convert(rbind(landings_data, index_data), as.is = TRUE) |>
  dplyr::pull(dateend) |>
  class()
[1] "character"
> rbind(landings_data, index_data) |> dplyr::pull(dateend) |> class()          
[1] "Date"

 type.convert(rbind(landings_data, index_data), as.is = TRUE) |>
  dplyr::pull(dateend) |> 
  max()  
[1] "9-12-31"
 rbind(landings_data, index_data) |> dplyr::pull(dateend) |> max()  
[1] "0030-12-31"
```

### Screenshots

_No response_

Windows

_No response_

'0.2.0.0' since #37e659c

### Additional Context

_No response_

kellijohnson-NOAA:
# What is the feature?
* Fix #639 where the leading zeros of the year if using fake years of single-digit integers are removed when saving to a csv file and reading the data back in. Instead of worrying about always having leading zeros, I have created a feature that ensures FIMSFrame creates date objects.
* This problem I think is being brought up by some change that is external to FIMS. That is, data_mile1 use to have leading zeros and now it does not but the code to create the data_mile1 has not changed in a way that would cause any differences. I think something within base might have changed. It is okay though because now FIMS will be more robust to user errors and differences in R versions.

Using as.Date() for datestart and dateend to ensure the leading zeros are present. And, the format portion of as.Date() ensures that the dates are formatted with dashes instead of slashes and will give users an error if they are not.

Objects from FIMSFrame are not really being used in the code right now so it is not a well-spread problem. We just noticed it in the data group when we started using the slots of FIMSFrame for things.

# How to test this change
The change is tested in the automated tests.

# Developer pre-PR checklist
- [x] Ran some of the R tests locally and also used this branch to rebase the m2-r-output branch to see if the broken tests there were finally passing and they are 🥳 

This PR is broken up into two commits (that should later be squashed) but I would like comments from @k-doering-NOAA on the first commit and comments from @iantaylor-NOAA on the second commit. Thank you.

Closes #639

### DIFF
--- a/R/fimsframe.R
+++ b/R/fimsframe.R
@@ -225,27 +225,8 @@ setValidity(
       errors <- c(errors, "data must have at least one row")
     }
 
-    # Check columns
-    if (!"type" %in% colnames(object@data)) {
-      errors <- c(errors, "data must contain 'type'")
+    errors <- c(errors, validate_data_colnames(object@data))
     }
-    if (!"datestart" %in% colnames(object@data)) {
-      errors <- c(errors, "data must contain 'datestart'")
-    }
-    if (!"dateend" %in% colnames(object@data)) {
-      errors <- c(errors, "data must contain 'dateend'")
-    }
-    if (!"dateend" %in% colnames(object@data)) {
-      errors <- c(errors, "data must contain 'value'")
-    }
-    if (!"dateend" %in% colnames(object@data)) {
-      errors <- c(errors, "data must contain 'unit'")
-    }
-    if (!"dateend" %in% colnames(object@data)) {
-      errors <- c(errors, "data must contain 'uncertainty'")
-    }
-    if (!"age" %in% colnames(object@data)) {
-      errors <- c(errors, "data must contain 'age'")
     }
 
     # TODO: Add checks for other slots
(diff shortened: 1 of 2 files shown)


iantaylor-NOAA (approved):
Nice work @kellijohnson-NOAA.

I like the new `validate_data_colnames()` function, and the extra format checks within `FIMSFrame()`. It's probably useful that the bug cropped up because this system of error checks will surely pay off in the future by being more robust (as you noted in the PR).

k-doering-NOAA (approved):
>The column names of the S4@data object were being checked with the
validation function but there was no check in FIMSFrame and to do the
calculations within FIMSFrame all of the columns need to be present.
Right now we are checking it twice, once in FIMSFrame and once in the
validator of the FIMSFrame class. We might need a different class like an
input data class and we could check the class upon input to FIMSFrame but
this works for right now. @k-doering-NOAA do you have any ideas here?

From reading on the [validate function in the advanced R book](https://adv-r.hadley.nz/s4.html#validator), it sounds like checking should be done in the validation function; however, if it means the user will not get a helpful error, that does seem like a problem. Maybe we should be a little unconventional (I think) and just do the checking when putting together the dataframe? Or is there a reason we should be checking the columns names twice?

Just an idea, though, and I think your solution is fine for now!

kellijohnson-NOAA:
Thanks @iantaylor-NOAA and @k-doering-NOAA for the review and comments. I have been thinking about the best way forward (hence the delay in my response) and I think we should leave the check in there twice for right now because we have multiple developers working on the methods and checking in the validity function for the column names of the output to be correct saves us from writing an additional formal test in the testthat code. So, in the future we might need to remove it but the extra precaution it provides right now appears to be a good thing.

PR #642, opened 2024-07-02
refactor: Deletes issue and PR templates

kellijohnson-NOAA:
# What is the feature?
* Moves PR and issue templates to NOAA-FIMS/.github/.github so they are available to all repositories within NOAA-FIMS.

Moved the files to NOAA-FIMS/.github/.github see the commit message for the exact commit.

The refactor issue template was not moved.

# How to test this change
Go to another repository in FIMS and see the available issue templates 🤯 or see below
image

# Developer pre-PR checklist


k-doering-NOAA (approved):
Thanks @kellijohnson-NOAA , I did a little bit of playing around with issue templates in this [repo](https://github.com/NOAA-FIMS/test-issue-forms) (which can be deleted shortly). I learned:

- You get all of the issue forms from the .github repository showing up (I don't think there is a way to change this?). This is unfortunate, as I could see the "seaside chat" request in the repo, when it probably isn't needed by all repos.
- If you add a new form to the `.github/ISSUE_TEMPLATE` subfolder in a single repository (regardless if it is named the same or differently than an existing form in the `.github` repository), you will ONLY see that individual template and none of the ones from the `.github` repository. I guess that means you would have to copy over templates from the `.github` repository if you wanted to use them, creating some duplication?

I guess knowing better how these issue templates in the `.github` repository work, do you still think moving them to the `.github` repository is the best course of action? If so, feel free to merge in the PR; if not, perhaps we should reconsider.

kellijohnson-NOAA:
Thanks @k-doering-NOAA for the review. I reverted the .github repository back one commit so the seaside chat issue template is no longer available globally, which addresses your first concern. Regarding your second concern, I think that if a repo feels like copying over some of the templates from .github to a specific repository we should think about the opposite action of why not use all the templates and/or add the specific template in the repo to .github? I am going to merge this pull request because I think the benefit of reducing current amount of duplication is more beneficial than the future problem of having duplication.

PR #654, opened 2024-07-25
Fix proportion female

kellijohnson-NOAA:
# What is the feature?
* Reverts proportion_female to a single constant of 0.5

* Using a constant scalar.

* The case studies need to be changed to not set proportion_female.


kellijohnson-NOAA:
Now in cleanup_dev branch to be brought into dev.

PR #663, opened 2024-08-23
Update DESCRIPTION to depend on R 4.1.0

iantaylor-NOAA:
# What is the feature?
We use the native pipe `|>` in many places which was only available in R version 4.1.0. Status-quo dependency was R 4.0, so someone happened to have an R version like 4.0.5, they could install FIMS but it would not work.

Change DESCRIPTION

No


kellijohnson-NOAA:
Thanks @iantaylor-NOAA for this PR. We are no longer making PR on main so I have added this to dev with b8ad1c6c3c832dda456e8163d572fb19c255f646 and I am closing this PR.

PR #672, opened 2024-10-02
Resolve R CMD check note about 'scales' package

arni-magnusson:
This pull request resolves a Note raised by R CMD check, moving the `scales` package from Imports to Suggests.

The `scales` package is not imported from, so I'm guessing it's not required as an Imports entry. Not a big issue - just trying to keep R CMD check happy :)

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -61,7 +61,6 @@ Imports:
     jsonlite,
     methods,
     Rcpp,
-    scales,
     TMB (>= 1.8.0)
 Suggests: 
     covr,
@@ -69,6 +68,7 @@ Suggests:
     parallel,
     remotes,
     rmarkdown,
+    scales,
     snowfall,
     testthat (>= 3.0.0),
     tidyverse,
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -61,6 +61,7 @@ Imports:
     jsonlite,
     methods,
     Rcpp,
+    scales,
     TMB (>= 1.8.0)
 Suggests: 
     covr,
@@ -68,7 +69,6 @@ Suggests:
     parallel,
     remotes,
     rmarkdown,
-    scales,
     snowfall,
     testthat (>= 3.0.0),
     tidyverse,
--- a/R/FIMS-package.R
+++ b/R/FIMS-package.R
@@ -25,3 +25,7 @@
 #' @export TMBDmultinomDistribution
 ## usethis namespace: end
 NULL
+
+ignore_unused_imports <- function() {
+  scales::date_format
+}
--- a/R/FIMS-package.R
+++ b/R/FIMS-package.R
@@ -26,6 +26,9 @@
 ## usethis namespace: end
 NULL
 
+# Resolve note from R CMD check on unused imports
+# See https://github.com/r-lib/devtools/blob/main/R/devtools-package.R and
+# https://r-pkgs.org/dependencies-in-practice.html#how-to-not-use-a-package-in-imports
 ignore_unused_imports <- function() {
   scales::date_format
 }


kellijohnson-NOAA:
Thanks @arni-magnusson for the PR. I am not entirely sure why it thinks that we do not need scales because it is used in some {ggplot2} code here
https://github.com/NOAA-FIMS/FIMS/blob/f0d4e766ddfc70c6dbb99e92385298c31ccff136/R/fimsframe.R#L183
Though we should change to `scales::label_date()` per the documentation for `scales::date_format()` but that is a different issue.

I always thought that if a package is used within the code it should be in imports and if it is used by developers say for creating data sets like {usethis} can be in the data-raw folder, then it should be in Suggests. Feel free to correct me if I am wrong.

arni-magnusson:
You're absolutely right. At closer inspection, I can see that the `scales` package should indeed be in the Imports field of the DESCRIPTION file, not in Suggests.

This Note from R CMD check is effectively a false positive. I hadn't come across this case of false positive from R CMD check before, as I tend to use `@ importFrom pkg function` for functions that I use. I fully appreciate if the FIMS style is not to import every function that is used, following the recommendation in Wickham and Bryan's book on R Packages, section [11.4.1](https://r-pkgs.org/dependencies-in-practice.html#how-to-not-use-a-package-in-imports).

Wickham and Bryan point to a workaround solution in their section [11.4.1.1](https://r-pkgs.org/dependencies-in-practice.html#how-to-not-use-a-package-in-imports), defining a dummy ignore_unused_imports() function. I've modified the pull request to apply this solution, also used in [devtools](https://github.com/r-lib/devtools/blob/main/R/devtools-package.R). Feel free to ignore the pull request if this solution seems clunky :)

An alternative solution pointed out by Wickham and Bryan section [11.4.1.1](https://r-pkgs.org/dependencies-in-practice.html#how-to-not-use-a-package-in-imports) would be to add this entry to FIMS-package.R:
`#' @importFrom scales date_format`

kellijohnson-NOAA:
@arni-magnusson thank you for your pull request. Upon further investigation, I believe that this is happening because {scales} is used inside of `methods::setMethod()` and not as standalone code in a normal function. I am going to figure out exactly what call the note is being generated from in `devtools::check()` so I can report it as a formal issue and then hopefully we will not have to have a workaround. I may implement your current suggestion in dev if we do not get a more permanent solution by the next release. So, I will leave this PR open as a reminder.

Second, all future contributions should be made to dev and not main. This is a new protocol for us as of ~ June so it is not well documented. I will try to fix that.

kellijohnson-NOAA:
I finally figured it out 😓! Line 169 in the following code is incorrect
https://github.com/NOAA-FIMS/FIMS/blob/f0d4e766ddfc70c6dbb99e92385298c31ccff136/R/fimsframe.R#L166-L177

It should be
```
setMethod(
  f = "plot",
  signature = "FIMSFrame",
  definition = function(x, y, ...) {
    ggplot2::ggplot(
      data = x@data,
      mapping = ggplot2::aes(
        x = as.Date(.data[["datestart"]]),
        y = .data[["value"]],
        col = .data[["name"]]
      )
    ) +
```
because the `args(plot)` are `plot(x, y, ...)` and the addition to the generic method needs to have the same arguments or more if `...` is present. My original code only has `x`, which was causing `lapply(exprs, deparse)`, where exprs are all code chunks in the package, to call out the plot method as a .local function rather than a standard function. This later leads to `find_bad_exprs()` not being able to pick up a package being used in the code chunk. FYI all of this is in R CMD check --as-cran or in `tools:::.check_packages_used()`.

@arni-magnusson let me know if you would like to modify your PR to have proposed fix to dev or I can open a new PR. Thanks again 🙏 for bringing this up. Although, it turned into quite the 🐰 🕳️ it was a fun one.

Below are some references that helped me find some answers.
* [tools:::.check_packages_used()](https://github.com/wch/r-source/blob/0c26529e807a9b1dd65f7324958c17bf72e1de1a/src/library/tools/R/QC.R#L6473)
* ["All declared Imports should be used"](https://github.com/wch/r-source/blob/0c26529e807a9b1dd65f7324958c17bf72e1de1a/src/library/tools/R/QC.R#L6372)
* [stack overflow on how to declare an s3 or s4 dependency](https://stackoverflow.com/questions/75382258/how-to-declare-a-dependency-on-an-r-package-from-which-you-only-use-s3-s4-method)
* [GitHub Issue on r-pkgs mentioned in previous stack overflow post](https://github.com/hadley/r-pkgs/issues/828)
* [R bug report](https://bugs.r-project.org/show_bug.cgi?id=18465)
* [Another R bug report, when package name is in quotes](https://bugs.r-project.org/show_bug.cgi?id=18466)

arni-magnusson:
Thanks for the investigation and following up, @kellijohnson-NOAA. I've modified the PR to target the dev branch. That was a new button for me to learn about in GitHub :)

kellijohnson-NOAA:
@arni-magnusson thank you for this PR I have fixed it in a20b806e6e568f7df2c3714206088e6540fe11ed with a slightly different method than what you outlined and I described it [above](https://github.com/NOAA-FIMS/FIMS/pull/672#issuecomment-2436440280). The warning is now taken care of. Thank you again for the contributions and bringing it to my attention.

Issue #677, opened 2024-10-18
[Developer Issue]: add a standalone GHA test for checking the speed of FIMS parallel runs


Bai-Li-NOAA:
### Description

The testthat check for FIMS parallel run speed was removed due to GHA failures on Mac, caused by the limited availability of cores (i.e., 2). A standalone test for evaluating parallel run speed will be added to CI once the wrapper functions for preparing FIMS inputs are fully developed.

PR #678, opened 2024-10-18
fix parallel tests to avoid comparing run time with serial execution

Bai-Li-NOAA:
# What is the feature?
* This PR fixes the failed GHA on the `dev` branch

* removed code that compares parallel run time with serial execution, focusing on ensuring that parallel runs work correctly and produce consistent results (see `tests/testthat/test-parallel-with-snowfall-with-wrappers.R`)
* deleted parallel tests that don't use the wrapper functions
* a standalone test for checking the speed of parallel runs will be added to CI once the wrapper functions for preparing FIMS inputs are fully developed. A new issue #677 has been created to track this task. 

* Other branches will need to be rebased to fix GHA failures on those branches


kellijohnson-NOAA:
@Bai-Li-NOAA I fixed a couple of formatting things, very nit picky ...
1. when you use {package-name} you should not describe that it is a package, see [style guide on package names](https://style.tidyverse.org/documentation.html#package-names)
2. Indenting for a function should be two spaces from the first letter of the function call if it cannot fit on one line instead of a hanging indent

I force pushed those changes and will merge it into dev once the checks have all passed.

PR #682, opened 2024-11-04
Macro logging system FIMS_DEBUG_LOG, FIMS_ERROR_LOG, FIMS_INFO_LOG, FIMS_WARNING_LOG

kellijohnson-NOAA:
# What is the feature?
* This PR remove the old logging and fully implements the M2 logging system with macros. The macros find information that is stored within the C++ environment like file, line, working directory, user name, timestamp, and other info as well as returning the erorr/log statement that lives in the C++ code. The system also has a DEBUG log and throw on error capability. Thus, if throw on error is initiated then the logging system will capture the output in a log even if the log is not available in the R session, e.g., your R session crashes.  
* refactored finalize
* pretty format for output
* added format string for  json
* created a logging vignette
* Adds information about DFIMS_DEBUG information
* Adds throw on error
* the json reader cannot read \ from windows machines and if they are not doubled then they don't work in R so went to using forward slash for windows machines.

TODO:
* think about renaming some of the Rcpp functions

* Macros were created for the logging class and the old logging is removed. R functions are being exported using Rcpp that allow the user to look at several of the log entries using automatic filtering.

* Yes, @nathanvaughan-NOAA will be rebasing/re-writing the length changes on top of this commit because many of the same files were touched. Thus, this PR is not completely final but good enough to get into dev.


kellijohnson-NOAA:
This PR was largely reviewed during the October 31st seaside chat and thus is being merged in without further review to dev so @nathanvaughan-NOAA can continue to work on length. More documentation, especially of Rcpp functions is needed and will be done in parallel with future development before this is merged into main as M2.

Additional shout out to @rklasky for letting me know about Doxygen Documentation Generator, which helped me document C++ code 💪 🥇.

Issue #662, opened 2024-08-07
[Developer Issue]: M2 Quality control task list


kellijohnson-NOAA:
### Description

This issue will be a collect all for small tasks that need to be completed for quality control after features are implemented in milestone 2. For example, @rklasky mentioned there are a few spelling mistakes in the code that need cleaned up. Feel free to edit this comment or post more comments below regarding small things that need to be done before our next release.
```[tasklist]
### Tasks
- [x] rename get_output() to not start with get because that is used in the accessors of S4 objects
- [x] Fix output of `get_output()` if `finalize()` has not been called, e.g., `Invalid request to "get_output". Please call finalize first.[1] ""` should not have [1] but it is not clear in the code why it is being printed because the c++ code is `"Invalid request to \"get_output()\". Please call finalize() first."` (Matthew)
- [x] Add Bayesian demo and other distribution tests, make this a Discussion (Kelli)
- [x] Review all R functions to identify which should be exposed to users with the `@export tag` and then group them in the `_pkgdown.yml` file
- [x] Document Rcpp functions with titles, e.g., `Rcpp::function("CreateTMBModel", &CreateTMBModel, "Initialize the FIMS/TMB model")`
- [x] Fix function names throughout to use camel case in the C++.
- [x] update initialize_fims() and make sure only length-comp data scenario works (Bai)
- [x] remove .vscode folder if we want to and add it to .gitignore (Matthew)
- [x] Ensure use :: for all packages other than FIMS, e.g., `methods::new()` in maturity vignette (Andrea)
- [x] Line 196 of rcpp_interface_base.hpp should be changed to `if (static_cast<size_t>(pos) == 0 || static_cast<size_t>(pos) > this->storage_m->size()) {` from `if (pos == 0 || pos > this->storage_m->size()) {`
- [ ] https://github.com/NOAA-FIMS/FIMS/issues/691
- [x] remove commented-out code, e.g., `// data_indicator<tmbutils::vector<Type> , Type> keep;`
- [x] Fix spacing in information.hpp `FIMS_INFO_LOG("Random effect size for distribution " + fims::to_string(d->id) + "is: " + fims::to_string(d->x.size()));` where it needs a space before is.
- [x] remove "TMB" from distribution class names and distribution files.
- [x] should be able to remove `library(TMB)` call from vignettes but this needs to be tested
- [x] fix data-raw/data_mile1.R so you can source the file from your working directory, right now it only works if you run it line by line (Bai)
[...]

PR #688, opened 2024-11-25
Add R wrapper functions to prepare parameters for running FIM

Bai-Li-NOAA:
# What is the feature?
* This PR adds R wrapper functions to prepare parameters for running FIMS

**Key updates**
- `R/create_default_parameters.R`
    - `create_default_*()`: creates default parameters for a FIMS model
    - `update_parameters()`: update default parameters for a FIMS model
- `R/initialize_modules.R`: initialize_*(): initialize all required modules for a FIMS model   
    - `set_param_vector()`: a core function for passing parameter inputs to FIMS parameter vectors
- `R/cli_messages.R`: contains reusable checks to verify the validity of function inputs

**Testing and documentation** 
- A working `fims-demo` vignette has been added.  
- Integration tests, including parallel tests, have passed successfully. except for those on macOS. Looks like it’s an issue with installing R 4.4.2 correctly. I’m assuming the r-lib/actions team will sort it out.
- Several new unit tests validate the wrapper functions exposed to users.  

**Help needed**
- Suppressing Printed Memory Addresses:
When initializing modules, the R console prints memory addresses for modules such as selectivity, fleet, index distribution, recruitment, maturity, and population. @msupernaw tips on suppressing these outputs would be appreciated.
- "No Visible Binding for Global Variable" Warning in R CMD Check:
This warning is caused by the use of the pipe operator (|>). @kellijohnson-NOAA, although [potential solutions](https://www.r-bloggers.com/2019/08/no-visible-binding-for-global-variable/) exist, I am unsure of the best approach. Suggestions would be greatly appreciated!

I can rebase the branch with `dev` and squash the commits once Matthew's distribution refactor work has been merged.

* Yes


R/distribution_formulas.R
```diff
@@ -0,0 +1,380 @@
...
+#' @param link A string specifying the model link function. For example,
+#'   `"identity"` or `"log"` are appropriate names for the [stats::gaussian()]
+#'   distribution. `"log"` and `"logit"` are the defaults for the lognormal and
+#'   the multinomial, respectively.
+#' @return
+#' An object of class `family` (which has a concise print method). This
+#' particular family has a truncated length compared to other distributions in
+#' [stats::family()].
+#' \item{family}{character: the family name.}
+#' \item{link}{character: the link name.}
```
kellijohnson-NOAA, line 349: All instances of `\item{}{}` should be wrapped in `\describe{}` like
```
#' \describe{
#'   \item{family}{character: the family name.}
#'   \item{link}{character: the link name.}
#' }
```
here and elsewhere. I noticed this while working on the PR to this PR. I am not going to fix it in Matthew's branch, instead, we can fix it here after we merge that one in.

kellijohnson-NOAA:
@Bai-Li-NOAA the current plan of attack is for you to squash all of these commits into a few commits, one is fine. Then, rebase to dev. We will merge this in before dev-length and I will rebase dev-length onto the new dev. That way we can create tests for dev-length using the new wrapper functions. While you are doing that, I will review the changes today in hope that we can merge into dev tomorrow. Please let me know if that timeline does not work for you for any reason. I am happy to amend it.

R/cli_messages.R
kellijohnson-NOAA, line 1: This file should be named `R/check_.R

R/cli_messages.R
```diff
@@ -0,0 +1,62 @@
+#' Check if the data argument is an object of class `FIMSFrame`
+#'
+#' @description
+#' This function checks whether the provided `data` argument is an object of the
+#' class `FIMSFrame`. It ensures that the data adheres to the required structure
+#' for use with FIMS-related functions.
+#' @param data The object to check.
+#' @return NULL If the check passes (i.e., `data` is of class `FIMSFrame`).
```
kellijohnson-NOAA, line 8: The text after @return should start on their own line, and I also provided some edits to the text
```
#' @return
#' If the check passes (i.e., `data` is of class `FIMSFrame`), then `NULL` is returned,
#' otherwise the function is aborted and an error is thrown.

File: R/cli_messages.R [hunk 3]
```diff
@@ -0,0 +1,62 @@
...
+#' The function throws an error if the argument is not of class `FIMSFrame`.
+#' @examples
+#' \dontrun{
+#' data("data_mile1")
+#' fims_data <- FIMSFrame(data_mile1)
+#' check_arg_data(fims_data)
+#' check_arg_data(data = 1:10)
+#' }
+#' @noRd
+check_arg_data <- function(data) {
```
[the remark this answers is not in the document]
kellijohnson-NOAA: After reading a lot of documentation on R functions, we equally call the input to a function a parameter and an argument. So, I am fine with keeping it as "arg" but I still think that it should be `check_data_arg` rather than `check_arg_data`.

R/cli_messages.R
```diff
@@ -0,0 +1,62 @@
...
+#'
+#' @description
+#' This function checks if a given input is valid by comparing it against a
+#' predefined set of valid options. It ensures that the input matches one of
+#' the acceptable options and throws an error with a descriptive
+#' message if it does not.
+#' @param input A character string representing the input to validate.
+#' @param valid_options A character vector of valid options against which the input will be checked.
+#' @param arg_name A character string describing the name of the argument or
+#' context of the input being validated
```
kellijohnson-NOAA, line 35: subsequent lines of @param should be indented two spaces and the sentence is missing a full stop at the end of the line.

R/cli_messages.R
```diff
@@ -0,0 +1,62 @@
...
+#' @examples
+#' \dontrun{
+#' valid_selectivity_options <- c("LogisticSelectivity", "DoubleLogisticSelectivity")
+#' check_valid_input("LogisticSelectivity", valid_selectivity_options, "form")
+#' check_valid_input("InvalidOption", valid_selectivity_options, "form")
+#' }
+#' @noRd
+check_valid_input <- function(input, valid_options, arg_name) {
+  # Check if the input is a character string
+  if (!is.character(input) || length(input) != 1) {
```
kellijohnson-NOAA, line 47: We should be using `inherits(input, "character")` rather than `is.character()` because `inherits()` is faster.

R/distribution_formulas.R
```diff
@@ -0,0 +1,380 @@
...
+    }
+
+    if (grepl("comp", data_type) &&
+        (family[["family"]] == "lognormal" ||
+         family[["family"]] == "gaussian")) {
+      fam_name <- family[["family"]]
+      cli::cli_abort(c(
+        "x" = "The family, {.code {fam_name}} is not available for the data
+               type, {.code {data_type}}.",
+        "i" = "The available families for this data type are: `multinomial`."
```
kellijohnson-NOAA, line 49: I think that we should store the options as data variables somewhere rather than having them saved here. This change does not need to happen today, it can be part of a larger code refactor later but it would be great if we could standardize the distributions that are available against all data types and make some kind of look up table or something.

R/distribution_formulas.R
```diff
@@ -0,0 +1,380 @@
...
+        "i" = "The available families for this data type are: `lognormal` or
+               `gaussian`."
+      ))
+    }
+  }
+
+  if (!all(sd[["value"]] > 0)) {
+    sd_value <- sd[["value"]]
+    cli::cli_abort(c(
+      "x" = "At least one value for sd, {.code {sd_value}} is out of bounds.",
```
kellijohnson-NOAA, line 67: Missing comma after `{.code {sd_value}}`

R/create_default_parameters.R
```diff
@@ -0,0 +1,584 @@
...
+    list(
+      log_Fmort.value = log(rep(0.00001, data@n_years)),
+      log_Fmort.estimated = TRUE
+    )
+  } else {
+    list(
+      log_q.value = 0,
+      log_q.estimated = TRUE
+    )
+  }
```
kellijohnson-NOAA, line 284: I do not think that being a survey or a fleet is mutually exclusive, or at least it is not supposed to be. But, will down-stream stuff fail if both F and q are present. You should be able to fit to catch data and index data from the same fleet if you want to. This has been a problem in the past with users not knowing it is possible to do so. I think I will take more time to ensure that this is true and that a list of length four here will work if both types of data are present.

R/create_default_parameters.R
```diff
@@ -0,0 +1,584 @@
...
+#'
+#' @description
+#' This function creates default fleet selectivity parameters based on the provided
+#' selectivity form. It supports different selectivity forms such as LogisticSelectivity
+#' and DoubleLogisticSelectivity.
+#' @param fleet_name A character. Name of the fleet.
+#' @param form A character. Selectivity form (e.g., LogisticSelectivity and DoubleLogisticSelectivity).
+#' @return A list containing default selectivity parameter values for the specified form.
+#' @noRd
+create_default_selectivity <- function(fleet_name, form = c("LogisticSelectivity", "DoubleLogisticSelectivity")) {
```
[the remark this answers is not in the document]
kellijohnson-NOAA: Thanks @Bai-Li-NOAA I have removed it.

R/create_default_parameters.R
```diff
@@ -0,0 +1,584 @@
...
+#' Create Default DnormDistribution Parameters
+#'
+#' @description
+#' Create default parameters for DnormDistribution.
+#' @param value Default value for `log_sd`.
+#' @param data An S4 object. FIMS input data.
+#' @param input_type A character. Specifies if input is data or process.
+#' @return A list of default parameters for DnormDistribution.
+#' @noRd
+create_default_DnormDistribution <- function(value = log(0.4), data, input_type = "data") {
```
kellijohnson-NOAA, line 370: The default here for value is `log(0.4)` but the only place that this function is used where value is not based on the uncertainty in the data is for the Beverton--Holt recruitment initialization. And there, the value is set to 0.1. I think that we should just be using the default in the Beverton--Holt recruitment initialization. @Bai-Li-NOAA were these values chosen with something in mind? Or, do you think it would be okay to use the default in all instances of the source code unless the values are based on data uncertainty. And, if so, what should the default be?

R/create_default_parameters.R
```diff
@@ -0,0 +1,584 @@
...
+    cli::cli_abort("The {.var modified_parameters} argument must be must be a named list.")
+  }
+
+  # Check if modified_parameters exists in current_parameters
+  missing_input <- setdiff(names(modified_parameters), names(current_parameters[["parameters"]]))
+  if (length(missing_input) > 0) {
+    cli::cli_abort(c(
+      "x" = "The following {length(missing_input)} input list{?s} from
+            {.var modified_parameters} {?is/are} missing from
+            {.var current_parameters}: {paste(missing_input, collapse = ', ')}"
```
kellijohnson-NOAA, line 523: Inside of `glue::glue()`, which is what is used by {cli} you do not have to put commas in between vector items because it will do it for you. It even puts the and 😄.

R/initialize_modules.R
```diff
@@ -0,0 +1,664 @@
...
+
+  distribution_input_names <- grep(distribution_name, names(module_input), value = TRUE)
+  for (field in distribution_fields) {
+    set_param_vector(
+      field = field, module = distribution_module,
+      module_input = module_input[distribution_input_names]
+    )
+  }
+
+  switch(distribution_type,
```
kellijohnson-NOAA, line 199: This `switch()` statement is not assigned to anything 🤔

R/initialize_modules.R
```diff
@@ -0,0 +1,664 @@
...
+      # Process distribution initialization
+      distribution_module$set_distribution_links("random_effects", linked_ids)
+    },
+    check_valid_input(input = distribution_type, valid_options = valid_distribution_types, arg_name = "distribution_type")
+  )
+
+  # Final message to confirm success
+  cli::cli_inform(c(
+    "i" = "{distribution_name} initialized successfully for {names(distribution_name)}."
+  ))
```
kellijohnson-NOAA, line 215: Do we want to have success messages for all things eventually? Where, users could turn off this level of information the same way they can turn it off in the creation of FIMSFit? I am hesitant to just have a single example of one though and not have them throughout. Perhaps we can just create a #TODO: statement at the top of this file and the top of create_default_parameters.R that identifies that these should be created in the future for all functions within the files? What do you think @Bai-Li-NOAA?

Bai-Li-NOAA: Creating a #TODO: task at the top of the file sounds good to me.

R/initialize_modules.R
Bai-Li-NOAA: It's not assigned to anything because it modifies a field of the existing `distribution_module`. Also, I am curious about your thoughts on the `initialize_distribution()`. Both `initialize_distribution()` (less function parameters to assigning values with a structure similar to other functions) and functions in `distribution_formulas.R` (more function parameters to setting up values) currently work with the high-level function `initialize_fims()`. but we only need one. Should we comment out `initialize_distribution()` temporarily and decide later when determining how to expose these lower-level functions to users?

R/initialize_modules.R
kellijohnson-NOAA: Done

R/create_default_parameters.R
Bai-Li-NOAA: That's awesome. I'll use glue more often!

R/create_default_parameters.R
Bai-Li-NOAA: Great point, I think we can use the BH recruitment initialization value here for consistency. In the future, it's fine to use the default unless the values are based on data uncertainty.

R/cli_messages.R
```diff
@@ -0,0 +1,62 @@
...
+#'
+#' @return NULL if the input is valid. Throws an error if the input is invalid.
+#' @examples
+#' \dontrun{
+#' valid_selectivity_options <- c("LogisticSelectivity", "DoubleLogisticSelectivity")
+#' check_valid_input("LogisticSelectivity", valid_selectivity_options, "form")
+#' check_valid_input("InvalidOption", valid_selectivity_options, "form")
+#' }
+#' @noRd
+check_valid_input <- function(input, valid_options, arg_name) {
```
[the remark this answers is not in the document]
Bai-Li-NOAA: Thank you! I wasn't aware of `arg_match()`, so I wrote this function. The only thing I am not sure about with `match.arg()` is the partial matching. but we can wait and see if any issues arise during the case study conversions.

R/cli_messages.R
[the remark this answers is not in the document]
Bai-Li-NOAA: It looks like a parameter is a variable defined during function definition, while an argument is a value passed to the function when it’s called, from a broader programming perspective. `check_data_arg` works for me. I was considering `check_arg_*` as a prefix in case we write more similar functions, but we can address that later since these are developer-specific functions. I agree that "data" should be renamed to "x".

R/cli_messages.R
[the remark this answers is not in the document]
kellijohnson-NOAA: Ahh 💡 partial matching. I have now changed them all to `rlang::arg_match()` which does not allow for partial matching by default and it gives users a pretty good message on how to fix things if they try partial matching.

R/cli_messages.R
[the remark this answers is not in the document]
kellijohnson-NOAA: I found out how to use the S4 class to do this rather than having our own check.
```
setMethod(
  "get_data",
  "data.frame",
  function(x) FIMSFrame(x)@data
)
```
This way, when someone passes a data frame to `get_data()` it tries to create a FIMSFrame object, which leads to the validity check for FIMSFrame to be called automatically and users get the following in return
```
> get_data(data.frame(1:10))
Error in FIMSFrame(x) : 
  Check the columns of your data, the following are missing:
data must contain 'type'
data must contain 'name'
data must contain 'datestart'
data must contain 'dateend'
data must contain 'value'
data must contain 'unit'
data must contain 'uncertainty'
data must contain 'age'
```
So, I will just ensure that whenever data is used that an accessor is used first, which will try to make a data.frame into a FIMSFrame object.

kellijohnson-NOAA:
Okay @Bai-Li-NOAA I have pushed some changes. There are still 1--2 comments that I need to deal with regarding the index and landings data but I am going to take a break. I will probably deal with those this afternoon and then merge this into dev. Please let me know if you disagree with any of the changes that I made. And, THANK YOU 🙏 👍 for all of your hard work on this and @Cole-Monnahan-NOAA also for your help in the design to get the code where it is today. Regardless of the TODO and FIXME comments in the code, this is a substantial step forward and I am excited to put it in M2 🥇.

R/distribution_formulas.R
kellijohnson-NOAA: I made a TODO statement in the code that we can revise later.

R/create_default_parameters.R
kellijohnson-NOAA: I was wrong 😞 so we will leave it as is but I have added a task to #662 to look into this.

R/initialize_modules.R
kellijohnson-NOAA: I see you left a comment in the code about this so I am going to punt it to the future but I agree that we need to decide on the best way forward before merging it into main.

PR #695, opened 2024-12-06
chore: Remove authors, email from file headers, some cleanup to cpp file header

JaneSullivan-NOAA:
# What is the feature?
This PR cleans up the head of the hpp files in inst/include. In addition to removing the author, email, and date, we also standardized the abbreviated copyright/licensing text and in general tried to clean up the header info on the files (using consistent spacing, etc.).

# Lingering Issues

- [ ] Use of `File:` vs `\file` vs not naming the file at all. It seems like the ones with `\file` get rendered in doxygen, whereas the ones with `File:` do not. Should all files have doxygen documentation?
- [ ] Not all files have a description in the header. 

No

@k-doering-NOAA @iantaylor-NOAA are awesome - thanks!

### DIFF
--- a/inst/include/common/data_object.hpp
+++ b/inst/include/common/data_object.hpp
@@ -1,31 +1,9 @@
 /*
  * File:   data_object.hpp
  *
- * Author: Matthew Supernaw
- * National Oceanic and Atmospheric Administration
- * National Marine Fisheries Service
- * Email: matthew.supernaw@noaa.gov, andrea.havron@noaa.gov
- *
- * Created on March 24, 2022, 2:37 PM
- *
  * This File is part of the NOAA, National Marine Fisheries Service
- * Fisheries Integrated Modeling System project.
- *
- * This software is a "United States Government Work" under the terms of the
- * United States Copyright Act.  It was written as part of the author's official
- * duties as a United States Government employee and thus cannot be copyrighted.
- * This software is freely available to the public for use. The National Oceanic
- * And Atmospheric Administration and the U.S. Government have not placed any
- * restriction on its use or reproduction.  Although all reasonable efforts have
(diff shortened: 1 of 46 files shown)


kellijohnson-NOAA:
@JaneSullivan-NOAA @k-doering-NOAA @iantaylor-NOAA thank you for this PR, the task turned out to be a bigger job than I was expecting. So, sorry about that. This one should have been its own issue rather than just a task in #662. The consistency that is built in this PR will help future authors have a pattern to follow, which is super helpful. THANK YOU.
Note that \file and @file should be the same thing, where doxygen translates the latter into the former. I believe it is preferred to use @ though for consistency. But `* File:    thisismyfile.cpp` does NOTHING. Also, the beginning of the comment section needs to be \** not a single star for everything to work, I think.
I went through every file in inst/include and built upon your consistency using @file @brief @details @copyright. With that, I had to add some documentation structures for undocumented macros, functions, etc. within some of the files. So, the problem of a lack of documentation, as you noted above, still remains in many places. I hope to work through some of that during the FIMS office hours next week.

iantaylor-NOAA:
Thanks @kellijohnson-NOAA for cleaning this up. It became clear to us that further design guidance was needed to make things consistent and you've provided that. Nice to see that almost half of the files have a brief description.

JaneSullivan-NOAA:
thanks for your extra work on this @kellijohnson-NOAA. we got a little hung up on whether or not we should make changes that were out of scope. In the future, is it accurate to say that all files in `inst/include` need `@file @brief @details @copyright`?

kellijohnson-NOAA:
@JaneSullivan-NOAA I would like all files to have at least @file @brief and @copyright. But, not all files need @details today. Goals for the future though.

PR #697, opened 2024-12-09
fix data_mile1.R

Bai-Li-NOAA:
# What is the feature?
* fix data_mile1.R so we can source the file from the working directory

* put on.exit() calls to the end of the file
* rename `maindir` variable to `main_dir`

* No

### DIFF
--- a/data-raw/data_mile1.R
+++ b/data-raw/data_mile1.R
@@ -38,7 +38,7 @@ library(dplyr)
 ###############################################################################
 working_dir <- getwd()
 
-maindir <- tempdir()
+main_dir <- tempdir()
 
 # Save the initial OM input using ASSAMC package (sigmaR = 0.4)
 model_input <- ASSAMC::save_initial_input()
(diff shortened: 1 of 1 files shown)


kellijohnson-NOAA:
Thanks @Bai-Li-NOAA this now works in a vanilla R environment. For some reason the on.exit() calls are still leading to weird messages in VS Code but they are benign so I will ignore them.
```
Warning message:
In file(con, "w") :
  cannot open file 'C:\Users\KELLI~1.JOH\AppData\Local\Temp\Rtmpm4nHWM/vscode-R/workspace.json': No such file or directory
```

Bai-Li-NOAA:
@kellijohnson-NOAA, thanks for pointing that out! I hadn’t tried it in VS Code before, but I just got an additional message: `cannot open the connectionwarning messages from top-level task callback 'vsc.workspace'`. Could you try running the code below before sourcing the file in VS Code and see if it fixes the issue? 
```r
removeTaskCallback("vsc.workspace")
```
If it works for you too, we can add it to the start of the file. Let me know what you think!

PR #705, opened 2024-12-16
demo vignette modification suggestions

k-doering-NOAA:
# What is the feature?
* Small changes to the vignette

* Committed small grammatical/spelling/clarity edits
* One suggestion not yet done: I see the `@data` slot is accessed directly, which I thought was not supposed to be how users access slots; instead, I think the `FIMS::get_data()` function should be used.
According to [advanced R on accessors](https://adv-r.hadley.nz/s4.html#accessors), 

>Slots should be considered an internal implementation detail: they can change without warning and user code should avoid accessing them directly. Instead, all user-accessible slots should be accompanied by a pair of accessors.

I can make this change, if we agree that it should be done!

* no


kellijohnson-NOAA:
Thank you @k-doering-NOAA for this. I have since rebased this to dev-length-wrapper-and-test and we will continue to make some edits during today's (Tuesday 2024-12-17) office hour. After which, we can merge this into dev. During the office hour we will focus on the accessor functions of FIMSFrame.

kellijohnson-NOAA:
We rocked it during the FIMS Office Hour today! We were able to clean up, document, and add to the R/fimsframe.R and R/fitfims.R files. I really cannot believe how much we got done. I restructured a few things this afternoon after the work session because I realized that it would be better to document all get_*() functions in R/fimsframe.R in their own file and likewise for m_*(). Additionally, the get_*() functions from R/fimsfit.R are also in a single file now. Then I knocked off a few trivial tasks from #662 like removing code that was commented out and fixing a few things being printed to the screen by Rcpp::Rcout().

PR #708, opened 2024-12-18
fix(initialize_modules) to return a named list

kellijohnson-NOAA:
# What is the feature?
* Allows |> without an anonymous function between `initialize_module()` and `fit_fims()`
* Also, fixes my blunder of adding eval = FALSE to two sections of the fims-demo vignette in an earlier commit.
* Fixes the documentation of what is being passed to fit_fims(input = ), where it is the output of initialize_modules().

* Uses a named list for the returned object of `initialize_modules()`

* Maybe the tests but I did not update them. @Bai-Li-NOAA let me know if I should do more.

### DIFF
--- a/R/fimsfit.R
+++ b/R/fimsfit.R
@@ -394,7 +394,7 @@ create_FIMSFit <- function(
 
 #' Fit a FIMS model (BETA)
 #'
-#' @param input Input list as returned by [create_default_parameters()].
+#' @param input Input list as returned by [initialize_fims()].
 #' @param get_sd A boolean specifying if the [TMB::sdreport()] should be
 #'   calculated?
 #' @param save_sd A logical, with the default `TRUE`, indicating whether the
--- a/R/initialize_modules.R
+++ b/R/initialize_modules.R
@@ -761,7 +761,9 @@ initialize_fims <- function(parameters, data) {
   # Set-up TMB
   CreateTMBModel()
   # Create parameter list from Rcpp modules
-  parameter_list <- list(p = get_fixed())
+  parameter_list <- list(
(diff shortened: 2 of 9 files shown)


kellijohnson-NOAA:
@Bai-Li-NOAA I will work on the tests if you think that the logic of the change is appropriate.

Bai-Li-NOAA:
@kellijohnson-NOAA, I think it’s a great idea to allow |> without needing an anonymous function between calls! The only thing I’m not totally sure about is whether we’d need to pass extra info to `fit_fims()`. I kind of remember that, for example, we can pass a version name to `fit_fims()`.

And yep, you can probably just search for `\(parameter_list)` and update the relevant tests. Thanks!

kellijohnson-NOAA:
@Bai-Li-NOAA I have the tests fixed locally but I am wondering about your comment with respect to version. The version is something that is automatically determined when running `fit_fim()` and it is a package_version object, e.g., 0.2.0, giving the semantic versioning number. The "version" that you are adding to the list is just an string that is not actually used anywhere nor are there any functions to access it. Is this string something that we think we should keep?

Bai-Li-NOAA:
@kellijohnson-NOAA, it doesn’t seem to be implemented yet. I copied the version code from the case study [here](https://github.com/NOAA-FIMS/case-studies/blob/test_list_inputs/content/test_fimsfit.R#L33), and the [documentation](https://github.com/NOAA-FIMS/FIMS/blob/dev/R/fimsfit.R#L416) suggests that additional items (e.g., path) can be added to the input. I think you are all set to go ahead and push the tests.

PR #713, opened 2025-01-03
format C++ code with clang format

ChristineStawitz-NOAA:
Auto-generated by [run-clang-format.yml][1]
  [1]: https://github.com/NOAA-FIMS/FIMS/blob/main/.github/workflows/run-clang-format.yml


kellijohnson-NOAA:
Deleting this PR because clang is moving the order of libraries included in the hpp file headers which is leading to clashes and the R package not building.

PR #719, opened 2025-01-07
fix codecov coverage calculation

Bai-Li-NOAA:
# What is the feature?
* fix call-calc-coverage workflow

* update the `DESCRIPTION` file and set `Config/testthat/parallel` to `false`. 
  This addresses the GHA run-time issue.
* skip parallel tests when running `covr::package_coverage()`, as these tests 
  don't contribute to increasing code coverage and are redundant due to 
  the reuse of wrapper functions. Integration tests are sufficient for 
  coverage calculation.
* update GHA build triggers.

* No

### DIFF
--- a/.github/workflows/call-calc-coverage.yml
+++ b/.github/workflows/call-calc-coverage.yml
@@ -13,6 +13,13 @@ on:
       - 'Rbuildignore'
       - '.gitignore'
       - 'man'
+  push:
+    branches:
+      - dev  
+  pull_request:
+    branches:
+      - main
+      - dev  
 jobs:
   call-workflow:
     uses: nmfs-fish-tools/ghactions4r/.github/workflows/calc-coverage.yml@main
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -89,7 +89,7 @@ LinkingTo:
 VignetteBuilder:
     knitr
 Config/testthat/edition: 3
-Config/testthat/parallel: true
+Config/testthat/parallel: false
 Encoding: UTF-8
 LazyData: true
 NeedsCompilation: yes
--- a/tests/testthat/test-parallel-with-snowfall-with-wrappers.R
+++ b/tests/testthat/test-parallel-with-snowfall-with-wrappers.R
@@ -8,6 +8,7 @@
 # modes. The parallel execution uses {snowfall} to parallelize the tasks across
 # multiple CPU cores.
 
+testthat::skip_on_covr()
 # Load the model comparison operating model data from the fixtures folder
 load(test_path("fixtures", "integration_test_data.RData"))
 


.github/workflows/call-calc-coverage.yml
```diff
@@ -4,15 +4,14 @@ name: call-calc_coverage
...
   workflow_dispatch:
-    paths-ignore:
-      - .devcontainer
-      - .github
-      - 'LICENSE'
-      - 'README.md'
-      - 'CONTRIBUTING.md'
-      - 'Rbuildignore'
-      - '.gitignore'
-      - 'man'
```
kellijohnson-NOAA, line 15: Shouldn't we continue to ignore these files/folders? It seems to me like we would not care if we have code coverage of the manual files and these other non-code files. Was there something that was not working with these being ignored?

.github/workflows/call-calc-coverage.yml
```diff
@@ -4,15 +4,14 @@ name: call-calc_coverage
...
+  push:
+    branches:
+      - main
```
kellijohnson-NOAA, line 9: Do we need to have push to main because there should only be PR to main? So, I guess I am thinking that the tests wouldn't take as long if we didn't double up on a PR and a push to main by just doing it on the PR. I am fine with leaving it though.

.github/workflows/call-calc-coverage.yml
```diff
@@ -4,15 +4,14 @@ name: call-calc_coverage
...
+      - dev  
+  pull_request:
+    branches:
+      - main
+      - dev
```
kellijohnson-NOAA, line 14: GREAT!

DESCRIPTION
```diff
@@ -89,7 +89,7 @@ LinkingTo:
 VignetteBuilder:
     knitr
 Config/testthat/edition: 3
-Config/testthat/parallel: true
+Config/testthat/parallel: false
```
kellijohnson-NOAA, line 92: Will this make all tests run slower? Is this a best practice change or was this change needed?

kellijohnson-NOAA (commented):
Thanks @Bai-Li-NOAA for these changes and @k-doering-NOAA for helping. I have a few questions, see comments. But, I also wanted to give more context for this PR for when I go back to it years later 🤣. GitHub actions for code coverage were running for over 6 hours leading to them failing with changes in 0.3.0.0. This is not the first time that it had happened but we did not address it before. Instead, I just turned them off for the merge into main knowing that we needed to fix them. The fix it to not have code coverage of parallel tests.

.github/workflows/call-calc-coverage.yml
Bai-Li-NOAA: Good point. It should only be a PR to main. I’ve made the changes, and @k-doering-NOAA, some changes from PR #720 might also need to be updated.

DESCRIPTION
Bai-Li-NOAA: Unfortunately, this is the actual fix for our long-running Codecov GHA issue. All the other changes are just enhancements. I should mention that the Codecov GHA still works when the codebase is smaller (about 6 months ago) and when `Config/testthat/parallel` is set to true. So, the issue may be related to the increased size of the codebase and tests. I checked packages like `devtools` and `tidyr`, and they don't run testthat in parallel either.

Yes, the tests will run slower (about 1-2 minutes longer). I think we can monitor the R CMD check runs over time. If the tests get significantly slower, we can either run benchmark tests on FIMS or move some long-running tests to a different repo. Another option is to use conditions to skip some long-running tests on local machines or run the full test suite less frequently, focusing on PRs.

Bai-Li-NOAA:
Thanks, @kellijohnson-NOAA. I'll address the questions in the comments and add more details to the PR description for future reference. I've updated the PR to target `dev`. I'm not sure if there will be any updates directly to `main` that bypass `dev`. If that happens, feel free to `edit` the PR to target `main`.

.github/workflows/call-calc-coverage.yml
kellijohnson-NOAA: Oh yeah, right. I forgot but, still, I think that we can keep these lines so the code coverage is not ran when a PR or push is made only to these files. For example, there is a GitHub action that only changes README.md and we wouldn't want to run the codecoverage test on that PR to dev. Or, if devtools::document() is ran and only changes are made to the man/*.Rd files we wouldn't want to run the coverage.

DESCRIPTION
kellijohnson-NOAA: Thanks for the explanation and this sounds fine!

.github/workflows/call-calc-coverage.yml
Bai-Li-NOAA: Got it! I just added those lines back without making any changes. I’ll leave the final edits to @k-doering-NOAA through PR #720. Basically I’m not sure where to place `paths-ignore`. Should it go under `workflow-dispatch`, `push`, or `pull`?

kellijohnson-NOAA: I think we have to repeat them under both push and pull so that they are all ignored on both PR and pushes. For the workflow dispatch i think maybe we just want to test them all?

kellijohnson-NOAA (approved):
@Bai-Li-NOAA I think this is ready. Please rebase and merge into dev whenever you are ready.

PR #721, opened 2025-01-08
Changes alphabetical = TRUE for allcontributors

kellijohnson-NOAA:
# What is the feature?
Now that the fix to allcontributors is in. Thanks @iantaylor-NOAA

* Sorts the icons alphabetically rather than in terms of the number of commits or lines of code
* PR on dev rather than main, which main is the default and nothing was declared. But we don't want any PR to main.
* Updates to use v4 of actions and v7 of create-pull-request
* Changes commit message to add a descriptor and make the tense present.

* alphabetical = TRUE
* base: dev
* "chore: Updates contributors"

* No

### DIFF
--- a/.github/workflows/call-allcontributors.yml
+++ b/.github/workflows/call-allcontributors.yml
@@ -15,7 +15,7 @@ jobs:
             pull-requests: write
         steps:
             - name: Checkout repository
-              uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


iantaylor-NOAA (approved):
Looks good. Thanks for making this happen.

PR #726, opened 2025-01-16
Fixes to v0.3.0.0 needed for the case studies

kellijohnson-NOAA:
# What is the feature?
* Adds checks for -999 in composition data before calculating the value * uncertainty
* Fixes the pluralisms to `cli::cli_abort()` in the checks

* Pluralism in cli was incorrect where a single integer needs to be passed before the {?*} to inform if it should be plural or not, instead of passing a vector like what was being done. `cli::qty()` can be used to determine the pluralism before the ? if the previous {} has nothing to do with the pluralism, which was the case here because the plural part came after the ?.
* Uses an `ifelse()` statement to determine if it should multiply value * uncertainty or value * 1 to keep the value at -999.

* Thanks to the scamp example for bringing these errors to our attention.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: FIMS
 Title: The Fisheries Integrated Modeling System
-Version: 0.3.0.0
+Version: 0.3.0.1
 Authors@R: c(
     person(c("Kelli", "F."), "Johnson", , "kelli.johnson@noaa.gov", role = c("aut", "cre"),
            comment = c(ORCID = "0000-0002-5149-451X")),
--- a/R/distribution_formulas.R
+++ b/R/distribution_formulas.R
@@ -54,7 +54,8 @@ check_distribution_validity <- function(args) {
     bad <- names(check_present[unlist(check_present)])
     abort_bullets <- c(
       abort_bullets,
-      "x" = "{.var {bad}} {?is/are} missing from {.var args}."
+      "x" = "{.var {bad}} {cli::qty(length(bad))} {?is/are} missing from
+             {.var args}."
     )
     # Abort early because not all of the necessary items were in args
     cli::cli_abort(abort_bullets)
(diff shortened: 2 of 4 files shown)


Bai-Li-NOAA (approved):
Thanks @kellijohnson-NOAA for the fixes! I have a non-blocking suggestion and a related question regarding handling uncertainty values associated with missing composition data.

**Question:** Do we want to modify the data behind the scenes or throw an error and let users adjust the uncertainty column themselves with the right inputs?

**Suggestion:** We could extend the "estimation test with age and length comp using wrappers" test in the `tests/testthat/test-integration-fims-estimation-with-wrappers.R` file by adding the following code at the end. I tested this locally, and the fix worked as expected.
```r
# Define the NA (missing value) placeholder and the index where it will be inserted
  na_value <- -999
  na_index <- 2

  # Create a copy of the original fims_data to modify
  fims_data_with_na <- fims_data
  # Extract and update the data by inserting missing values at the specified index
  updated_data <- get_data(fims_data_with_na) |>
    dplyr::mutate(
      value = ifelse(
        name == "survey1" & type %in% c("age", "length") & year == na_index,
        na_value,
        value
      )
    )
  # Replace the data with the updated data
  fims_data_with_na@data <- updated_data

  # Initialize the FIMS parameter list with the modified data
  parameter_list_with_na <- initialize_fims(
    parameters = parameters,
    data = fims_data_with_na
  )
  fit_with_na <- fit_fims(parameter_list_with_na, optimize = TRUE)

  clear()

  validate_fims(
    report = fit_with_na@report,
    sdr = fit_with_na@estimates,
    sdr_report = fit_with_na@estimates,
    om_input = om_input_list[[iter_id]],
    om_output = om_output_list[[iter_id]],
    em_input = em_input_list[[iter_id]],
    use_fimsfit = TRUE
  )
```

kellijohnson-NOAA:
`Do we want to modify the data behind the scenes or throw an error and let users adjust the uncertainty column themselves with the right inputs?`
I do not think so. It is a trivial task to just multiply by 1 behind the scenes and all that really matters is that the -999 is present.

Regarding the test, I prefer to not have an example that puts something in the data slot using @data just because I do not want someone to get ahold of that as an example on a way to do things. But, I think that adding a test is a great idea. Feel free to force push an addition so that we have everything in one commit.

Bai-Li-NOAA:
@kellijohnson-NOAA Got it, thanks for letting me know! Please go ahead and merge the current commit to main as is. I’ll work on reorganizing the test suite and cleaning up any hacky data modifications in a separate PR.

PR #734, opened 2025-01-22
chore(devcontainer.json): add packages required for R tests

Bai-Li-NOAA:
# What is the feature?
* Makes sure required R packages for tests are installed automatically, so developers don’t have to reinstall them every time they open a new Codespace.

* Added the R packages to `.devcontainer/devcontainer.json`

* No


.devcontainer/devcontainer.json
```diff
@@ -13,7 +13,7 @@
 			"packages": "make,gcc,g++,cmake,clang-tidy,clang-format,clang,doxygen,ninja-build,libxtst6,libxt6"			
 		},
 		 "ghcr.io/rocker-org/devcontainer-features/r-packages:1": {
-			"packages": "RcppEigen,dplyr,ggplot2,jsonlite,magrittr,methods,Rcpp,scales,TMB,usethis,devtools,tibble",
+			"packages": "RcppEigen,dplyr,ggplot2,jsonlite,magrittr,methods,Rcpp,scales,TMB,usethis,devtools,tibble,tidyr,mockery,snowfall",
```
kellijohnson-NOAA, line 16: We can probably get rid of magrittr and would you mind alphabetizing the list that is there?

kellijohnson-NOAA (approved):
How hard would it be to add a GitHub action that reminds us this container should be updated every time the DESCRIPTION file is updated?

MOshima-PIFSC:
> How hard would it be to add a GitHub action that reminds us this container should be updated every time the DESCRIPTION file is updated?

@kellijohnson-NOAA One option that might be of interest is the `renv` package. It can automatically detect what packages are used in the DESCRIPTION file and create a project library based on that. Then you can just install renv in the container and after the codespace is built, call a `setup.r` script that will restore the r environment. You can also use it in a GHA workflow to set up an r environment based on the project library it creates (like for tests and rendering websites). Nicholas and I recently used this in a project we were working on and it worked pretty smoothly. I think it also helps speed up the set up time as well. Here's an article about how to use if for[ package development](https://rstudio.github.io/renv/articles/packages.html) and I'm happy to talk more about it with you.

Bai-Li-NOAA:
@kellijohnson-NOAA, @MOshima-PIFSC's suggestion sounds promising! I’ve copied the conversation over to [Discussions](https://github.com/orgs/NOAA-FIMS/discussions/735#discussion-7858825) so it doesn’t get lost in the PR.

PR #737, opened 2025-01-23
add examples for calculating parameter lengths

Bai-Li-NOAA:
# What is the feature?
* add examples for calculating parameter lengths using `purrr::map_vec()`

* update documentation for `update_parameters()` and include a `dontrun` example
* add an example in the `fims-demo` vignette

* No

### DIFF
--- a/R/create_default_parameters.R
+++ b/R/create_default_parameters.R
@@ -29,11 +29,15 @@
 #' }
 #' @export
 #' @seealso
+#' * [FIMSFrame()]
 #' * [update_parameters()]
 #' @examples
 #' \dontrun{
+#' # Load the example dataset and create a FIMS data frame
 #' data("data1")
 #' fims_frame <- FIMSFrame(data1)
+#'
+#' # Define fleets specifications for each fleet in the example dataset
 #' fleet1 <- survey1 <- list(
 #'   selectivity = list(form = "LogisticSelectivity"),
 #'   data_distribution = c(
@@ -41,17 +45,12 @@
 #'     AgeComp = "DmultinomDistribution"
 #'   )
 #' )
-#' fleet2 <- list(
-#'   selectivity = list(form = "DoubleLogisticSelectivity"),
-#'   data_distribution = c(
-#'     Index = "DlnormDistribution",
-#'     AgeComp = "DmultinomDistribution",
-#'     LengthComp = "DmultinomDistribution"
-#'   )
-#' )
+#'
+#' # Create a list of default parameters given the fleet specifications set up
+#' # above, recruitment, growth, and maturity specifications
 #' default_parameters <- fims_frame |>
 #'   create_default_parameters(
-#'     fleets = list(fleet1 = fleet1, fleet2 = fleet2, survey1 = survey1),
+#'     fleets = list(fleet1 = fleet1, survey1 = survey1),
 #'     recruitment = list(
 #'       form = "BevertonHoltRecruitment",
 #'       process_distribution = c(log_devs = "DnormDistribution")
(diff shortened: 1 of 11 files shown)


R/create_default_parameters.R
```diff
@@ -602,6 +602,57 @@ create_default_recruitment <- function(
...
+#'   )
+#' )
+#' fleet2 <- list(
+#'   selectivity = list(form = "DoubleLogisticSelectivity"),
+#'   data_distribution = c(
+#'     Index = "DlnormDistribution",
+#'     AgeComp = "DmultinomDistribution",
+#'     LengthComp = "DmultinomDistribution"
+#'   )
+#' )
```
kellijohnson-NOAA, line 626: fleet2 does not actually exist in fims_frame so when I try to run this example I get
```
Error in `create_default_parameters()`:
ℹ The name of the fleets for selectivity settings must match the fleet names present in the `data`.
✖ The following 1 fleet name is missing from the data: fleet2.
```
I think we can delete these lines and simplify line 631 to just have fleet1 and survey1.

R/create_default_parameters.R
```diff
@@ -602,6 +602,57 @@ create_default_recruitment <- function(
...
+#'            0.343685540, 0.314171227, 0.308026829,
+#'            0.431745298, 0.328030899, 0.499675368
+#'          ))
+#'        )
+#'      )
+#'   )
+#' # purrr::map_vec() can be used to compare the length of adjusted parameter vectors with defaults
+#' default_fleet1 <- purrr::map_vec(default_parameters[["parameters"]][["fleet1"]], \(x) length(x))
+#' updated_fleet1 <- purrr::map_vec(updated_parameters[["parameters"]][["fleet1"]], \(x) length(x))
+#' }
```
kellijohnson-NOAA, line 655: What do you think about also adding the following code:
```
purrr::map_df(
  updated_parameters[["parameters"]], \(x) purrr::map_vec(x, length),
  .id = "module"
) |>
  tibble::column_to_rownames(var = "module") |>
  t()
```
which shows something like this
image

Bai-Li-NOAA:
> Quick question, how come you chose to have this example wrapped in don't run? I personally would not have thought twice about it until I noticed that the code coverage went down for the PR which confused me because we are not adding any functions to be tested.

I wrapped it in `dontrun` to avoid increasing the R CMD Check runtime. I’ve read some unofficial advice about keeping R CMD Check under 10 minutes, and we’re already pretty close to that. 

Also, some examples are just for demo syntax and might not run successfully, like the fleet2 example. @kellijohnson-NOAA, **Your comment makes me wonder if we should just remove fleet2 from all the examples in `R/create_default_parameters.R`.** It was originally included to show that we could have two different logistic forms for different fleets. Another option is to keep it but add some documentation and maybe specific conditional execution for fleet2, depending on whether you think it’s worth keeping as an example for `create_default_parameters()`.

As for the coverage drop, it happens when we add more lines of code without adding new tests. That’s why it’s good to review the tests folder regularly to make sure we’re covering everything important instead of just chasing the coverage percentage.

R/create_default_parameters.R
Bai-Li-NOAA: It looks nice! I will add it to both the example and the vignette.

kellijohnson-NOAA:
Thanks @Bai-Li-NOAA for the comments. I made a suggestion regarding fleet2 but in general I think that all examples should be executable if the user wants to. Please leave the example as dontrun for now but make the suggested changes to fleet2 and we can talk about using dontrun and dontetst in terms of the entire package and testing time on Tuesday at Code Club.

R/create_default_parameters.R
Bai-Li-NOAA: Sounds good. I updated the examples for `create_default_parameters()` based on the suggestions for fleet2. For the examples for `update_parameters()`, I removed fleet2 entirely and focused on different ways to checking parameter lengths.

R/create_default_parameters.R
Bai-Li-NOAA: Done!

kellijohnson-NOAA:
@Bai-Li-NOAA I am still working on the creation of `create_default_fleet_list()` and integrating its use within the package but here is an initial commit so you can see the function. Let me know if you have any thoughts. I will have more time to work on this when hake is over in a couple of days.

kellijohnson-NOAA:
@Bai-Li-NOAA I took out my addition of create_default_fleet_list because I am worried that the fleet structure might change with the CPUE stuff so I will make an issue that this still needs to be done. As soon as the tests pass I will merge this into dev.

kellijohnson-NOAA:
Thanks @Bai-Li-NOAA I assigned you to https://github.com/NOAA-FIMS/FIMS/issues/774

Bai-Li-NOAA:
@kellijohnson-NOAA, the bug has been fixed in commit d88444542bd5d6588278f3759fc9885af9b09166, and all checks have passed. Let me know if you have any questions.

Issue #736, opened 2025-01-23
[Bug:] run-clang-format failed on dev


Bai-Li-NOAA:
### Describe the bug

The `run-clang-format` GHA failed to run after merging changes from a feature branch to the dev branch. It failed when creating a pull request. Below is the error message:
```yaml
Attempting creation of pull request
  A pull request already exists for NOAA-FIMS:format-c++-code
  Fetching existing pull request
  Attempting update of pull request
  Error: Cannot read properties of undefined (reading 'number')
```
Is it related to the failed PR for formatting C++ code with clang format [here](https://github.com/NOAA-FIMS/FIMS/pull/731)?
In addition, I saw a warning after the GHA ran `Peter-evans/create-pull-request@v4`:
```yaml
(node:2089) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. 
Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
```

### Describe how to reproduce the bug

See GHA log [here](https://github.com/NOAA-FIMS/FIMS/actions/runs/12917396991/job/36061812369#step:5:1)

### Which OS are you seeing the problem on?

Linux

### Which version of FIMS are you seeing the problem on?

0.3.0.1 from the dev branch

### Additional context

We may need to update the build triggers (e.g., run on pull requests) if we want to keep dev passing all the checks and catch failed GHA before merging the code into dev.

k-doering-NOAA:
@Bai-Li-NOAA , I think the issue is that the branch name where the styling occurs is *always* the same, so if there is already a PR open to main, it will error if it tries to make a PR to a feature branch as well.

Actually, workflows like [doc and style in ghactions4r](https://github.com/nmfs-fish-tools/ghactions4r/blob/main/.github/workflows/doc-and-style-r.yml#L132) have this same issue.

I think the solution would be to figure out how to pass different names of branches depending on where the PR should be merged to. I'm not quite sure how to do this, but it seems like it should be possible?

If we can figure this out, it would be good to update doc and style in ghactions4r, too!

k-doering-NOAA:
@k-doering-NOAA and @Bai-Li-NOAA , do I need to work on this?

kellijohnson-NOAA:
[...]

PR #742, opened 2025-02-10
turn off clang format on dev

k-doering-NOAA:
due to #736

# What is the feature?
* Stop running clang format on dev, until a real solution for #736 is implemented.

* remove the dev build trigger

*


Bai-Li-NOAA (approved):
Looks good to me!

PR #753, opened 2025-02-27
test: clean up R tests and add test template

Bai-Li-NOAA:
# What is the feature?
* delete some R tests and add a test template for developers

* remove tests that are not part of the GHA workflows
* add a test template with three key testing criteria (see `inst/templates/testthat_template.R`)
* add use_testthat_template() helper function for developers (see `R/use_template.R`)
* add R tests to validate use_testthat_template() (see `tests/testthat/test-use-testthat-template.R`)
* update the EWAAgrowth Rcpp tests to follow test template

* No

**Note:** We can now generate a bookdown report of {testthat} results using [{testdown}](https://github.com/ThinkR-open/testdown) and [{lozen}](https://github.com/ThinkR-open/lozen). After installing both packages, run `lozen::build_pkgdown_with_reports(reports = "testdown")` to automatically add the {testdown} report to the report tab of the FIMS {pkgdown} website. If we go ahead with this, I'll need to work with @k-doering-NOAA to add some extra code to `build-pkgdown.yml` from {ghactions4r}.

**TODO:**
- [ ] create an issue with one checkbox for each test file, so the team can work on refactoring the tests using the template


Bai-Li-NOAA:
The HTMLl files for the {testdown} report can be downloaded here:
[html.zip](https://github.com/user-attachments/files/19010723/html.zip)

kellijohnson-NOAA:
Just an FYI that I am reviewing this now and will finish by the EOD.

kellijohnson-NOAA (approved):
@Bai-Li-NOAA this is brilliant. Thank you for the PR and using the usethis package. I think this is a great addition to the package. I added two commits. The first one is just some clean up. The second commit includes changes, which is why I split them in two. So, feel free to reject the second commit. I used the `data` argument of the {usethis} function to pass the test name as a function that then populates text in the test file. I think that the template should only generate one test file for a function. Later, we can add tests for other functions within the file. Feel free to augment any of my changes or reject the second commit by just deleting the commit and force pushing. Again, thank you!

R/use_template.R
```diff
@@ -47,7 +47,8 @@ use_testthat_template <- function(name) {
   # TODO: add the ability to also pass the arguments for the function or find
   #       them within the code base and ensure that the template includes the
   #       necessary structure for each input argument
-  path <- file.path("tests", "testthat", paste0("test-", name, ".R"))
+  file_name <- gsub("_", "-", name)
```
kellijohnson-NOAA, line 50: I am wondering if we want all the underscores to be changed to dashes? I am fine with either but this means that some current file names would need to be changed, e.g., test-create_default_parameters.R. I think that we should just leave it however the user puts it in that way there is not a hidden mismatch.

Bai-Li-NOAA:
Thanks @kellijohnson-NOAA for using the `data` argument! Great idea! I made a small change to update the test file name to use a hyphen (`-`) instead of an underscore(`_`) (e.g., `test-my-function.R` for the R function `my_function`). I’ll merge the changes into the `dev` branch once all checks pass.

R/use_template.R
Bai-Li-NOAA: I left a comment before seeing your message here. I think some files with underscores in their names need to be updated during the test cleanup to maintain consistency. According to the `use_testthat_template()` documentation, we should use function names with underscores. If developers accidentally use a dash in the name argument, it will still generate the correct file name, but they’ll need to manually correct the function names in the tests to ensure they run successfully.

Issue #756, opened 2025-03-04
[Developer Issue]: combine initialize_length_comp() and initialize_age_comp() into a single function


kellijohnson-NOAA:
### Description

Inside initialize_modules.R both `initialize_length_comp()` and `initialize_age_comp()` exist but they have almost duplicate code. The only difference is the name of the module. I think they can be rewritten to use just one function where the type of the module could be passed. Additional code would need to be changed where the functions are called.

kellijohnson-NOAA:
Thank you @mciechanumich for working on this. As noted in the ro-hackathon-2025 feel free to open a PR even if the work is not finished and you just want some help or you would like to pass it off to one of the core developers to finish. We appreciate your efforts.

PR #759, opened 2025-03-07
refactor: combine two initialize functions (#756)

mciechanumich:
Combine functions initialize_length_comp() and initialize_age_comp() into one function initialize_comp(). Also then replaced calls to the old functions with calls to the new function.

test: Also added several tests for the new function, including tests that verify the function correctly returns error messages on error.

Still TODO is to remove both old original (now unused) functions from the code.

# What is the feature?
* Completed most of issue#756 by combining the functions initialize_length_comp() and initialize_age_comp() into one function initialize_comp(). This change reduces code duplication.

* I have created a data table at the top of the new function. The table contains all of the variables that the later code needed to implement either AgeComp or LengthComp. Additional future compositions simply by adding rows to the table (and also adding a module initialization that the end of the function). The code selects the proper row based on the function user's parameter called "type".
* Still need to remove the old original functions initialize_length_comp() and initialize_age_comp().
* Still need to come up with a better way to initialize the module. Right now, the code switches on "type" parameter. R probably has a better way of doing this that can eliminate the remain code duplication.

* I am not sure where else in the project the code uses the old original functions.

### DIFF
--- a/R/initialize_modules.R
+++ b/R/initialize_modules.R
@@ -482,6 +482,104 @@ initialize_index <- function(data, fleet_name) {
   return(module)
 }
 
+# Issue #756 combine initialize_length_comp() and initialize_age_comp()
+# into a single function
+#' Initialize any composition module
+#'
+#' @description
+#' Initializes a composition module for a specific fleet,
+#' setting the composition data for the fleet over time.
+#' @inheritParams initialize_module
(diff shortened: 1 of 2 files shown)


mciechanumich:
Here is the rest of the change. This part removes the duplicate code in the if-else branch.

PR #768, opened 2025-03-19
refactor(initialize_comp): Combines functions

kellijohnson-NOAA:
# What is the feature?
* Combine functions `initialize_length_comp()` and `initialize_age_comp()` into one function `initialize_comp()`.

* A list is present at the top of `initialize_comp()` that sets the appropriate downstream objects and functions based on the type, where type is an input argument.

* I don't think so because everything is embedded in wrapper functions.


Bai-Li-NOAA (approved):
Looks great! Many thanks to @mciechanumich and @kellijohnson-NOAA for taking care of this TODO task.

PR #789, opened 2025-04-01
fix(cmake_minimum_required): Updates to 4.0

kellijohnson-NOAA:
# What is the feature?
* Updates CMake version to recent.

* Changes minimum version from 4.0.
* Updates link to documentation

* 🤷‍♀️

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 # Top level CMakeLists.txt: setting up top level project dependencies
 # and options. CMake Referecen documentation can be found here:
-# https://cmake.org/cmake/help/v3.22/
-cmake_minimum_required(VERSION 3.14)
+# https://cmake.org/cmake/help/v4.0/
+cmake_minimum_required(VERSION 3.14...4.0)
 project(FIMS
   CXX
 ) # CXX is the language name
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 # Top level CMakeLists.txt: setting up top level project dependencies
-# and options. CMake Referecen documentation can be found here:
+# and options. CMake Reference documentation can be found here:
 # https://cmake.org/cmake/help/v4.0/
 cmake_minimum_required(VERSION 3.14...4.0)
 project(FIMS
@@ -14,7 +14,7 @@ include(FetchContent)
 # Set up GoogleTest
 FetchContent_Declare(
   googletest
-  URL https://github.com/google/googletest/archive/refs/tags/release-1.11.0.zip
+  URL https://github.com/google/googletest/archive/refs/tags/v1.16.0.zip
 )
 
 # For Windows: Prevent overriding the parent project's compiler/linker


kellijohnson-NOAA:
We are running actions on [ubuntu-latest](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md) which has CMake 3.31.6. Not sure what to do because this is less than 3.5. @k-doering-NOAA do you have any ideas?

kellijohnson-NOAA:
Others have noted a break in the CI as well in https://github.com/actions/runner-images/issues/11926#issuecomment-2770326140. If you are so inclined, you can give this a 👍 in the issue over on actions/runner-images 😃!

Bai-Li-NOAA:
I noticed a warning from CMake in the gtest GHA run earlier:

```
CMake Warning (dev) at /usr/local/share/cmake-4.0/Modules/FetchContent.cmake:1373 (message):
  The DOWNLOAD_EXTRACT_TIMESTAMP option was not given and policy CMP0135 is
  not set.  The policy's OLD behavior will be used.  When using a URL
  download, the timestamps of extracted files should preferably be that of
  the time of extraction, otherwise code that depends on the extracted
  contents might not be rebuilt if the URL changes.  The OLD behavior
  preserves the timestamps from the archive instead, but this is usually not
  what you want.  Update your project to the NEW behavior or specify the
  DOWNLOAD_EXTRACT_TIMESTAMP option with a value of true to avoid this
  robustness issue.
```

One possible fix is to use an if statement to configure the CMake policy. I found a potential solution [here](https://github.com/ethereum/solidity/pull/13429/files) and the CMake documentation on CMP0135 [here](https://cmake.org/cmake/help/latest/policy/CMP0135.html).

kellijohnson-NOAA:
@Bai-Li-NOAA I have tried a few things for this but I think the error regarding cmake is coming from the CMakeList.txt stored in the [Google Test Archive](https://github.com/google/googletest/archive/refs/tags/release-1.11.0.zip) that we reference. Where the current version is [v1.16.0](https://github.com/google/googletest/releases/tag/v1.16.0) not 1.11.0 that we are using. So, I upgraded to the newest version of google tests but I am getting different errors now from the three failing actions that I have not had time to debug. I will be invested in the all hands for an hour, then come back to this.

kellijohnson-NOAA:
@Bai-Li-NOAA several have complained about the upgrade to cmake 4.0 and they are actively working on it to go back to CMake 3.31.6. I allow for a range of CMake but I am not sure if we should leave the range? Additionally, I upgraded to v 1.12.0 for the google tests and now things appear to be passing. Should I keep incrementing through the google checks until we find the most recent version that it is successful on?

Bai-Li-NOAA (approved):

Bai-Li-NOAA:
🎉  Nice work, @kellijohnson-NOAA! 🎉 I’m not super familiar with dependency management, so I don’t have a strong answer to your questions. But if everything is passing, it seems logical to keep incrementing until we hit the latest stable version.
I wonder if we could set up a version matrix in GHA (kind of like use_full_build_matrix in {ghactions4r}) to automatically test a range of CMake and Google Test versions. That might help us figure out the best stable combo. 
That said, it’s not the highest priority. We can merge the current changes for now, and maybe tackle the version matrix setup in a code club meeting with @k-doering-NOAA’s help.

k-doering-NOAA:
The matrix approach sounds good! If we are going to support a range, it makes sense that we should be testing the range, too.

kellijohnson-NOAA:
Well we were using google test version 1.11 and when I originally tried to increment to 1.16 I did not realize that the URL was different 🤕 so the test failed b/c it couldn't find the URL. Now that I have the URLs figured out, it doesn't matter what version of google test we are using as long as it is newer than 1.11.

See below for how the ULRs changed.
```
https://github.com/google/googletest/archive/refs/tags/release-1.12.0.zip
https://github.com/google/googletest/archive/refs/tags/release-1.12.1.zip
https://github.com/google/googletest/archive/refs/tags/v1.13.0.zip
https://github.com/google/googletest/archive/refs/tags/v1.14.0.zip
https://github.com/google/googletest/archive/refs/tags/v1.15.0.zip
https://github.com/google/googletest/archive/refs/tags/v1.15.2.zip
https://github.com/google/googletest/archive/refs/tags/v1.16.0.zip
```

As far as a next step for this, I think that we do not need to use a matrix approach for testing multiple versions of CMake because all we are specifying is the range of versions that we accept and the GitHub runner actually dictates what version is used. So, as written the tests will fail when CMake is greater than 4.0. The question that I do have is how do we ensure that we are using the most up-to-date version of google test 🤷‍♀️. I started following [their repository](https://github.com/google/googletest/) and I will be automatically notified via email when a new release comes out. But that will only be helpful if I remember what I am supposed to do when I get that email. What would be more helpful is if we could automatically create an issue when they publish a release. I created a new issue for this.

PR #798, opened 2025-04-06
wrap of constant variables

Srajald:
# What is the feature?
* Ensuring type safety in templated code by wrapping numeric constants (e.g., `-999`) with the `Type()` constructor.
* This is particularly important for compatibility with automatic differentiation (AD) types such as `CppAD::AD<double>`, which are commonly used in FIMS.

* Identified instances where numeric constants were directly compared with templated variables of type `Type`.
* Wrapped these constants with the `Type()` function (e.g., changed `-999` to `Type(-999)`) to ensure correct type matching and avoid implicit conversion issues.
* Focused changes on modules interacting with fleets and recruitment where IDs are assigned and compared.

* This change only affects the current repo's use of templated logic in C++.
* It does not affect external repos but improves robustness and correctness of this codebase, especially when compiled with TMB or similar AD libraries.

### DIFF
--- a/inst/include/interface/rcpp/rcpp_objects/rcpp_data.hpp
+++ b/inst/include/interface/rcpp/rcpp_objects/rcpp_data.hpp
@@ -160,8 +160,8 @@ class AgeCompDataInterface : public DataInterfaceBase {
   template <typename Type>
   bool add_to_fims_tmb_internal() {
     std::shared_ptr<fims_data_object::DataObject<Type>> age_comp_data =
-        std::make_shared<fims_data_object::DataObject<Type>>(this->ymax,
-                                                             this->amax);
+        std::make_shared<fims_data_object::DataObject<Type>>(Type(this->ymax),
+                                                             Type(this->amax));
 
     age_comp_data->id = this->id;
     for (int y = 0; y < ymax; y++) {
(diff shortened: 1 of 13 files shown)


inst/include/interface/rcpp/rcpp_objects/rcpp_data.hpp
```diff
@@ -160,8 +160,8 @@ class AgeCompDataInterface : public DataInterfaceBase {
   template <typename Type>
   bool add_to_fims_tmb_internal() {
     std::shared_ptr<fims_data_object::DataObject<Type>> age_comp_data =
-        std::make_shared<fims_data_object::DataObject<Type>>(this->ymax,
-                                                             this->amax);
+        std::make_shared<fims_data_object::DataObject<Type>>(Type(this->ymax),
+                                                             Type(this->amax));
```
kellijohnson-NOAA, line 164: There are five instances in this file where `this->*max` was wrapped in `Type(*)` erroneously. Or at least, it is causing the model to fail to compile and if I remove the changes in this file then FIMS compiles just fine. Thus, we either need documentation for why this change was made and how to get FIMS to compile with this change or the file needs to be reverted.

inst/include/population_dynamics/recruitment/functors/recruitment_base.hpp
```diff
@@ -72,14 +72,14 @@ struct RecruitmentBase : public fims_model_object::FIMSObject<Type> {
...
-    Type sum = 0.0;
+    Type sum = Type(0.0);
 
     for (size_t i = 0; i < this->log_recruit_devs.size(); i++) {
       sum += this->log_recruit_devs[i];
     }
 
     for (size_t i = 0; i < this->log_recruit_devs.size(); i++) {
-      this->log_recruit_devs[i] -= sum / (this->log_recruit_devs.size());
+      this->log_recruit_devs[i] -= sum / Type(this->log_recruit_devs.size());
```
kellijohnson-NOAA, line 82: I am uncertain why this line was changed to be wrapped in `Type(*)`

kellijohnson-NOAA (changes requested):
Thank you @Srajald for this Pull Request. I have some proposed changes on two files where I am unsure why Type was used for `this->*` where the changes in rcpp_data.hpp lead to the model failing to compile.

inst/include/population_dynamics/recruitment/functors/recruitment_base.hpp
Srajald: done ma'am

PR #799, opened 2025-04-07
Dev cpue selected at age

nathanvaughan-NOAA:
# What is the feature?
* This branch implements the ability to assign observed landings and indices of abundance to a fleet independently.
* A fleet can now include either separately, both together, or neither. 
* Expected indices default to q=1 when no observed index data are present.
* Expected landings default to log_F=-200 when no observed landings are present.
* Composition data has now been made independent of landings so F_mort does not impact the selected number at age.
* Both landings and indices can now also be specified as observed in terms of weight or numbers.

* Fleet landings inputs were created by replicating existing index values throughout the codebase under a new landings label. 
* The population module now includes separate calculations for derived landings and indices with default values of log_q=1 and log_F=-200 being fixed when no observed data are available. 
* Fleets were modified to include a boolean toggle for observed landings and index in weight or number. 
* Both landings and index in weight and number are always calculated and the boolean is used to assign one of these as the expected value.
* Number at age and weight at age are now calculated independently for landings, indices, and composition data.
* Composition data work at the population level where composition number at age is equal to population number at age multiplied by selectivity at age.

* This pull request includes significant refactoring throughout the codebase and likely effects other repos and ongoing development. 
* This pull request was rebased onto the latest dev branch before submitting.
* This pull request is also currently failing tests due to the new structure slightly changing estimates. This will require particular review and potential refactoring of the test data objects.

# Key areas for review: 

Implementation method (Tim could you focus on these changes they are the logic changes the rest is just all the interface changes needed)
* Primary structural changes are found in Population.hpp lines 390:516 and Fleet.hpp lines 237:338

Testing results errors and discussion
* Testing errors are found in test-integration-fims-estimation-without-wrappers lines 244,245,249 and test-integration-fims-estimation-with-wrappers lines 250,251,255 due to fleet1 age and length comp miss matches
* Other tests were modified to adjust to expect_lt from expect equal with an e-9 threshold to ignore rounding error differences that were generally in the e-12 to e-15 range.
* Tests in helper-integration-tests-validation.R could use modification to function for single fleets rather than all simultaneously because now that landings and index are separated it is only possible to test the fishing fleet because those are the first rows selected.

Interface changes throughout
* I updated the reporting and interface files throughout to include the new values as they appeared to be needed. It could be worth discussing the logic of what should and shouldn't be included in each of these areas, as it was not immediately clear what was needed where
* Model.hpp
* rcpp_interface.hpp
* rcpp_fleet.hpp
* Particularly in the json reporting and add_to_fims_tmb_internal() sections


R/create_default_parameters.R
```diff
@@ -348,54 +348,103 @@ create_default_fleet <- function(fleets,
...
   # Determine default fleet parameters based on types of data present
-  # FIXME: allow for a fleet to have both landings and index data
-  process_default <- if ("landings" %in% data_types_present) {
-    list(
-      log_Fmort.value = log(rep(0.00001, get_n_years(data))),
-      log_Fmort.estimated = TRUE
-    )
-  } else {
-    list(
+  if ("index" %in% data_types_present &&
```
timjmiller, line 355: seems undesirable to use "index" for one category and "Index" for another.

R/create_default_parameters.R
```diff
@@ -348,54 +348,103 @@ create_default_fleet <- function(fleets,
...
+
+    index_distribution_default <- NULL
   }
 
-  names(process_default) <- paste0("Fleet.", names(process_default))
+  names(q_default) <- paste0("Fleet.", names(q_default))
 
-  # Create index distribution defaults
-  index_distribution <- fleets[[fleet_name]][["data_distribution"]]["Index"]
+  if ("landings" %in% data_types_present &&
```
timjmiller, line 398: similar to index/Index above: landings/Landings

R/create_default_parameters.R
```diff
@@ -348,54 +348,103 @@ create_default_fleet <- function(fleets,
...
-      value = index_uncertainty,
-      input_type = "data",
-      data = data
-    ),
-    "DlnormDistribution" = create_default_DlnormDistribution(
-      value = index_uncertainty,
-      input_type = "data",
-      data = data
+  } else {
+    log_Fmort_default <- list(
```
timjmiller, line 431: I hope F in this situation is not used at all in mortality/survival calculations. Seems it could be inefficient.

R/data1.R
```diff
@@ -27,7 +27,7 @@
 #'     viable units for the composition data, where the former is the preferred
 #'     unit of measurement.}
 #'   \item{uncertainty}{A real value providing a measurement of uncertainty
-#'     for value. For catches and survey indices of abundance this should be
+#'     for value. For landingses and survey indices of abundance this should be
```
timjmiller, line 30: typo

timjmiller: I'm guessing changing from catch to landings is because discards will be modeled separately

R/initialize_modules.R
```diff
@@ -74,45 +74,48 @@ initialize_module <- function(parameters, data, module_name) {
 
   if (module_class_name == "Fleet") {
     module_fields <- setdiff(module_fields, c(
-      "log_expected_index",
-      "proportion_catch_numbers_at_age"
+      "log_index_expected",
+      "log_landings_expected",
+      "composition_numbers_at_age",
+      "proportion_numbers_at_age",
+      "observed_index_in_weight",
```
timjmiller, line 81: if indices are only allowed to be in biomass/weight. That would exclude many/all assessments in the NEUS.

R/initialize_modules.R
```diff
@@ -74,45 +74,48 @@ initialize_module <- function(parameters, data, module_name) {
...
-        "estimate_q"
-      ))
-    } else {
-      module_fields <- setdiff(module_fields, c(
-        "log_Fmort"
-      ))
-    }
+    # if (!"index" %in% fleet_types) {
+    #   module_fields <- setdiff(module_fields, c(
+    #     "log_q",
```
timjmiller, line 92: might want to consider a logit option for q to force q to be below some value. Scale is difficult to estimate sometimes.

inst/include/interface/rcpp/rcpp_interface.hpp
```diff
@@ -632,25 +635,31 @@ RCPP_MODULE(fims) {
...
-      .field("proportion_catch_numbers_at_age", &FleetInterface::proportion_catch_numbers_at_age)
-      .field("proportion_catch_numbers_at_length", &FleetInterface::proportion_catch_numbers_at_length)
-      .field("age_length_conversion_matrix", &FleetInterface::age_length_conversion_matrix)
-      .method("SetObservedAgeCompData", &FleetInterface::SetObservedAgeCompData)
+      .field("observed_landings_in_weight", &FleetInterface::observed_landings_in_weight)
+      .field("observed_index_in_weight", &FleetInterface::observed_index_in_weight)
+      .field("log_index_expected", &FleetInterface::log_index_expected)
+      .field("log_landings_expected", &FleetInterface::log_landings_expected)
+      .field("composition_numbers_at_age", &FleetInterface::composition_numbers_at_age)
+      .field("composition_numbers_at_length", &FleetInterface::composition_numbers_at_length)
```
timjmiller, line 650: what is difference between composition and proportion?

inst/include/common/model.hpp
```diff
@@ -67,18 +67,28 @@ namespace fims_model {
             // vector< vector<Type> > creates a nested vector structure where
             // each vector can be a different dimension. Does not work with ADREPORT
             // fleets
-            vector<vector<Type> > exp_index(n_fleets);
-            vector<vector<Type> > exp_catch(n_fleets);
-            vector<vector<Type> > cnaa(n_fleets);
-            vector<vector<Type> > cwaa(n_fleets);
-            vector<vector<Type> > cnal(n_fleets);
-            vector<vector<Type> > pcnaa(n_fleets);
-            vector<vector<Type> > pcnal(n_fleets);
```
timjmiller, line 76: if the difference between proportion and composition is whether it is data or the predicted value, should probably be consistent with index/exp_index and landings/exp_landings.

timjmiller: sorry should be index_exp and landings_exp, I think

R/create_default_parameters.R
nathanvaughan-NOAA: Good point Tim, I was just following the existing format for this. Maybe we can brainstorm better naming on Thursday.

R/create_default_parameters.R
nathanvaughan-NOAA: This F is used which is why the tiny value is defaulted. We can't use zero because of the log transformations but this way the use doesn't have to specify if the fleet is fishing or survey. We didn't have any other good ideas for how to set this up because we cannot have if forks based on parameter values or it breaks the autodiff chain.

R/data1.R
nathanvaughan-NOAA: Yep, the issue was also that it was already landings in the data input frames and catches everywhere else which was confusing. This is a wonderful autoreplace catch with landings error :)

inst/include/interface/rcpp/rcpp_objects/rcpp_fleet.hpp
```diff
@@ -125,46 +132,82 @@ class FleetInterface : public FleetInterfaceBase {
...
+   */
+  ParameterVector age_to_length_conversion;
+  /**
+   * @brief Should index scaling parameter (q) be estimated? 
+   * The default is false.
    */
   SharedBoolean estimate_q = false;
   /**
-   * @brief Is catchability (q) a random effect? The default is false.
+   * @brief Is landingsability (q) a random effect? The default is false.
```
timjmiller, line 171: this nomenclature seems odd.

R/initialize_modules.R
nathanvaughan-NOAA: They were only allowed in weight before. Now this and the landings equivalent code allow either to be in weight or numbers.

inst/include/interface/rcpp/rcpp_interface.hpp
nathanvaughan-NOAA: Composition is the raw numbers, currently NAA*selectivity, while proportion is scaled to sum to one. There is commented out code in the fleet calcs that tries to rescale composition to match the input data but I took it out while debugging and haven't tried putting it back yet. Comparing observed to expected comps was really hard previously because the expected values were in proportion and raw were in total landings not in comp units.

inst/include/common/model.hpp
nathanvaughan-NOAA: This one is confusing because the comp data can either be in real numbers or proportions. Internally the proportion is what is used for the expected likelihood calculation but on the user end this is hard to compare because we don't report the observed values as proportion. We could refactor to report observed and expected in both units I guess. I did try to refactor code throughout to have expected values be clearer but there is still more to do.

inst/include/interface/rcpp/rcpp_objects/rcpp_fleet.hpp
nathanvaughan-NOAA: Hahahaha I was confused by landingsability too! didn't realize it was my own replace function that caused it!! We can go back to catchability.

inst/include/population_dynamics/fleet/fleet.hpp
```diff
@@ -142,81 +171,134 @@ namespace fims_popdy {
...
 
             for (size_t year = 0; year < this->nyears; year++) {
                 this->Fmort[year] = fims_math::exp(this->log_Fmort[year]);
             }
+
+            // derived quantities
+            // landings
+            std::fill(landings_weight.begin(), landings_weight.end(),
+                    0); /**<model landings in weight*/
+            std::fill(landings_numbers.begin(), landings_numbers.end(),
```
timjmiller, line 186: are the weight and numbers also "expected"? If so, should everything that is a derived quantity have _expected for consistency?

nathanvaughan-NOAA: So at one point I did make all the derived quantities expected but then that also got confusing. for landings and index you actually have three values one in weight , one in numbers, and one expected. The evaluate landings and evaluate index functions check which units the observed data are in and then use either weight or numbers to fill the expected vector. I figured this allows users to always be able to get expected values from one place but also see the results in numbers or weight.

inst/include/population_dynamics/fleet/fleet.hpp
```diff
@@ -142,81 +171,134 @@ namespace fims_popdy {
...
+         * Evaluate the proportion of landings numbers at age.
          */
         void evaluate_age_comp() {
             for (size_t y = 0; y < this->nyears; y++) {
                 Type sum = 0.0;
+                //Type sum_obs = 0.0;
                 for (size_t a = 0; a < this->nages; a++) {
                     size_t i_age_year = y * this->nages + a;
-                    sum += this->catch_numbers_at_age[i_age_year];
+                    sum += this->composition_numbers_at_age[i_age_year];
```
timjmiller, line 243: so, I think i understand now that composition is in numbers and proportions sum to 1, but we always call proportions "comps" so this will probably be confusing to others (not just me :)). why is "composition" needed in the name? Perhaps to denote difference between the population naa and the values scaled by selectivity, q, etc.? why not landings_numbers_at_age_expected and index_numbers_at_age_expected and the like?

inst/include/population_dynamics/fleet/fleet.hpp
```diff
@@ -142,81 +171,134 @@ namespace fims_popdy {
...
-                            this->catch_numbers_at_age[i_age_year] * 
-                            this->age_length_conversion_matrix[i_length_age];
+                            this->composition_numbers_at_length[i_length_year] += 
+                            this->composition_numbers_at_age[i_age_year] * 
+                            this->age_to_length_conversion[i_length_age];
                         }
-                        sum += this->catch_numbers_at_length[i_length_year];
+                        sum += this->composition_numbers_at_length[i_length_year];
+                        
+                        //if( this->fleet_observed_lengthcomp_data_id_m != -999) {
```
timjmiller, line 285: can these commented out sections be deleted?

R/create_default_parameters.R
timjmiller: but I thought there is an "is_index" boolean somewhere?

inst/include/population_dynamics/fleet/fleet.hpp
nathanvaughan-NOAA: Getting away from landings at age is why there are now comps. We never actually have landings at age just a composition subsample that has the same selectivity. The goal here was to make the landings, index, and composition calculations independent from each other so that fleets can have any one without the others. That is why it is now composition_numbers_at_age.

inst/include/population_dynamics/population/population.hpp
```diff
@@ -120,7 +119,8 @@ namespace fims_popdy {
 
             // size all the vectors to length of nages
             nfleets = fleets.size();
-            expected_catch.resize(nyears * nfleets);
+            total_landings_weight.resize(nyears);
```
timjmiller, line 122: is there another type of landings besides "total"? or there also a fleet_landings_weight, etc.?

R/create_default_parameters.R
nathanvaughan-NOAA: We got rid of that input so the fleet could use both. Andrea has a refactor for estimation that replaces booleans with text options so we could replace the landings_in_weight boolean with landings_units and have weight, number, or none options then skip the landings calculations if none is entered?

inst/include/population_dynamics/population/population.hpp
nathanvaughan-NOAA: There are landings by weight by fleet in the fleet module. This is total landings summed across all fleets. We could actually remove this, I'm not sure why landings are double tracked in the population and fleet modules. Seems like it just adds opportunity for errors.

tests/gtest/test_population_Index.cpp
```diff
@@ -18,22 +18,21 @@ namespace
...
             int index_yf = year * population.nfleets + fleet_;
             
             // Currently q is not a vector and not changing over years. 
             // When testing time varying q, better to test entire vector.
             // If not possible to test entire vector, test middle or second to last 
             // than earlier years (collapses to mean in early years)
-            expected_index[index_yf] += population.numbers_at_age[i_age_year]*
+            index_expected[index_yf] += population.numbers_at_age[i_age_year]*
                                         population.fleets[fleet_]->q[0]*
                                         population.fleets[fleet_]->selectivity->evaluate(population.ages[age])*
```
timjmiller, line 29: does this mean the index is only on Jan 1?

inst/include/interface/rcpp/rcpp_objects/rcpp_fleet.hpp
timjmiller: I often "over-replace" too. I noticed it happens elsewhere in the code so probably need to do a global search for landingsability.

timjmiller: also "landingses"

tests/testthat/helper-integration-tests-validation.R
```diff
@@ -125,25 +125,37 @@ validate_fims <- function(
...
   # Expected survey number at age
   validate_error(
     expected = c(
-      t(om_output[["L.age"]][["fleet1"]]),
-      t(om_output[["survey_age_comp"]][["survey1"]])
+      t(om_output[["L.age"]][["fleet1"]])#,
+      #t(om_output[["survey_age_comp"]][["survey1"]])
     ),
-    param_name = "CNAA",
+    param_name = "ExpectedLandingsAtAge",
```
timjmiller, line 158: Expected in front here?

tests/gtest/test_population_Index.cpp
nathanvaughan-NOAA: Yes at the moment there is no capacity for timing in FIMS or index or any other data types.

tests/testthat/helper-integration-tests-validation.R
nathanvaughan-NOAA: Hmm this is one of those situations were we need to decide when expected should be included. There are no landings at age observations to compare to now that comps are independent so this should maybe just be LandingsAtAge

tests/testthat/test-integration-fims-estimation-with-wrappers.R
```diff
@@ -78,44 +78,45 @@ test_that("deterministic test of fims", {
   # F (fixed at initial "true" values)
   expect_equal(report[["F_mort"]][[1]], om_output_list[[iter_id]][["f"]])
 
-  # Expected catch
-  fims_index <- report[["exp_index"]]
+  # Expected landings
+  fims_landings <- report[["landings_exp"]]
```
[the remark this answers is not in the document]
nathanvaughan-NOAA: I aggree that the shortening is not consistent throughout. We have some things that are written out but then a lot of reporting that is shortened. We should discuss a general plan on Thursday.

timjmiller (commented):
I don't see any problems. Just made a few suggestions that may or may not be useful.

nathanvaughan-NOAA:
I figured out the issue. It turns out it wasn't just a rounding error. The problem was that fleet comps use the fraction removed from the population (F/Z)*naa*(1-exp(-Z)), while the survey uses q and selectivity directly q*selectivity*naa. Because the fishery uses the exponential, there is a slight non-linear difference in the composition results. I have modified the code now to use the fishery calculation for comps if there are any observed landings and revert to the survey for comps otherwise. This solved the test errors once I fixed some g-test naming/reporting issues. I did try the robust multinomial modification (which made a slight change but obviously didn't fix the issue) and the code is still in there but I reverted it to adding zero adjustment for now as it also changed values slightly messing up other tests.

nathanvaughan-NOAA:
Just rebased this onto dev, so I think this is ready to merge. Assuming we add needing more tests into the existing code coverage issue.

R/initialize_modules.R
```diff
@@ -74,45 +74,50 @@ initialize_module <- function(parameters, data, module_name) {
...
+    #     "random_q",
+    #     "estimate_q"
+    #   ))
+    # }
+
+    # if (!"landings" %in% fleet_types) {
+    #   module_fields <- setdiff(module_fields, c(
+    #     "log_Fmort"
+    #   ))
+    # }
```
kellijohnson-NOAA, line 104: @nathanvaughan-NOAA did you mean to leave this commented out? Can we remove the code?

nathanvaughan-NOAA: Yep we can delete it, it's not needed.

kellijohnson-NOAA:
An update on this PR. I have rebase all of the files except the integration tests with the wrappers in testthat. I found a few mistakes in how the json output was being added in rcpp_fleet.hpp that I have commented out for now. You can see my progress in dev-cpue. I have to put this down for a bit but I will come back to it later today.

nathanvaughan-NOAA:
I rebased as well and updated the rcpp_fleet and it looks like it is all working again. It was just a bracketing issue. At least now I think I understand how the JSON reporting works.

kellijohnson-NOAA:
Thanks @nathanvaughan-NOAA 😃 I will squash and merge these commits into dev even though the ubuntu quit on us again.

PR #813, opened 2025-04-19
workflow for pr-commands

Srajald:
# What is the feature?
* Added GitHub Actions support for triggering automated documentation and styling of R code by commenting `/document` or `/style` on a pull request.
* This mimics the behavior used in the `scales` repo and aligns with best practices for code maintainability in R.

# How have you implemented the solution?
* Created a new workflow file `.github/workflows/pr-commands.yml` that listens to PR comments and triggers based on commands (`/document`, `/style`).
* The new workflow leverages the existing `call-doc-and-style-r.yml` file, which already uses `devtools::document()` and `styler::style_active_package()` via a reusable workflow.
* Ensured only comments from repo members or owners trigger the workflow for security.
* Automatically commits and pushes changes (e.g., updated `man/` files or re-styled `.R` files) back to the PR branch.

# Does the PR impact any other area of the project, maybe another repo?
* No direct impact on the R code or other repositories.
* This enhancement improves developer workflow and PR hygiene within **NOAA-FIMS/FIMS**, making it easier to maintain up-to-date documentation and consistent styling across contributors.

PR #829, opened 2025-04-28
workflow for pr-commands updated

Srajald:
# What is the feature?
Implement GitHub Actions that automatically document and style R code when a comment command is posted on a Pull Request (PR).
Supports new slash commands /document-r and /style-r to be more explicit, since the project contains both R and C++ code.

Created .github/workflows/pr-commands.yml that listens for issue_comment events.
Configured the workflow to detect /document-r and /style-r commands from authorized users.
On /document-r, it runs roxygen2::roxygenise() to generate documentation.
On /style-r, it runs styler::style_pkg() to format R code.
After running the action, it automatically commits and pushes the changes back to the PR branch.

No, this PR only affects the GitHub Actions workflows in this repository.
It does not modify any source code (R or C++).
It only adds automation for easier R code maintenance via PR commands.

### DIFF
--- a/.github/workflows/pr-commands.yml
+++ b/.github/workflows/pr-commands.yml
@@ -1,0 +1,90 @@
+# .github/workflows/pr-commands.yml
+
+on:
+    issue_comment:
+      types: [created]
+  
+name: pr-commands
+
+permissions: read-all
+
+jobs:
+    document:
+        if: ${{ github.event.issue.pull_request && 
+                (github.event.comment.author_association == 'MEMBER' || github.event.comment.author_association == 'OWNER') &&
+                startsWith(github.event.comment.body, '/document-r') }}
+        name: document
+        runs-on: ubuntu-latest
+        env:
+            GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
+        permissions:
+            contents: write
+        steps:
+            - uses: actions/checkout@v4
+    
+            - uses: r-lib/actions/pr-fetch@v2
+              with:
+                repo-token: ${{ secrets.GITHUB_TOKEN }}
+    
+            - uses: r-lib/actions/setup-r@v2
+              with:
+                use-public-rspm: true
+    
+            - uses: r-lib/actions/setup-r-dependencies@v2
+              with:
+                extra-packages: any::roxygen2
+                needs: pr-document
+    
+            - name: Document
+              run: roxygen2::roxygenise()
+              shell: Rscript {0}
+    
+            - name: commit
+              run: |
+                git config --local user.name "$GITHUB_ACTOR"
+                git config --local user.email "$GITHUB_ACTOR@users.noreply.github.com"
+                git add man/* NAMESPACE
+                git commit -m 'Document'
+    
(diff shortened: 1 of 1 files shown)


kellijohnson-NOAA:
@k-doering-NOAA and @Bai-Li-NOAA what do y'all think about this PR compared to #813? @Srajald has made some great proposals here. The first PR allowed you to use one of two keywords to trigger the reusable workflow where this PR makes the same two keywords implement just one portion of the workflow, though it is no longer using the reusable workflow and the GitHub action is wrote out. Or should ask @Srajald to kindly revise this PR to use the reusable workflow but keep the two key word triggers in this PR? I personally think that it will be fine to have just a single word both style and document as long it is clear that the styling and documenting pertains to the R code, maybe just style-r. Thoughts 🤔 💭 💭

Bai-Li-NOAA:
This is really cool! I don’t have a strong preference for the keyword we use, but I agree that including `-r` makes sense.

For reusable workflows, could we make `.github/workflows/pr-commands.yml` a reusable workflow across the NOAA-FIMS org, or include it in {ghactions4r}?

I'm curious who's responsible for posting the PR comment command (PR author or code reviewer). How can we make sure it's not forgotten during code review? We might want a safeguard to ensure it’s not forgotten during code review. One idea is to add a GHA that performs a dry-run check and fails the PR if the code isn’t documented or styled. For styling, we could use `styler::style_pkg(dry = "fail")`. I’m not sure how to do a dry-run for documentation, but this got me looking into [{precommit}](https://lorenzwalthert.github.io/precommit/index.html). It looks like we could use [pre-commit.ci](https://pre-commit.ci/) to run hooks as CI checks: if we want to enforce passing hooks on PRs (and auto-fix trivial problems like styling) even if the committer does not have a local installation. Feel free to ignore the ideas if I’ve misunderstood anything!

kellijohnson-NOAA:
Thank you for this PR. I have essentially merged this PR with the previous PR (#813) into a single commit (e71f62ed) that uses `\style-r` as the command to a single preexisting workflow to help decrease the amount of code that we need to maintain.

@Bai-Li-NOAA regarding who is in charge of running it ... No one needs to be in charge because we still have GitHub actions that will style things if no one does them. This is meant to be more of a convenience feature.

k-doering-NOAA:
Hi, sorry I'm so late on this! I think this is how tidyverse authors have decided to do their styling, as this is the [pr-commands example that the workflow comes from](https://github.com/r-lib/actions/blob/v2-branch/examples/pr-commands.yaml).

I think it is great if people would like to use this option, as an alternative to the more automated option - the only downside is needing to maintain more github actions.

I think we should explore adding it to {ghactions4r} if it is likely to be used by many of our NOAA Fisheries repositories - otherwise, keeping it in FIMS or a FIMS reusable workflow makes sense!

kellijohnson-NOAA:
Thank you so much @Srajald for this PR. We were able to combine your two PRs into a single commit (i.e., e71f62ed) that is now in dev and will be in main shortly. So, I am going to close this PR. Thank you and please feel free to continue contributing to NOAA-FIMS/FIMS we value your contributions.

Issue #818, opened 2025-04-25
[Developer Issue]: reorganize tests to reduce R CMD Check execution time


Bai-Li-NOAA:
### Description

The current R test suite takes too long to run during R CMD check on GitHub Actions, leading to recent failures (see #761, #817). Also, my bad for putting the test data prep script in a helper file, it ends up running during both `devtools::load_all()` and `devtools::test()`, which is pretty annoying for developers. I will clean this up so the tests run at the right time and don’t take forever.

kellijohnson-NOAA:
I think that this might also be related to the inability to build FIMS on R-universe 🤷‍♀ 
```
      5. │       └─testthat:::test_files_setup_state(...)
      6. │         └─testthat::source_test_helpers(".", env)
      7. │           └─testthat::source_dir(path, "^helper.*\\.[rR]$", env = env, wrap = FALSE)
      8. │             └─base::lapply(...)
      9. │               └─testthat (local) FUN(X[[i]], ...)
     10. │                 └─testthat::source_file(path, env = env, chdir = chdir, wrap = wrap)
     11. │                   ├─base::withCallingHandlers(...)
     12. │                   └─base::eval(exprs, env)
     13. │                     └─base::eval(exprs, env)
     14. │                       └─FIMS:::setup_and_run_FIMS_with_wrappers(...) at tests/testthat/helper-integration-tests-setup-run.R:144:1
     15. │                         └─Rcpp (local) clear() at tests/testthat/helper-integration-tests-setup-function.R:480:3
```
It looks like `clear()` cannot be run because it is not tied to `FIMS::clear()` but I am not positive about that.

Bai-Li-NOAA:
Thanks, @kellijohnson-NOAA. I wasn’t able to reproduce the R-universe error, but I’ll keep exploring it using the `main-r-universe-check` branch during the week of May 12. 

PR #833 resolves the issue that prevented FIMS from running during `devtools::load_all()`.

Bai-Li-NOAA:
@kellijohnson-NOAA I think we can close this issue. The tests have been reorganized, and FIMS is no longer run during `devtools::load_all()`. I opened a [new issue](https://github.com/NOAA-FIMS/FIMS/issues/875) to track the FIMS check failure on R-universe. I looked into it but couldn’t reproduce the error locally. It might be helpful to schedule a debugging session with more people to investigate further.

Issue #832, opened 2025-05-01
`check_fims()` needs mocked package structure for testing


kellijohnson-NOAA:
The one downside of `check_fims()` is that it is not really tested. In the future we could add a mock package of FIMS to a temporary directory and test the functionality of it with just a few files.

For example,
```
current_directory <- getwd()
temporary_package_directory <- file.path(tempdir(), "FIMS")
usethis::create_package(temporary_package_directory)
on.exit(setwd(current_directory), add = TRUE)
on.exit(unlink(temporary_package_directory, recursive = TRUE), add = TRUE)
setwd(temporary_package_directory)
# Create inst/include/population to make some make cpp files to be able to run a C++ test
# Create a vignette
usethis::use_vignette()
...
check_results <- check_fims()
```
Then run some tests on the output.

_Originally posted by @kellijohnson-NOAA in https://github.com/NOAA-FIMS/FIMS/issues/831#issuecomment-2844891276_

PR #833, opened 2025-05-01
Improve R testing workflow and fix bugs

Bai-Li-NOAA:
# What is the feature?
* Address issue #818 

* Refactor(helper-integration-tests-setup-run.R)
  * Put original code in a single function `prepare_test_data()` to prevent FIMS from running during `devtools::load_all()'.
  * `devtools::test()` will run FIMS models to prepare test data.
  * If developer wants to run a single test such as `devtools::test(filter = "distribution-formulas")`, it will skip FIMS model runs since it doesn't need test data from the helper function.
* Add `FIMS:::run_r_unit_tests()` to run R unit tests only.
* Add `FIMS:::run_integration_tests()` to run integration tests only.
* Add `FIMS:::remove_test_data()` to remove test data from the `tests/testthat/fixtures` directory if changes in the core code affect FIMS intput or output.
* Fix(verify_fims_deterministic()): replace nonexistent `om_output_list[[iter_id]]` with `om_output`, etc.
* Fix(test-use-testthat-template.R): forcibly remove the output folder.

*

### DIFF
--- a/R/setup_gtest.R
+++ b/R/setup_gtest.R
@@ -87,3 +87,87 @@ run_gtest <- function(...) {
   system(paste("ctest --test-dir build", ...))
   TRUE
 }
+
+#' Run R integration tests
+#'
+#' Intended for developers to run the R integration tests.
+#' @keywords developer
+#' @examples 
+#' \dontrun{
+#' run_r_integration_tests()
+#' }
+run_r_integration_tests <- function() {
+  devtools::test(filter = "integration")
+  devtools::test(filter = "parallel")
+}
+
+#' Run R unit tests
+#'
+#' Intended for developers to run the R unit tests.
+#' @keywords developer
+#' @examples
+#' \dontrun{
+#' run_r_unit_tests()
+#' }
+run_r_unit_tests <- function() {
+  # List all R files in the tests/testthat directory with the pattern 
+  # "test-integration*.R"
+  integration_tests <- list.files(
+    path = file.path("tests", "testthat"),
+    pattern = "^test-integration.*\\.R$"
+  )
+  # List parallel tests
+  integration_tests <- c(integration_tests, list.files(
+    path = file.path("tests", "testthat"),
+    pattern = "^test-parallel-.*\\.R$"
+  ))
+  # List all files in the tests/testthat directory that are not integation tests 
+  all_tests <- list.files(
+    path = file.path("tests", "testthat"),
+    pattern = "^test-.*\\.R$"
+  )
+  # Exclude integration tests
+  unit_tests <- setdiff(all_tests, integration_tests)
+  # Remove "test-" and ".R" from the file names
+  test_files <- gsub("^test-|\\.R$", "", unit_tests)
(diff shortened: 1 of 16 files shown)


Bai-Li-NOAA:
Developer functions will be tested through task #832

R/setup_gtest.R
```diff
@@ -87,3 +87,87 @@ run_gtest <- function(...) {
...
+#' 
+#' @keywords developer
+#' @examples \dontrun{
+#' remove_test_data()
+#' }
+remove_test_data <- function() {
+  # List the data files starting with "integration-"
+  data_to_keep <- list.files(
+    path = file.path("tests", "testthat", "fixtures"),
+    pattern = "^integration-.*\\.R$"
```
kellijohnson-NOAA, line 157: blocking: This search is for .R files not for data files.

Also, I think you can makes line 155--165 a negation test like
```
# Exclude any file from fixtures that starts with an "i" and
# include all files that have an extension that starts with R
list.files(
  path = file.path("tests", "testthat", "fixtures"),
  pattern = "^[^[i]]*\\.R"
)
```

tests/testthat/helper-integration-tests-setup-run.R
```diff
@@ -1,319 +1,332 @@
...
+#' data with missing values. The integration_test_data_components.RData is 
+#' generated after running the script R/data1.R.
+#' 
+#' @return None. The function saves the generated datasets as RDS files in the
+#' specified directory.
+#' @examples
+# \dontrun{
+#   prepare_test_data()
+# }
+prepare_test_data <- function(){
```
kellijohnson-NOAA, line 17: nit: there should be a space between () and {

R/setup_gtest.R
kellijohnson-NOAA: For all of these searches should we be using `testthat::test_path()` rather than `file.path()`?

tests/testthat/test-fimsfit.R
```diff
@@ -9,10 +9,12 @@
 # FIMSFit ----
 ## Setup ----
 # Load the test data from an RDS file containing the fitted model estimates
+if (!file.exists(test_path("fixtures", "fit_age_length_comp.RDS"))) {
```
kellijohnson-NOAA, line 12: All of these if statements are pretty brilliant so the tests don't take forever because it is just repeating the same process. Great work 🥇

kellijohnson-NOAA (changes requested):
Great job Bai. It is so nice to have `load_all()` working quickly again when there are no changes to the C++ code. I made a few comments regarding regular expressions that might be helpful to reduce the number of lines of code. 
I did find one blocking change that needs to happen regarding searching for `.R$` which excludes RData files. The rest of the comments can be ignored if you would like.

tests/testthat/helper-integration-tests-setup-run.R
kellijohnson-NOAA: Fixed by using styler in f3042bc7

PR #837, opened 2025-05-08
feat(random effects): Adds random effects

kellijohnson-NOAA:
# What is the feature?
* random effects for the power user

* add new virtual function to recruitment_base, evaluate_process
* add new children classes, LogDevs and LogR that handle the error of recruitment differently
* add rcpp interface for LogDevs and LogR
* expose new interfaces to R
* delete conditional statement from CalculateRecruitment in population and replace with new evaluate_process virtual function
* remove estimate_log_devs from interface
* reorder random effects call in model.hpp
* refactor distribution input to be pointers
* add struct for DensityElementObjects and call elements from distributions
* clear pointers to DistributionElementObjects
* setup recruitment process pointer inside the recruitment module
* add both log_devs and log_r options to helper tests setup
* change booleans to string that take either constant, fixed_effects, or random_effects
* update rcpp_objects to use new interface
* create RegisterRandomEffectName and get_random_names() that is exposed to R

* the advanced case study will need to be updated to use this rebased branch that it was originally based on
* wrapper functions will be needed for the random effects to expand beyond the power user


inst/include/distributions/functors/density_components_base.hpp
```diff
@@ -20,25 +20,84 @@
...
+    if(this->input_type == "data"){
+      return observed_values->at(i);
+    }
+    if(this->input_type == "random_effects"){
+      return (*re)[i];
+    }
+    if(this->input_type == "prior"){
+      return (*(priors[i]))[0];
+    }
+    return x[i];
```
kellijohnson-NOAA, line 52: Is there a way to have an informative abort if the user input something other than the three options?

inst/include/distributions/functors/lognormal_lpdf.hpp
```diff
@@ -90,12 +85,20 @@ namespace fims_distributions
...
+                  }
+                  if(this->input_type == "random_effects"){
+                    (*this->re)[i] = 
+                      fims_math::exp(rnorm(this->get_expected(i),
+                      fims_math::exp(log_sd.get_force_scalar(i))));
+                  }
+                  if(this->input_type == "prior"){
+                    (*(this->priors[i]))[0] = 
+                      fims_math::exp(rnorm(this->get_expected(i),
+                      fims_math::exp(log_sd.get_force_scalar(i))));
```
kellijohnson-NOAA, line 101: The code coverage indicates that we do not have tests for priors, is that true?

inst/include/distributions/functors/density_components_base.hpp
msupernaw: I suggest adding an error to the FIMS log. If FIMS log is set to throw on error, it will abort immediately.

inst/include/distributions/functors/lognormal_lpdf.hpp
Andrea-Havron-NOAA: Current tests set up normal priors, not lognormal. The C++ test [here](https://github.com/NOAA-FIMS/FIMS/blob/dev-random-effects-rebase/tests/gtest/test_info_setup_priors.cpp) can be expanded to include a lognormal prior.

kellijohnson-NOAA: @Andrea-Havron-NOAA can you add a test for the lognormal as you indicated above and also increase the comments in the testing files to indicate what they are doing? After that, we can squash all of the commits into one again and merge this into dev.

inst/include/distributions/functors/density_components_base.hpp
kellijohnson-NOAA: I don't like having to require the user to set "throw on error" to make things happen. I would rather the user have the same consistent experience of informative error messages to the console or to a log file without crashing. But, this is a conversation for somewhere outside of this PR as it is a global change that should happen.

inst/include/distributions/functors/lognormal_lpdf.hpp
msupernaw: [this](https://github.com/NOAA-FIMS/FIMS/blob/dev-random-effects-rebase/inst/include/distributions/functors/density_components_base.hpp#L56) method should return Type& rather than Type if a reference is the desired return type, otherwise it's returning a copy.

inst/include/distributions/functors/density_components_base.hpp
msupernaw: I think there should also be a dimension check before the vector access.

Andrea-Havron-NOAA: DImension checks are in the density functors (eg. [here](https://github.com/NOAA-FIMS/FIMS/blob/17d4479f812c300044725694a35302393b12c2f2/inst/include/distributions/functors/normal_lpdf.hpp#L48)), but I can see it being more efficient to check dimesions in the get_observed and get_expected functions.

msupernaw: That works, but the question is do we just throw invalid_argument, or just log the error, or both?

kellijohnson-NOAA: Error early to 🛑 the model if the model won't work.

PR #885, opened 2025-06-30
Fix pr commands.yml

k-doering-NOAA:
# What is the feature?
* correct reusable workflow path in pr-commands github acction
* change the permissions to match those required by [doc-and-style](https://github.com/nmfs-ost/ghactions4r/blob/main/.github/workflows/doc-and-style-r.yml)

* made changes to github action yml
* Note: I wasn't sure if this should be merged to main or dev.
* Note: I have not tested this to ensure it works, but I can set up a separate repository to try it out if needed!

* should fix so commands will work and error msgs wont be sent to those commenting

### DIFF
--- a/.github/workflows/pr-commands.yml
+++ b/.github/workflows/pr-commands.yml
@@ -13,5 +13,6 @@ jobs:
             (github.event.comment.author_association == 'MEMBER' || github.event.comment.author_association == 'OWNER') &&
             (startsWith(github.event.comment.body, '/document-r') || startsWith(github.event.comment.body, '/style-r')) }}
     name: Document and Style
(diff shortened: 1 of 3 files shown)


kellijohnson-NOAA:
@k-doering-NOAA thanks! I think that you are right for this PR to go to main because it is essentially a hot fix to the previous version, and then we will rebase dev onto main to bring it in. One thing that we need though is an increment in the version in the DESCRIPTION file and a bullet in the NEWS file. I can do those two things before I merge it in.

k-doering-NOAA:
Thanks @kellijohnson-NOAA !

PR #892, opened 2025-07-02
Update rcpp_interface_base.hpp

msupernaw:
# What is the feature?
* Added handling for mismatched  ParameterVector(Rcpp::NumericVector x, size_t size) arguments

*   ParameterVector(Rcpp::NumericVector x, size_t size){
    this->id_m = ParameterVector::id_g++;
    this->storage_m = std::make_shared<std::vector<Parameter> >();
      this->resize(size);
      if(x.size() != size){
          FIMS_WARNING_LOG("Mismatch in constructor ParameterVector(Rcpp::NumericVector x, size_t size). x.size() = " + fims::to_string(x.size())+ " != "+ fims::to_string(size) +". resulting vector will be initialized to " + fims::to_string(x[0]) +" for all values");
          for (size_t i = 0; i < size; i++) {
              storage_m->at(i).initial_value_m = x[0];
          }
      }else{
          for (size_t i = 0; i < size; i++) {
              storage_m->at(i).initial_value_m = x[i];
          }
      }
  }

* Test now pass.

### DIFF
--- a/inst/include/interface/rcpp/rcpp_objects/rcpp_interface_base.hpp
+++ b/inst/include/interface/rcpp/rcpp_objects/rcpp_interface_base.hpp
@@ -198,12 +198,16 @@ public:
    * @param size The number of elements to copy over.
    */
   ParameterVector(Rcpp::NumericVector x, size_t size){
-    this->id_m = ParameterVector::id_g++;
-    this->storage_m = std::make_shared<std::vector<Parameter> >();
-    this->resize(size);
-    for (size_t i = 0; i < size; i++) {
-      storage_m->at(i).initial_value_m = x[i];
-    }
+      if(x.size() < size){
+          throw std::invalid_argument("Error in call to ParameterVector(Rcpp::NumericVector x, size_t size): x.size() < size argument.");
+      }else{
+          this->id_m = ParameterVector::id_g++;
+          this->storage_m = std::make_shared<std::vector<Parameter> >();
+          this->storage_m->resize(size);
+          for (size_t i = 0; i < size; i++) {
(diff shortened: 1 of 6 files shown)


kellijohnson-NOAA:
Thanks @msupernaw for this! I am curious though about the use of the size argument, I thought that you were going to remove the second argument because we use $resize() in R to size the vector.

inst/include/interface/rcpp/rcpp_objects/rcpp_interface_base.hpp
```diff
@@ -200,10 +200,17 @@ class ParameterVector{
...
-    this->resize(size);
-    for (size_t i = 0; i < size; i++) {
-      storage_m->at(i).initial_value_m = x[i];
-    }
+      this->resize(size);
+      if(x.size() != size){
+          FIMS_WARNING_LOG("Mismatch in constructor ParameterVector(Rcpp::NumericVector x, size_t size). x.size() = " + fims::to_string(x.size())+ " != "+ fims::to_string(size) +". resulting vector will be initialized to " + fims::to_string(x[0]) +" for all values");
+          for (size_t i = 0; i < size; i++) {
+              storage_m->at(i).initial_value_m = x[0];
+          }
```
kellijohnson-NOAA, line 208: @msupernaw are you sure that we want to do a warning and initialize to zero? Or, should we do a full stop with an error? I think that I prefer the latter but I could be convinced otherwise.

msupernaw: @kellijohnson-NOAA no I'm not sure! I'd rather throw an error. I'll make the change!

Issue #900, opened 2025-07-08
[Bug:] Derived quantities need _ separator


kellijohnson-NOAA:
### Describe the bug

The following derived quantities need underscores between the words:
indexexpected, indexnumbers, indexweight, landingsexpected, landingsnumbers, landingsweight
I am guessing that this is happening because there is a space in the json file between the words and the space is getting taken out somewhere, where we really should have an underscore rather than a space.

### Describe how to reproduce the bug

```
get_estimates(deterministic_results) |> dplyr::count(module_name,label)|>as.data.frame()
```

### Which version of FIMS are you seeing the problem on?

0.5.0

PR #908, opened 2025-07-09
fix: replace spaces with underscores in derived quantity names

abhinav-1305:
# What is the feature?
* fixes: #900 

* Replace space-separated names with underscore-separated names in `Fleet to_json()` method to prevent concatenated labels like "indexexpected" and "landingsweight" in get_estimates() output.

*


abhinav-1305:
please review - @kellijohnson-NOAA

Issue #866, opened 2025-06-16
[Developer Issue]: Silence model output in R tests


Bai-Li-NOAA:
### Description

The current R tests produce too much output, making it hard to see {testthat} feedback or instructions (e.g., how to update a snapshot). We should suppress model output during test runs.

kellijohnson-NOAA:
I think that we might be able to use the logic in `is_fims_verbose()` to turn off some of the output, i.e.,
```
options("rlib_message_verbosity" = "quiet")
```
and what is not suppressed, we can create arguments to check to see what verbosity level there is to know if output should be printed to the screen or not.

Bai-Li-NOAA:
I changed `silent = FALSE` to `silent = TRUE` since it was the only instance with `silent = FALSE`.

PR #922, opened 2025-07-17
silence test output

Bai-Li-NOAA:
# What is the feature?
* address issue #866 and silence model output in R tests

* set verbosity to `quiet` in `tests/testthat.R` to suppress test output
  when running `devtools::test()` locally.
* set verbosity to `quiet` in `tests/testthat/helper-aaa-quiet-test-output.R`
  to suppress output when running `testthat::test_package()` locally.
  This mirrors the "Show testthat output" (`test_check()`) step in the
  `call-r-cmd-check` GitHub Action workflow.
* set `silent = TRUE` in `TMB::MakeADFun()` calls to
  suppress TMB output during tests.
* use `suppressMessages()` to silence output from fims_frame tests and
  testthat template tests.
* don't use `Rcerr` to print a message to the console when the ParameterVector
  index is out of range, since `throw` already handles the error output.

* NA

**Notes for Reviewers:**
- The [GHA logs from this PR](https://github.com/NOAA-FIMS/FIMS/actions/runs/16357556058/job/46219245761#step:12:44) show that the R tests no longer produce model output. For comparison, earlier runs included extensive model output in the logs [here](https://github.com/NOAA-FIMS/FIMS/actions/runs/16332052562/job/46195620901#step:12:37).
- I'm currently having issues running `devtools::test()` locally on a Codespace (Linux), so I haven’t been able to fully verify the behavior. It would be very helpful if the reviewer could run `devtools::test()`, `devtools::check()`, and `testthat::test_package()` locally on a Windows machine to confirm that no model output is printed to the console.

### DIFF
--- a/inst/include/interface/rcpp/rcpp_objects/rcpp_interface_base.hpp
+++ b/inst/include/interface/rcpp/rcpp_objects/rcpp_interface_base.hpp
@@ -252,7 +252,6 @@ class ParameterVector {
   SEXP at(R_xlen_t pos) {
     if (static_cast<size_t>(pos) == 0 ||
         static_cast<size_t>(pos) > this->storage_m->size()) {
-      Rcpp::Rcerr << "ParameterVector: Index out of range.\n";
       throw std::invalid_argument("ParameterVector: Index out of range");
       FIMS_ERROR_LOG(fims::to_string(pos) + "!<" +
                      fims::to_string(this->size()));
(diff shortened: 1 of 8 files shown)


kellijohnson-NOAA:
@Bai-Li-NOAA I looked at the Windows R CMD check from the GitHub actions and the output is gone, as well as I ran the tests with an installed version of your branch and there is no output.

PR #935, opened 2025-07-24
Fix TMBad windows bug

Andrea-Havron-NOAA:
# What is the feature?
* This PR fixes a bug in the code that was causing rcmdcheck to fail when running TMBad in Windows

* The population derived quantity, `unfished_biomass` was not being initialized. This was causing an instability that was only picked up by Windows when running TMBad.

* No


kellijohnson-NOAA:
Great find 🔍 @Andrea-Havron-NOAA. I am wondering if you mind adding a comment on how you found the 🐛? Did you use vagrind, gdb, or some other method?

PR #946, opened 2025-07-31
Fix basic math calls when TMB defined

Andrea-Havron-NOAA:
# What is the feature?
* With the change to TMBad, models are sometimes crashing when calling `exp()` and other basic math functions. This has to do with how `exp()` was referenceing the global namespace. This worked when CppAD was defined because CppAD has an `exp()` and other basic math functions in the global namespace while TMBad does not. This PR fixes the issue by involking an Argument-Dependent Lookup instead of relying on the global namespace.

* removed `using ::exp;` and replaced with `using std::exp;`
* removed extra function defining std::double type

This solution now initially looks for the type in std. If it is not found, it will search in other libraries for specialized versions of the function with the specific type defined. 

* no

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: FIMS
 Title: The Fisheries Integrated Modeling System
-Version: 0.6.2
+Version: 0.6.3
 Authors@R: c(
     person(c("Kelli", "F."), "Johnson", , "kelli.johnson@noaa.gov", role = c("aut", "cre"),
            comment = c(ORCID = "0000-0002-5149-451X")),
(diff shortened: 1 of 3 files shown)


Andrea-Havron-NOAA:
How I tracked this down: I ran into an R session crash when trying to run tests in the surplus production branch. I used gdb backtrace to isolate the line of code in fims_math that was throwing the error. I then used a Gemini query to come up with the solution that was implemented.

kellijohnson-NOAA:
@msupernaw when you are back from 🏖️ can you please review this PR so we can get it out has a hot fix? Thank you.

kellijohnson-NOAA:
@Andrea-Havron-NOAA I force pushed because I reset the base branch to main and cherry-picked your commit.

kellijohnson-NOAA (approved):

PR #958, opened 2025-08-15
Add data_small dataset and associated test and documentation

iantaylor-NOAA:
# What is the feature?
* Adds a second dataset `data_small` to the package, alongside the existing `data1`, as discussed in https://github.com/orgs/NOAA-FIMS/discussions/945

* Add R script `data-raw/data_small.R` which generates the dataset
* Add `data/data_small.rda` created by the script above, which runs the `usethis::use_data()` command at the end.
* Add test `tests/testthat/test-data_small.R` which confirms that the dataset can be converted to a FIMSFrame object and used in a model.
* Rename `R/data1.R` to `R/data.R` and adds documentation of the second data set in the same file, following the {tidyr} example described in https://r-pkgs.org/data.html#sec-documenting-data (they could be documented in separate files, but that might make keeping them aligned harder)
* Add `man/data_small.Rd` created by `devtools::document()` running on the `R/data.R` file noted above.

* No, it should have zero impact on all other areas. In the future we could include `data_small` in testing if it proves helpful.


kellijohnson-NOAA:
Thanks @iantaylor-NOAA for this PR. I plan on reviewing it today.

kellijohnson-NOAA:
@iantaylor-NOAA thank you for this PR and sorry that I have been so slow to review it. I am going to mark it as draft and I will come back to it after the CIE Review.

iantaylor-NOAA:
@kellijohnson-NOAA, the notifications about this PR reminded me that it's out of date relative to the current FIMS. I can commit further changes to update it. But it also looks like I should perhaps cherry pick the original commits related to data_small to separate them from other unrelated commits in August 2025.

kellijohnson-NOAA:
Sorry about those emails, they came from me deleting the dev branch before changing the base branch of this PR. I think we can just leave things as they are right now and maybe we can decide on a way forward after we get the comments back from the CIE Review?

kellijohnson-NOAA:
@iantaylor-NOAA what do you think about closing this PR and if you need a small data set, using code in @awilnoaa's growth vignette to generate a small data set with known growth parameters? She has a function in the vignette that will do it. That way we don't have to store another data set. Or, we can store the dataset she created and move its creation out of the growth vignette?

iantaylor-NOAA:
@kellijohnson-NOAA, thanks for circling back to this PR.

I agree with closing this vignette because data_small is far out of date.
I also see a lot of merit to the functional approach to creating example datasets, as applied in the @awilnoaa vignette.

However, it's probably worth stepping back to think about the overall goals of the example data to figure out how to move forward.

I first created data_small because data_big has 2808 rows and it was hard to debug whatever it was that I was debugging with that much data. The data_small had 60 rows and it looks like the demo data in the new vignette has 179, which is similarly small. Unfortunately I don't remember what I was working on at the time, but if you have an `nan` likelihood for some data type, it's definitely easier to figure out the cause if you have a handful of data of that type rather than hundreds of observations.

A few goals of data sets that I can think of include
- automated tests
- vignettes
- sandbox work for developers while expanding FIMS
- sandbox work for users to explore different options for data configurations

All of these options will likely benefit from more than just data_big, especially as the feature set of FIMS expands. Having the functions required to create a demo dataset within a vignette is great as a first step, but not a convenient approach to make it available for other uses. Moving the creation out of the vignette, and possibly also storing it seem like good ideas to me.

PR #959, opened 2025-08-18
Adds information from collaborative workflow to README and CONTRIBUTING

kellijohnson-NOAA:
# What is the feature?
* Brought in a lot of contributing information from the collaborative workflow to CONTRIBUTING.md, which is great because this file is visible to users from this repository rather than having to go to another repository to check stuff out.

* check_fims has been deprecated and the list of items was moved to CONTRIBUTING.md, though the helper files still remain in R/check_fims.R and I am not sure if they should be moved to R/utils.R
* README contains better information on how to install FIMS rather than just R-universe
* Potential settings for the .json file to set up VS Code were actually brought over to the file from the collaborative workflow but they are commented out, largely. We could uncomment them but I am not sure if they should be imposed on all users of VS Code. @MOshima-PIFSC I am not not certain if this file works for you because I think you are using something similar to but not exactly VS Code or maybe that was an extension in VS Code?

* Part of NOAA-FIMS/collaborative_workflow#170


CONTRIBUTING.md
```diff
@@ -1,21 +1,48 @@
...
+
+Along the development process it is important to add tests and ensure that the code is well documented. Below are the steps to follow to ensure that the code is ready for a pull request:
+* Build the doxygen-generated C++ documentation using `cmake --build build`, where the resulting html files will be in build/html. This can also be done using `setup_and_run_gtest()` in R.
+* Implement the suite of Google tests using `cmake --build build` and `ctest --test-dir build`. This can also be done using `setup_and_run_gtest()` in R. If there are failing tests, run `ctest --test-dir --rerun-failed --output-on-failure` to re-run the failed tests verbosely.
+* Run clang format to format the C++ code using TODO: document how to run clang-format.
```
MOshima-PIFSC, line 22: are the TODOs meant to be left in?

MOshima-PIFSC (approved):
I just had one small comment but other than that it looks good to me!

CONTRIBUTING.md
kellijohnson-NOAA: Thanks @MOshima-PIFSC for reading through the text so carefully. Yes, the TODO statement is meant to be in there because I did not know how to run clang-format but I did not want it to be a blocker of moving this PR forward.

PR #975, opened 2025-09-02
Refactor(input): convert nested input lists to tibbles

Bai-Li-NOAA:
# What is the feature?
* Refactor nested lists of input parameters to tibbles. Addresses discussion #957 

* add create_default_configurations() to set up configurations based on data
* update create_default_*() to use output from create() for parameter specification and it returns a tibble
* update initialize_*() to accept a tibble of parameters as input
* update vignettes and tests
* change "age" and "length" types to "age_comp" and "length_comp" in the data

*

### DIFF
--- a/tests/testthat/helper-integration-tests-setup-function.R
+++ b/tests/testthat/helper-integration-tests-setup-function.R
@@ -547,7 +547,7 @@ setup_and_run_FIMS_with_wrappers <- function(iter_id,
   # The model will not always run when log_q is very small.
   # We will need to make sure log_q is the true value for deterministic runs but
   # then reset to log(1.0) for estimation runs.
-  if (estimation_mode == TRUE){
+  if (estimation_mode == TRUE) {
     parameters <- parameters |>
       dplyr::mutate(
         value = dplyr::if_else(
--- a/R/create_default_configurations.R
+++ b/R/create_default_configurations.R
@@ -19,7 +19,7 @@
 #'   \item **Growth:** An empirical weight-at-age (EWAA) growth module.
 #'   \item **Maturity:** A logistic maturity module.
 #' }
-#' The final output is a nested tibble, which serves as a starting point for 
+#' The final output is a nested tibble, which serves as a starting point for
 #' building a complete FIMS model configuration.
 #'
 #' @param data An S4 object of class `FIMSFrame`. FIMS input data.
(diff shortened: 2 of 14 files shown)


Bai-Li-NOAA:
- [ ] I just realized that I need to add R tests for `create_default_configurations()`.

inst/WORDLIST
kellijohnson-NOAA, line 1: @Bai-Li-NOAA I know that this is really nit-picky but can you please alphabetize these words in the list without respect to case?

R/distribution_formulas.R
```diff
@@ -264,7 +264,11 @@ get_expected_name <- function(family, data_type) {
 initialize_data_distribution <- function(
     module,
     family = NULL,
-    sd = list(value = 1, estimation_type = "constant"),
+    # Create a tibble with value and estiamtion_type column for sd
+    sd = dplyr::tibble(
```
kellijohnson-NOAA, line 268: We have {tibble} as an import if you want to use the direct function instead of the exported one.

tests/testthat/helper-integration-tests-setup-function.R
```diff
@@ -542,55 +542,18 @@ setup_and_run_FIMS_with_wrappers <- function(iter_id,
...
-    update_parameters(
-      modified_parameters = modified_parameters[[iter_id]]
-    )
+  parameters <- modified_parameters
 
   # The model will not always run when log_q is very small.
   # We will need to make sure log_q is the true value for deterministic runs but
   # then reset to log(1.0) for estimation runs.
-  if (estimation_mode == TRUE) {
+  if (estimation_mode == TRUE){
```
kellijohnson-NOAA, line 550: There should be a space here so the change is not needed, might run styler on the changes?

R/distribution_formulas.R
```diff
@@ -294,47 +298,32 @@ initialize_data_distribution <- function(
...
     new_module <- methods::new(DnormDistribution)
 
     # populate logged standard deviation parameter with log of input
-    new_module$log_sd$resize(length(sd[["value"]]))
-    for (i in seq_along(sd[["value"]])) {
-      new_module$log_sd[i]$value <- log(sd[["value"]][i])
-    }
+    purrr::walk(
+      seq_along(sd[["value"]]),
+      \(x) new_module[["log_sd"]][x][["value"]] <- log(sd[["value"]][x])
```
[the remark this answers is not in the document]
Bai-Li-NOAA: Looks like the `dev` version of FIMS has an internal setter for setting the position of a ParameterVector from R (see code [here](https://github.com/NOAA-FIMS/FIMS/blob/c4bd6da125faa0ac431c0a9d8653003da7954d92/inst/include/interface/rcpp/rcpp_objects/rcpp_interface_base.hpp#L286)). I can take this on in a separate PR? I can check what methods are available through the Rcpp interface and update the R tests and vignettes to match.

tests/testthat/helper-integration-tests-setup-function.R
Bai-Li-NOAA: Good catch. Done!

PR #976, opened 2025-09-05
Dev json caa

msupernaw:
# What is the feature?
* Refactored JSON output for The catch at age model

* Moved the output to be model-family specific. 

* Tests will be impacted.

### DIFF
--- a/tests/testthat/helper-integration-tests-validation.R
+++ b/tests/testthat/helper-integration-tests-validation.R
@@ -105,7 +105,7 @@ validate_fims <- function(
   naa1_id <- seq(1, (om_input[["nyr"]] * om_input[["nages"]]), by = om_input[["nages"]])
   if (use_fimsfit) {
     expect_equal(
-      report[["recruitment"]][[1]][1:om_input[["nyr"]]],
+      report[["expected_recruitment"]][[1]][1:om_input[["nyr"]]],
       estimates |>
         dplyr::filter(
           label == "NAA"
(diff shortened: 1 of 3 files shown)


inst/include/interface/rcpp/rcpp_interface.hpp
```diff
@@ -174,6 +174,22 @@ std::string finalize_fims(Rcpp::NumericVector par, Rcpp::Function fn,
   return ret;
 }
 
+/**
+ * @brief Sets the fixed parameters vector object.
+ *
+ * @param par A vector of parameter values.
+ */
```
kellijohnson-NOAA, line 181: Thank you for documenting this parameter and the next one. Unfortunately I still don't know what they do or when I would use it. Can you add some more information?

manual.R
kellijohnson-NOAA, line 1: This file needs removed from the PR.

src/FIMS-ab26917b.o.tmp
kellijohnson-NOAA, line 1: Should this file be committed?

caa_output.json_
kellijohnson-NOAA, line 1: This file needs removed from the PR.

src/FIMS-ab26917b.o.tmp
msupernaw: nope

kellijohnson-NOAA:
I am closing this PR because we have an updated branch dev-json-caa-rebased, where this one is too difficult to rebase to dev.

Issue #852, opened 2025-05-28
[feature:] Allow `update_parameters()` to change a parameter vector's estimation type through time


kellijohnson-NOAA:
### Describe the situation that led to the request and a solution

I would like to be able to change a parameter vector that is by default not time-varying, i.e., a fixed effect parameter vector like fishing mortality, into a vector of both constant values and fixed effect values. But, right now the wrapper functions prohibit this because the default is to set vectors with a single estimation type, e.g., "fixed_effect", and when you try to pass a vector of strings as the new vector to `update_parameters()` you get an error saying that the vectors are of different length.

### Alternative solutions

We can leave it as is and just have users change things manually, which is what I am doing in the test for right now, e.g.,
```
  parameters[[1]][["fleet1"]][["Fleet.log_Fmort.estimation_type"]] <- c(
    rep("constant", 3),
    rep("fixed_effect", 27)
  )
```

@Bai-Li-NOAA can you comment on if you think that we need this feature or not or if we should just provide examples of how users can change it manually like how I show above.

Andrea-Havron-NOAA:
> [Edited]: [@Andrea-Havron-NOAA](https://github.com/Andrea-Havron-NOAA), are you working on `update_process()` to add a random effect after the parameters are set up. Just wondering if this is something I should look into before refactoring `update_parameters()`.

@Bai, I put development of update_process on hold until you've made your changes. You can look into the structure of update_process for ideas, but I am happy to refactor this function as needed.

kellijohnson-NOAA:
@Bai-Li-NOAA I think we can close this issue right because `update_parameters()` is removed in your PR #975

kellijohnson-NOAA:
Brilliant, thanks Bai.

kellijohnson-NOAA:
@Bai-Li-NOAA I know that you have completed the portion in the wrapper functions that would allow a user to change a parameter's estimation_type over time but do we have a test that it actually works with the C++ code and produces output consistent with changing from fixed_effect to constant?

kellijohnson-NOAA:
@Bai-Li-NOAA can you comment on this issue's status? It is the last issue related to [Milestone 13](https://github.com/NOAA-FIMS/FIMS/milestone/13) and it would would be great to be able to close out that milestone.

kellijohnson-NOAA:
@Bai-Li-NOAA sorry to bug you again but what is the status of this issue?

Bai-Li-NOAA:
[...]

PR #990, opened 2025-09-22
refactor(tests): update integration test setup

### LINKED ISSUE #940: [feature:] let nan values play through the JSON output without crashing to help with debugging
### Describe the situation that led to the request and a solution

As described by @MOshima-PIFSC in #933, if there are `nan` values in the model results, the functions crash at the stage of calling `reshape_json_estimates()`. This seems to be because `nan` isn't a valid JSON value and `jsonlite::fromJSON()` can't handle them. I think that if we converted any `nan` or `NA` values to `null` in the JSON output at the step of `finalized_fims <- finalize()`, then they could be converted back to NA via arguments in `jsonlite::fromJSON()` and then debugging would be easier.

I can expand #933 if that would be easier than separate issues, but initially it seems like that one should focus on what's causing the `nan` values vs this one focused on helping find the cause of the nan values the next time it occurs for some different reason.

### Alternative solutions

Add error messages when `nan` values are generated in the dynamics before creating the JSON files.

### Statistical validity

_No response_

### Additional context

_No response_

### LINKED ISSUE #983: [Developer Issue]: `prepare_test_data()` is inefficient for running individual tests
### Description

Running any test that calls `prepare_test_data()` in `tests/testthat/helper-integration-tests-setup-run.R` triggers the creation of all test datasets and model fits, including those unrelated to the specific test. This slows down the development cycle. We need a solution that allows moving the code from `prepare_test_data()` into individual tests without causing significant code duplication.

Bai-Li-NOAA:
# What is the feature?
* This PR addresses issues #983 and #940

- Create helper functions (`load_*()`) to load or create data for a FIMS model run, or run a FIMS model with wrappers to save the fit output. These functions check if the corresponding `.RDS` file exists; if not, they generate the data or run the model and save the output to an `.RDS` file. If the file exists, they read and return the object from the `.RDS` file. This ensures that test data is generated only once and is available for subsequent test runs, speeding up the process.
- Add error handling (e.g., `tryCatch()`) in data loading functions to provide clearer messages when file operations fail.
- Update README to document new helper functions for loading data and running FIMS with wrappers.
- Rename `test-integration-caa-mle.R` to `test-integration-caa-mle-without-wrappers.R` so it can be filtered when running tests (`devtools::test(filter = "xxx")`), speeding up single-test execution.
- Update snapshot tests outputs from `.md` files to `.csv` files.
- Replace nan with -999 in JSON files. Thanks to @msupernaw for the fix to issue #940.

* 

Task:
[ ] add an integration test where we manually set the fixed-effect _F_s to constant (address issue #852)


kellijohnson-NOAA:
@Bai-Li-NOAA should we continue to leave this PR as a draft or close it?

Bai-Li-NOAA:
@kellijohnson-NOAA I will close it for now. I still need to rebase the branch and then refactor the integration tests.

PR #1010, opened 2025-10-02
Update data1 to use timing

Bai-Li-NOAA:
# What is the feature?
* using timing from now on instead of date where timing is a
double which allows for the year to be incremented using whatever
resolution you input. Right now, the only resolution that is available
in FIMS is year so timing must be an integer value.

* 

*


Bai-Li-NOAA:
Cherry-picked the actual commit from `dev-json-caa-rebased-data1` to `dev-data1` and opened a different issue.

Issue #973, opened 2025-08-26
[Developer Issue]: error on first instance in doxygen


kellijohnson-NOAA:
### Description

In the yml file for the doxygen rendering, it specifies
https://github.com/NOAA-FIMS/FIMS/blob/2c7101a0cca7117fe7f3591c3e2f5245eaa8b193/.github/workflows/build-doxygen.yml#L35
which means that on the first instance of the doxygen not being correct it stops and issues an error. It would be really nice if we were able to see all of the errors that way you would know everything you need to fix rather than it just erroring each time until you get them fixed. I am uncertain though if we change it to
```
         cmake -S . -B build -G Ninja -DDOXYGEN_WARN_AS_ERROR=NO
```
if the action will fail if there are warnings. @k-doering-NOAA or @Bai-Li-NOAA do either of you know what the pass/fail behavior will be if we change this to NO? I would rather not have the test pass if there are problems with the doxygen.

kellijohnson-NOAA:
Brilliant 🥇 thank you for investigating this! I will mark it as a good first contributor issue in hopes that someone signs up to make the change.

PR #1016, opened 2025-10-07
fix(gha) fix doxygen error warning

awilnoaa:
# What is the feature?
* This updates the Doxygen build workflow to use FAIL_ON_WARNINGS.
Doxygen will now run to completion and report all warnings,
then return a non-zero exit code so the workflow still fails if warnings exist.

- addresses issue #973 

* Used: cmake -S. -B build -G Ninja -DDOXYGEN_WARN_AS_ERROR=FAIL_ON_WARNINGS

* N/A

### DIFF
--- a/.github/workflows/build-doxygen.yml
+++ b/.github/workflows/build-doxygen.yml
@@ -32,7 +32,7 @@ jobs:
     # Do not run -DDOXYGEN_WARN_AS_ERROR=YES locally
     # FIX? run cmake --build build --parallel 16
       run: |
-        cmake -S. -B build -G Ninja -DDOXYGEN_WARN_AS_ERROR=FAIL_ON_WARNINGS
+        cmake -S . -B build -G Ninja -DDOXYGEN_WARN_AS_ERROR=FAIL_ON_WARNINGS
         cmake --build build
     
     - name: Deploy html documentation with Doxygen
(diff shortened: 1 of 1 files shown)


.github/workflows/build-doxygen.yml
```diff
@@ -32,7 +32,7 @@ jobs:
     # Do not run -DDOXYGEN_WARN_AS_ERROR=YES locally
     # FIX? run cmake --build build --parallel 16
       run: |
-        cmake -S . -B build -G Ninja -DDOXYGEN_WARN_AS_ERROR=YES
+        cmake -S. -B build -G Ninja -DDOXYGEN_WARN_AS_ERROR=FAIL_ON_WARNINGS
```
Bai-Li-NOAA, line 35: @awilnoaa, could you add a space here: `-S .`? That was my bad, when I posted the solution on the issue page, I forgot to include the space between `-S` and `.`. It technically works without it, but adding the space makes it clearer and more consistent. After that, could you also squash the commits into one? Happy to hop on a quick chat if you’d like to go over the squashing together!

Bai-Li-NOAA (changes requested):
Everything looks great! I just have a small request to adjust the command format. Thanks so much for working on this issue!

.github/workflows/build-doxygen.yml
awilnoaa: @Bai-Li-NOAA I added the space and I believe I squashed the commits. Let me know if there were any issues with the process. Thank you for all your help!

Issue #998, opened 2025-09-28
[Bug:] No validity checks from "type" in FIMSFrame


kellijohnson-NOAA:
### Describe the bug

Recently, I was updating the case studies and noticed that no age-composition data were getting passed through. The age and length "type" have been changed in dev to "age_comp" and "length_comp" but there were no warnings or messages when running FIMSFrame.

### Describe how to reproduce the bug

```
data("data1")

test <- data1 |>
  dplyr::mutate(type = ifelse(type == "age_comp", "age", type))
# Prepare data for FIMS model
data_4_model <- FIMSFrame(data1)
parameters <- data_4_model |>
  create_default_configurations() |>
  create_default_parameters(data = data_4_model)
initialize <- parameters |>
  initialize_fims(data = data_4_model)
```

### Which version of FIMS are you seeing the problem on?

dev 2025-09-28

PR #1019, opened 2025-10-16
feat(issue-998): add validation for supported and missing data types in FIMSFrame

awilnoaa:
# What is the feature?

* Stops the model if any unsupported data types are present (e.g., “indexes” instead of “index”)
* Warns the user (but continues) if expected data types are missing (e.g., length_comp)

addresses issue #998 

* Defined which data types are allowed and/or expected
* Added a check for unsupported tops and stops execution if any are found
* Added a check for missing expected types and issues a warning instead of stopping execution
* Added three test_that() blocks to verify each scenario (error, warning, valid inputs)

* No, changes should be contained within FIMSFrame

### DIFF
--- a/R/fimsframe.R
+++ b/R/fimsframe.R
@@ -569,7 +569,24 @@ methods::setValidity(
     }
 
     # TODO: Add checks for other slots
+    # Add validity check for types
+    allowed_types <- c(
+  "landings", "index", "age_comp", "length_comp",
+  "weight-at-age", "age-to-length-conversion"
+)
+present_types <- unique(object@data[["type"]])
 
+    # Issues warning if there are any unrecognized types
+    unknown_types <- setdiff(present_types, allowed_types)
+    if (length(unknown_types) > 0) {
+      cli::cli_warn(c(
+        "!" = "Data contains unexpected type(s): {paste(sort(unknown_types), collapse = ', ')}",
+        "i" = "Allowed types are: {paste(allowed_types, collapse = ', ')}",
+        "i" = paste("Model will continue to run," ,
+                    "but check that data types are correct."
+                    )
+  ))
+}
     # Return
     if (length(errors) == 0) {
       return(TRUE)
--- a/tests/testthat/test-fimsframe.R
+++ b/tests/testthat/test-fimsframe.R
@@ -70,7 +70,9 @@ test_that("fims_frame() works with the correct inputs", {
   #' @description Test that the `show()` method works as expected on a FIMSFrame
   #' object.
   expect_output(suppressMessages(show(fims_frame)))
-})
+  #' @description Test that 'FIMSFrame()' succeeds cleanly with valid inputs.
+  expect_no_error(FIMS::FIMSFrame(data1))
+})  
 
 ## Edge handling ----
 # No edge cases to test.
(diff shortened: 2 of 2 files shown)


R/fimsframe.R
```diff
@@ -569,7 +569,35 @@ methods::setValidity(
     }
 
     # TODO: Add checks for other slots
+    # Add validity check for types
+    allowed_types <- c(
+  "landings", "index", "age_comp", "length_comp",
+  "weight-at-age", "age-to-length-conversion"
+)
+present_types <- unique(object@data$type)
```
Bai-Li-NOAA: Could you change `object@data$type` to `object@data[["type"]]`? In FIMS, we use double brackets with quoted names instead of the dollar sign to avoid partial matching.

tests/testthat/test-fimsframe.R
```diff
@@ -208,4 +208,33 @@ test_that("get_n_fleets() works with correct inputs", {
 # No edge cases to test.
 
 ## Error handling ----
-# No built-in errors to test.
+
+## Check that user does not include unsupported data types
+test_that("FIMSFrame() aborts on unsupported data types", {
+  data("data1", package = "FIMS")
```
Bai-Li-NOAA: We can remove this line and use `data1` directly.

tests/testthat/test-fimsframe.R
```diff
@@ -208,4 +208,33 @@ test_that("get_n_fleets() works with correct inputs", {
...
+  bad <- dplyr::mutate(
+    data1,
+    type = ifelse(type == "index", "indexes", type)  # Introduce an unsupported type
+  )
+  expect_error(FIMS::FIMSFrame(bad), regexp = "unsupported type\\(s\\)")
```
Bai-Li-NOAA: Could you add a `@description` tag with a **one-line** summary of the test above `expect_error()` so it can be rendered in future {testdown} reports? Please see the example at line 191.

tests/testthat/test-fimsframe.R
```diff
@@ -208,4 +208,33 @@ test_that("get_n_fleets() works with correct inputs", {
...
+
+  expect_warning(
+    ff <- FIMS::FIMSFrame(no_len),
+    regexp = "expected data type\\(s\\) are not present"
+  )
+  expect_s4_class(ff, "FIMSFrame")
+})
+
+##Confirm that valid inputs work without error
+test_that("FIMSFrame() succeeds cleanly with valid inputs", {
```
Bai-Li-NOAA: Thanks for writing this test! Do you think we could move it to the IO correctness section, maybe around line 25?

Bai-Li-NOAA (changes requested):
Thanks for tackling this issue! It looks really good. I have a few small notes on documentation and style, and one comment on the warning message. Let me know if anything’s unclear. Feel free to push more commits to the feature branch and squash them into one at the end. Happy to demo squashing using the VS Code terminal if needed.

tests/testthat/test-fimsframe.R
awilnoaa: Ok, added.

tests/testthat/test-fimsframe.R
```diff
@@ -208,4 +208,33 @@ test_that("get_n_fleets() works with correct inputs", {
...
+  data("data1", package = "FIMS")
```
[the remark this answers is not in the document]
awilnoaa: Test was moved to IO correctness section.

PR #1026, opened 2025-11-11
Issue 1020-Remove nseasons from FIMS

awilnoaa:
# What is the feature?
* Refactors the codebase to remove the unused `nseasons` parameter and related logic.

* Removed all instances of  `nseasons` and its related functions from the codebase. 

* Minimal. This cleanup only removed seasonality placeholders and should have no effect on existing models or tests.


awilnoaa:
Thank you @Bai-Li-NOAA , will do!

PR #1034, opened 2025-11-13
fix(devcontainer): use `github::username/repository` to install non-CRAN packages

Bai-Li-NOAA:
# What is the feature?
* My bad! I introduced a bug in the `devcontainer.json` that caused a Codespace on dev to fail with `pak` installation errors. I’ve updated the file to resolve the issue.

* Used `github::username/repository` to install non-CRAN packages

* 

To test it, open a new Codespace from this page: https://github.com/NOAA-FIMS/FIMS/tree/dev-fix-devcontainer. The Codespace should start up without any errors.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,68 +1,66 @@
 // For format details, see https://aka.ms/devcontainer.json. For config options, see the README at:
 // https://github.com/microsoft/vscode-dev-containers/tree/v0.234.0/containers/r
 {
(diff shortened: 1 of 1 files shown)


kellijohnson-NOAA:
@Bai-Li-NOAA it is hard to see the changes that were made because of all of the white-space changes that are also included. Which formatting style is more standard? I don't mind including the white-space changes if that is seen as best practice here but I don't really want to accept them to have an automated formatter come back and delete them later.

PR #1035, opened 2025-11-17
Dev projections caa

nathanvaughan-NOAA:
# What is the feature?
* Add projection capabilities to FIMS

* Allow FIMS to estimate annual fishing mortality without catch data, set priors on derived quantities like relative spawning biomass, and estimate a single mean parameter across multiple years to fix projection F as constant.  

* probably yes, can check for other conflicts once we figure out the snapshot changes

### DIFF
--- a/inst/include/distributions/functors/normal_lpdf.hpp
+++ b/inst/include/distributions/functors/normal_lpdf.hpp
@@ -40,6 +40,8 @@ struct NormalLPDF : public DensityComponentBase<Type> {
   virtual const Type evaluate() {
     // set vector size based on input type (prior, process, or data)
     size_t n_x = this->get_n_x();
+    // get expected value vector size
+    size_t n_expected = this->get_n_expected();
     // setup vector for recording the log probability density function values
     this->lpdf_vec.resize(n_x);
     this->report_lpdf_vec.resize(n_x);
@@ -50,14 +52,14 @@ struct NormalLPDF : public DensityComponentBase<Type> {
     lpdf = static_cast<Type>(0);
 
     // Dimension checks
-    if (n_x != this->expected_values.size()) {
-      throw std::invalid_argument(
-          "NormalLPDF::Vector index out of bounds. The size of observed data "
-          "does not equal the size of expected values. The observed data "
-          "vector is of size " +
-          fims::to_string(n_x) + " and the expected vector is of size " +
-          fims::to_string(this->expected_values.size()));
+    if (n_x != n_expected) {
+      if (n_expected == 1) {
+        n_expected = n_x;
+      } else if (n_x > n_expected) {
+        n_x = n_expected;
+      }
     }
+
     if (this->log_sd.size() > 1 && n_x != this->log_sd.size()) {
       throw std::invalid_argument(
           "NormalLPDF::Vector index out of bounds. The size of observed data "
--- a/inst/include/interface/rcpp/rcpp_objects/rcpp_distribution.hpp
+++ b/inst/include/interface/rcpp/rcpp_objects/rcpp_distribution.hpp
@@ -36,6 +36,30 @@ class DistributionsInterfaceBase : public FIMSRcppInterfaceBase {
    */
   SharedString input_type_m;
   /**
+   * @brief Control flag indicating whether to use the expected mean in the
+   * distribution calculations.
+   *
+   * This shared string member serves as a boolean flag (i.e., "yes" or "no")
+   * that determines whether the distribution should use the `expected_mean`
+   * vector or other expected values (e.g., from data or random effects) when
+   * computing the expected value in the likelihood calculations.
+   *
+   * When set to "no" (default), the distribution uses expected values based on
+   * the `input_type` setting (data expected values for "data", random effects
+   * expected values for "random_effects", or standard expected values
+   * otherwise).
+   *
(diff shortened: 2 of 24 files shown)


kellijohnson-NOAA:
@nathanvaughan-NOAA I created another branch dev-projections-caa-kfj where I am working on rebasing this branch to dev. One of the major changes was the removal of seasons and changes of nyears, nfleets, nages, and nlengths to have underscores after the n_*. Another was the removal of a bunch of detritus code from changing over to model families. I clearly have not been 100% successful with the rebase because some of the google tests are failing but at least it compiles now. I am still waiting to see which tests fail in testthat for R. But, I need to take a break. If you have some time today and can push any changes that you see are necessary to the dev-projections-caa-kfj branch please feel free.

nathanvaughan-NOAA:
Sure thing, taking a look now :)

nathanvaughan-NOAA:
I fixed a few simple test issues. It currently just crashes in Github actions from a fims vector out of bounds error. I ran tests locally and for me it's crashing R in "test-integration-caa-mle.R at the end when it trys to run setup_and_run_FIMS_without_wrappers after changing the age comp sample size to 1. When I ran that locally it gets all the way through MakeADFun and then crashes during optimization.

nathanvaughan-NOAA:
@kellijohnson-NOAA  I pushed up another rebase to a new branch dev-projections-caa-again and have it paired down to just the setting comp sample size to 1 test crashing R (I have commented it out of the test at the moment) and a failing no estimation test because spawning_biomass_ratio is recalculated inside fims even when it is set as a constant so it's value changes and the test fails. This just needs a test modification that ignores these values. I'm not sure what is causing the crash though maybe @Andrea-Havron-NOAA or @msupernaw have an idea, I'm happy to walk through it with folks if that would help. Or maybe we should focus on fixing the R crash issue then we can just debug whatever the error is directly.

kellijohnson-NOAA:
gdb will not work on windows with the combination of R and compiled C++ code. You can use it for debugging the google tests though, just not R integration tests. You will need @msupernaw to do it on his mac or fire up a codespace and do it on a machine with 16 cores.

nathanvaughan-NOAA:
Awesome! Thanks @msupernaw

inst/include/distributions/functors/normal_lpdf.hpp
```diff
@@ -50,14 +52,25 @@ struct NormalLPDF : public DensityComponentBase<Type> {
...
+    //       "LognormalLPDF::Vector index out of bounds. The size of observed "
+    //       "data does not equal the size of expected values. The observed data
+    //       " "vector is of size " +
+    //         fims::to_string(n_x) + " and the expected vector is of size " +
+    //         fims::to_string(n_expected) + ". with input type " +
+    //         fims::to_string(this->input_type) + ". Final data value = " +
+    //         fims::to_string(this->get_observed(n_x - 1)) +
+    //         " and the final expected value = " +
+    //         fims::to_string(this->get_expected(n_expected - 1)));
+    // }
```
kellijohnson-NOAA: Remove code that is commented out.

inst/include/interface/rcpp/rcpp_objects/rcpp_distribution.hpp
```diff
@@ -96,6 +101,13 @@ class DistributionsInterfaceBase : public FIMSRcppInterfaceBase {
     return false;
   }
 
+  /**
+   * @brief Set mean expected value for distribution.
+   *
+   * @param input_value Value to use for the distribution mean.
+   */
```
kellijohnson-NOAA: @copilot can you suggest more verbose documentation for this virtual function please.

inst/include/interface/rcpp/rcpp_objects/rcpp_models.hpp
kellijohnson-NOAA: I am not sure why this was removed?

nathanvaughan-NOAA: No worries it's ParameterVector log_r on line 143(old file) 147(modified file) that could be removed, it only exists there I think and never gets used. I was just working on the logic that if we intend to move everything to catch at age we should figure out the bare minimum that has to stay in population so copy cats like me don't get confused in the future and add things in the wrong place.

kellijohnson-NOAA: No worries. I have removed log_r from rcpp_population.hpp now. You were right, it was only in that file one time.

inst/include/interface/rcpp/rcpp_objects/rcpp_population.hpp
```diff
@@ -337,6 +330,34 @@ class PopulationInterface : public PopulationInterfaceBase {
     population->recruitment_id = this->recruitment_id.get();
     population->maturity_id = this->maturity_id.get();
     population->log_M.resize(this->log_M.size());
+
+    if (this->log_f_multiplier.size() == (this->n_years.get())) {
+      population->log_f_multiplier.resize(this->log_f_multiplier.size());
+    } else {
+      FIMS_WARNING_LOG(
+          "The log_f_multiplier vector is not of size nyears. Filling with "
```
kellijohnson-NOAA:
```suggestion
          "The log_f_multiplier vector is not of size n_years. Filling with "
```

PR #1062, opened 2025-12-09
fix(doxygen): adds the correct header to doxygen website

kellijohnson-NOAA:
# What is the feature?
* Fix the header in the doxygen to match the website.

* The doxygen header was not working so this changes the reference in the GitHub workflow

* No


e-perl-NOAA (approved):
praying that this finally works!

PR #1081, opened 2025-12-16
chore(deps): Bot targets dev instead of main

kellijohnson-NOAA:
# What is the feature?
* Opens PRs from dependabot to dev instead of main
* Assigns FIMS Project Lead to review them

* Added target-branch and reviewers to the dependabot.yml file
* Added a Team on NOAA-FIMS called "Project Lead" that can be referenced using @noaa-fims/project-lead to assign all team members.

* No

### DIFF
--- a/.github/workflows/pr-checklist.yml
+++ b/.github/workflows/pr-checklist.yml
@@ -7,7 +7,7 @@ on:
     types: [opened]
 
 permissions:
-  contents: write
+  pull-requests: write
 
 jobs:
   pr-checklist:


k-doering-NOAA:
@kellijohnson-NOAA I see this! I can take a look by the end of the week.

.github/dependabot.yml
```diff
@@ -7,3 +7,6 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    target-branch: "dev"
+    reviewers:
```
k-doering-NOAA: Is reviewers an option for the dependabot file? I couldn't find it in the documentation, but perhaps I missed it I did find an option for [`assignees`](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#assignees--) (For assignees, it looks like it may need to be an individual rather than a team)

It looks like for reviewers, they are added through the CODEOWNERS file: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/customizing-dependabot-prs#automatically-adding-reviewers

R/create_default_parameters.R
```diff
@@ -251,6 +251,10 @@ create_default_Population <- function(
   n_years <- get_n_years(data)
   n_ages <- get_n_ages(data)
 
+  # Define age and year vectors needed for the rep() calls ===
+  ages <- get_ages(data)
+  years <- get_start_year(data):get_end_year(data)
```
k-doering-NOAA: Is this other work that snuck in? It looks like there are a few commits that are unrelated to dependabot and updating the github action version in files. I didn't review these, but can if it is intended to be in this PR.

PR #1104, opened 2025-12-31
Dev parameter to from vector

### LINKED ISSUE #766: Use fromRVector() and toRVector() in ParameterVector and RealVector
The interface refactor [PR](https://github.com/NOAA-FIMS/FIMS/pull/765) introduces fromRVector() and toRVector() methods for RealVector. These methods simplify usage by being more concise than the previous approach, which required calling resize(), purrr::walk(), and set() to assign values. It would be beneficial to implement these methods consistently in ParameterVector as well. Once that's done, we can update the tests to use fromRVector().

_Originally posted by @Bai-Li-NOAA in https://github.com/NOAA-FIMS/FIMS/issues/765#issuecomment-2733867261_

### LINKED ISSUE #893: [Developer Issue]: `purrr::walk()` is taking things from the global environment
### Description

The use of `purrr::walk()` takes one argument, a vector of integers, to do the sequencing but it is taking other things from the environment, the module and the values being passed to the module. I am not sure if this should be changed? Some of it is easily changed using `purrr::walk2()` but we cannot really fix all of it because we cannot return the changed module. I think we need some testing here.
For more context, here is an example
https://github.com/NOAA-FIMS/FIMS/blob/44548591b4eeaa3b66123dd622b0a3d0a6d5e6ce/R/distribution_formulas.R#L296-L299

At a bare minimum, I think that we should be using `purrr::walk2()` which allows .x and .y to be passed, which would allow us to pass the integer vector and the vector of values. Then maybe we can do some tests about passing of the module?

msupernaw:
# What is the feature?
* 
Updates toRVector and fromRVector in RealVector class to use snake case.
Adds  the following methods to the ParameterVector class:
values_to_R_vector
min_values_to_R_vector
max_values_to_R_vector
estimation_types_to_R_vector
uncertainty_values_to_R_vector
final_values_to_R_vector
values_from_R_vector
final_values_from_R_vector
min_values_from_R_vector
max_values_from_R_vector
estimation_types_from_R_vector
uncertainty_values_from_R_vector

* 
Added new methods and refactored names for existing ones. 

* Yes, I had to update two lines in the testthat tests (test-initialize_modules.R).


msupernaw:
> @msupernaw did you mean to open this PR to main? I thought we talked about branching off of dev and doing the PR to dev? You can change the branch by going to the top and selecting "Edit" and switching main to dev. Also, there are no tests for any of these functions. Can we write tests using google tests or will they have to be in R because they are meant to be used by R users?

@kellijohnson-NOAA Nope, meant to be a PR to dev. I'll work on some R tests today.

msupernaw:
@kellijohnson-NOAA The tests for the new member functions are implemented in test-rcpp-parameter-vector.R and they are all passing. You should be good to go

kellijohnson-NOAA:
@msupernaw can you please ensure that all tests are passing so I can review this PR. thanks

PR #1125, opened 2026-01-09
add tests to check dimensions of log_sd and log_Fmort

Andrea-Havron-NOAA:
# What is the feature?
* Adds tests that check dimensions of log_sd and log_Fmort

* Created a new I/O test for log_sd that checks model runs correctly when log_sd is a scalar, both when the parameter's `estimation_type` is `fixed_effects` or `constant` (Correct dimensions when log_sd and log_Fmort are length n_years are checked in `test-integration-caa-mle`)
* Created new error tests for log_sd and log_Fmort when parameter vector lengths are either too short or too long. 

To implement checks on tests, the following changes were made in the codebase:
  - log_sd was added to JSON and FIMS output. 
  - increased precision of JSON output to R default precision
  - added log_Fmort dimension check to rcpp_fleet
  - added log_sd dimension check to check_distribution_validity()
  - added try/catch statement to model->Evaluate in FIMS.cpp to handle std::exceptions
  - updated ParameterVector dimension check


kellijohnson-NOAA:
Does this PR need to be closed?

tests/testthat/test-integration-fleet-log-obs-error-input.R
```diff
@@ -0,0 +1,186 @@
...
+# work when using the reporting tool {testdown} because it cannot
+# find the path back to the original test file in the GitHub Actions runner. This
+# leads to errors like: In normalizePath(attr(result$srcref, "srcfile")$filename) :
+# path[1]="test-use-test-template.R": No such file or directory. To fix the issue,
+# we need to skip the test file when generating {testdown} reports.
+if (Sys.getenv("INPUT_TESTDOWN") == "true") {
+  #' @description Skip the test in {testdown} reports generation.
+  skip("Skipping test in {testdown} reports generation.")
+}
+
```
kellijohnson-NOAA: Did you mean to include this?

tests/testthat/test-integration-fleet-log-obs-error-input.R
```diff
@@ -0,0 +1,186 @@
...
+    dplyr::filter(
+      !(fleet_name == "fleet1" & label == "log_sd" & time > 1) |
+        is.na(fleet_name == "fleet1" & label == "log_sd" & time > 1)
+    )
+
+  test_fit <- parameters_4_model |>
+    initialize_fims(data = data_4_model) |>
+    fit_fims(optimize = FALSE)
+
+  json_estimates <- reshape_json_estimates(test_fit@model_output)
```
kellijohnson-NOAA: We should not use @ to get things out of a class. This should be `get_estimates(test_fit)`. This comment should be carried forward to other places where the @ symbol is used as well.

tests/testthat/test-integration-fleet-log-obs-error-input.R
```diff
@@ -0,0 +1,186 @@
...
+  for (i in seq_along(log_sd_output$log_sd_values)) {
+    expect_equal(log_sd_input$value, log_sd_output$log_sd_values[i])
+  }
+})
+
+
+## Error handling ----
+test_that("`log_sd` returns correct error messages when wrong dimensions", {
+  #' @description Test that returns correct error message when log_sd is too short.
+  #' The error is caught in initialize_modules, not in C++.
```
kellijohnson-NOAA: #' @description lines should be a single line that is not wrapped. We have to do this because the testdown package will not work with multiple lines even though our instructions say we can use multiple lines. The instructions need updating.

R/reshape_output.R
```diff
@@ -401,8 +401,19 @@ dimensions_to_tibble <- function(data) {
...
+  }
+  df <- tidyr::unnest(df, c(value, expected_values, observed_values))
+
+  # Only unnest log_sd_values if it is not all NULL
+  if (!all(sapply(df$log_sd_values, is.null))) {
+    df <- tidyr::unnest(df, log_sd_values, keep_empty = TRUE)
+  }
+
+  df <- dplyr::rename(df, likelihood = value)
+  df
```
kellijohnson-NOAA: Should be the following I think
```
dplyr::rename(df, likelihood = value)
```

inst/include/interface/rcpp/rcpp_objects/rcpp_distribution.hpp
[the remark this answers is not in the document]
nathanvaughan-NOAA: @Andrea-Havron-NOAA  and @kellijohnson-NOAA  That all seems to look ok to me if it is still passing the projections tests. I agree that the resize could use force scalar. My only thought is could the 

 FIMS_WARNING_LOG(
              "log_sd size does not match number of observations and is not "
              "scalar."); 

code result in a segfault crash later if it isn't stoping the code. Or do we need that because of my projections subsetting? I can't wait to fix this all properly in the likelihood calculations :)

inst/include/interface/rcpp/rcpp_objects/rcpp_distribution.hpp
```diff
@@ -258,24 +258,44 @@ class DnormDistributionsInterface : public DistributionsInterfaceBase {
...
+      size_t n_x = dnorm->get_n_x();
+
+      if (this->log_sd.size() != n_x) {
+        // If log_sd size == 1 (scalar), repeat the entry
+        if (this->log_sd.size() == 1) {
+          auto tmp = this->log_sd[0];  // copy the one log_sd param
+          this->log_sd.resize(n_x);
+          for (size_t i = 0; i < n_x; ++i) {
+            this->log_sd[i] = tmp;  // copies all fields in Param
+          }
```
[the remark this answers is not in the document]
Andrea-Havron-NOAA: There are two issues. First, `get_force_scalar` is a function designed for the `fims::Vector` class, so it won't work with `this->log_sd`, which is a `ParameterVector`, but we would add a `get_force_scalar` function to the `ParameterVector` class. The main issue, though is with the code logic. The code needs to resize `this->log_sd` to `n_x` prior to starting the loop, so `get_force_scalar` would see a vector of length `n_x` and pull from the respective `i`, not from the first position, `0`. I added a comment in front of the loop to explain what the code is doing for better clarity: //If input log_sd is a scalar, resize to n_x and fill with the scalar value. I did go ahead and update line: `this->log_sd[i].final_value_m = dnorm->log_sd[idx];` to use `get_force_scalar` instead.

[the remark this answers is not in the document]
kellijohnson-NOAA: I think we should leave it as is and revisit what we can do after the revamp given density component changes.

inst/include/interface/rcpp/rcpp_objects/rcpp_distribution.hpp
```diff
@@ -402,6 +419,17 @@ class DnormDistributionsInterface : public DistributionsInterfaceBase {
...
+    if (this->log_sd.size() == 0) {
+      ss << "],\n";
+    } else {
+      for (R_xlen_t i = 0; i < this->log_sd.size() - 1; i++) {
+        ss << this->value_to_string(this->log_sd[i].final_value_m) << ", ";
+      }
+      ss << this->value_to_string(
+                this->log_sd[this->log_sd.size() - 1].final_value_m)
+         << "],\n";
+    }
```
kellijohnson-NOAA: Need to check this for the formatting in the output and the ability to read it back in.

inst/include/interface/rcpp/rcpp_objects/rcpp_distribution.hpp
```diff
@@ -719,6 +762,17 @@ class DlnormDistributionsInterface : public DistributionsInterfaceBase {
...
```
kellijohnson-NOAA: need to check this in the json and the ability to read it into the tibble

inst/include/interface/rcpp/rcpp_objects/rcpp_distribution.hpp
```diff
@@ -953,15 +1007,16 @@ class DmultinomDistributionsInterface : public DistributionsInterfaceBase {
...
+      size_t n_x = dmultinom->report_lpdf_vec.size();
+      this->lpdf_vec = Rcpp::NumericVector(n_x);
+      if (this->expected_values.size() != n_x) {
+        this->expected_values.resize(n_x);
       }
-      if (this->x.size() != nx) {
-        this->x.resize(nx);
+      if (this->x.size() != n_x) {
+        this->x.resize(n_x);
       }
```
kellijohnson-NOAA: There is no warning or catch statement in dmultinom like there is in dnorm and dlnorm.

inst/include/interface/rcpp/rcpp_objects/rcpp_distribution.hpp
```diff
@@ -970,8 +1025,8 @@ class DmultinomDistributionsInterface : public DistributionsInterfaceBase {
...
       if (dmultinom->input_type == "data") {
         dims.resize(2);
-        dims[0] = dmultinom->observed_values->get_imax();
-        dims[1] = dmultinom->observed_values->get_jmax();
+        dims[0] = dmultinom->dims[0];
+        dims[1] = dmultinom->dims[1];
```
kellijohnson-NOAA: where does dims come from?

R/reshape_output.R
```diff
@@ -401,8 +401,19 @@ dimensions_to_tibble <- function(data) {
...
+  }
+  df <- tidyr::unnest(df, c(value, expected_values, observed_values))
+
+  # Only unnest log_sd_values if it is not all NULL
+  if (!all(sapply(df$log_sd_values, is.null))) {
+    df <- tidyr::unnest(df, log_sd_values, keep_empty = TRUE)
+  }
+
+  dplyr::rename(df, likelihood = value)
+  df
```
kellijohnson-NOAA: 418 needs to be removed.

inst/include/interface/rcpp/rcpp_objects/rcpp_distribution.hpp
kellijohnson-NOAA: Are there any tests for this?

kellijohnson-NOAA (approved):
Andrea, I have a few comments but I think this looks good.

kellijohnson-NOAA:
There are still failed tests that appear to have a segfault in them. Though, I am not getting any output other than the following
```
Saving _problems/test-get_estimates-48.R
Saving _problems/test-get_estimates-88.R
Saving _problems/test-integration-fims-bayesian-prior-predictive-252.R
Segmentation fault (core dumped)

1 error ✖ | 2 warnings ✖ | 4 notes ✖
```
I will run them locally again.

PR #1130, opened 2026-01-12
test(integration with wrappers): add test for mixed Fmort estimation …

Bai-Li-NOAA:
# What is the feature?
* add a test for mixed estimation types (Fmort) in catch-at-age model
* It addresses the suggestions in issue #852 

* Added an integration test in `tests/testthat/test-integration-caa-mle-wrappers.R`
* Configured fleet1's Fmort to be constant for the years 1 to 10 and estimated using `fixed_effects` for years 11 to 30.

* No.


Bai-Li-NOAA:
@kellijohnson-NOAA Sorry for the late reply. I was going to address the comment this morning, but I see you’ve already added the test. Thanks for taking care of it!

PR #1166, opened 2026-01-21
fix(GHA): add 16 GB swap for parallel tests on Linux

Bai-Li-NOAA:
# What is the feature?
* fix `.github/workflows/run-slow-tests.yml` to resolve R termination on Linux and lockfile creation failure on macOS.

* Allocated 16 GB swap to prevent out-of-memory issues when running FIMS models in parallel
* Added `GITHUB_PAT` to avoid failure creating lockfile '.github/pkg.lock' on macOS

*

### DIFF
--- a/.github/workflows/run-slow-tests.yml
+++ b/.github/workflows/run-slow-tests.yml
@@ -31,9 +31,27 @@ jobs:
     
     env:
       RUN_SLOW_TESTS: "true"
+      GITHUB_PAT: ${{ secrets.PAT }}
 
     steps:
       - uses: actions/checkout@v6
+      
+      # Current Ubuntu runners lack enough RAM for parallel tests.
+      # The step below creates a 16 GB swap file to prevent R processes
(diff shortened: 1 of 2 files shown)


kellijohnson-NOAA (approved):

k-doering-NOAA:
@Bai-Li-NOAA nice work figuring this out, cryptic log message and all!

@kellijohnson-NOAA asked:
> Are the current actions that we are running free because we are in a public repository and they are on low-powered machines?

Yes, exactly! I think we also have included minutes for private repositories and low powered machines with our GitHub enterprise plan.

Currently, we have a $0 budget set for GitHub actions because we've never had a use case that needed to use larger runners - we could change this, and the billing would be set up already - but we probably need to clear this with Patrick and Melissa.

@Bai-Li-NOAA what's the benefit of using a larger runner over the memory swap?

k-doering-NOAA:
I just emailed Patrick, Melissa, and Abby about this funding topic. Do you want me to loop you in?

Issue #576, opened 2024-05-02
[Feature]: ability to zero out specific ages from selectivity curve


ChristineStawitz-NOAA:
### Is your feature request related to a problem? Please describe.

@Cole-Monnahan-NOAA asked for the ability to zero out ages 1 and 2 in the double-logistic selectivity curve 

### Describe the solution you would like.

Ability to zero certain ages within a selectivity curve

### Describe alternatives you have considered

no ability to zero out ages

Andrea-Havron-NOAA:
@Cole-Monnahan-NOAA,  could you provide a use case example?

iantaylor-NOAA:
SS3 has this option as well. I believe the most common use case is for species which are never seen at the youngest ages, even when there are big recruitment events, whereas a smooth curve for ascending selectivity will always have a non-zero probability of selecting all ages.

Cole-Monnahan-NOAA:
In my case I have a survey on a spawning aggregation. Immature fish are not expected to spawn and thus are not reliably seen. So in my code I estimate a selectivity curve, but then zero out the first two ages and there is no likellihood component associated with those ages. An alternative would be to leave the selex curves the same but redefine the composition to be without those ages.

iantaylor-NOAA:
I don't remember a discussion of age or length bin specification for FIMS, but I like the idea of a flexible bin structure that could be different for each fleet as suggested in the alternative from @Cole-Monnahan-NOAA above.

nathanvaughan-NOAA:
@iantaylor-NOAA  and @Cole-Monnahan-NOAA  We are working on the comp data format while adding length structure so I think it would be fairly simple to incorporate comp likelihood calculations that ignore ages/lengths that are not included in the observed data. We report back here on alternatives once we have things more fleshed out. We could also account for this with something like the SS retention curve where we layer an availability pattern that could zero out some ages?

Cole-Monnahan-NOAA:
I just remembered that this approach likely won't work because if you have selex>0 for those bins then that will contribute to the expected biomass index which we don't want because that would bias it high. In my case it's an acoustic survey so the selex is estimated at 1 for those two ages so it would be exaggerated.

PR #1170, opened 2026-01-21
Dev descending logistic selectivity

grantdadams:
# What is the feature?
* Descending logistic age-based selectivity

* Added `descending_selectivity` to selectivity functors for use in TMB
* Added `DescendingLogisticSelectivity` to methods for use in R
* Selecting `DescendingLogistic` will work exactly the same as `Logistic` or `DoubleLogistic` when selecting a selectivity `module_type`
* Added tests to `test-rcpp-selectivity` to make sure rcpp descending logistic selectivity works with correct inputs

* Not that I know of

### DIFF
--- a/inst/include/population_dynamics/selectivity/functors/descending_logistic.hpp
+++ b/inst/include/population_dynamics/selectivity/functors/descending_logistic.hpp
@@ -23,64 +23,64 @@ namespace fims_popdy {
 template <typename Type>
 struct DescendingLogisticSelectivity : public SelectivityBase<Type> {
   fims::Vector<Type>
-  inflection_point_desc;     /**< 50% quantile of the value of
- the quantity of interest (x) on the descending limb of the Descending
- logistic curve; e.g. age at which 50% of the fish are selected */
-fims::Vector<Type> slope_desc; /**<scalar multiplier of difference between
- quantity of interest  value (x) and inflection_point on the
- descending limb of the Descending  logistic  curve */
+      inflection_point_desc;     /**< 50% quantile of the value of
+     the quantity of interest (x) on the descending limb of the Descending
+     logistic curve; e.g. age at which 50% of the fish are selected */
+  fims::Vector<Type> slope_desc; /**<scalar multiplier of difference between
+   quantity of interest  value (x) and inflection_point on the
+   descending limb of the Descending  logistic  curve */
 
-DescendingLogisticSelectivity() : SelectivityBase<Type>() {}
+  DescendingLogisticSelectivity() : SelectivityBase<Type>() {}
 
-virtual ~DescendingLogisticSelectivity() {}
+  virtual ~DescendingLogisticSelectivity() {}
 
-/**
- * @brief Method of the Descending logistic selectivity class that implements
- * the Descending logistic function from FIMS math.
- *
- * \f$\left(1.0-\frac{1.0}{ 1.0 + exp(-1.0 * slope\_desc (x -
- * inflection_point\_desc))} \right)\f$
- *
- * @param x  The independent variable in the descending logistic function
- * (e.g., age or size in selectivity).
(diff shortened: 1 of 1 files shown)


kellijohnson-NOAA:
Thanks @grantdadams for this PR. I looked at the code and in general it looks great. Some logistical stuff though, we use a linear commit history in FIMS, i.e., NO merge commits. Merges makes it hard to see which commits are part of which features, especially when the PR evolves over time and the commits cannot be squashed before they are brought into dev and then main. Thus in the future, to rebase to dev you will need to run the following while on your feature branch, which I named "dev-feature-branch" in the example but obviously you would put your branch name there
```
git checkout dev
git fetch -a
git rebase origin/dev
git checkout dev-feature-branch
git rebase origin/dev
git push -f origin dev-feature-branch
```
There is also an option to do "rebase and merge" when you are bringing in any PR into your feature branch that needs to be used. I know that this is a huge pain but it really does help keep the linear commit history, which I think is very important.

Unfortunately, the above workflow is very difficult after you have already committed merge commits. Thus, I went the cherry-pick route and I was able to "rebase" your branch to dev on my local machine. I am happy to push that up if you would like but if you are to do any more work on your machine with it you will have to run
```
git checkout dev-descending-logistic
git fetch -a
git reset --h origin/dev-descending-logistic
```

kellijohnson-NOAA:
I am wondering if your PR addresses https://github.com/NOAA-FIMS/FIMS/issues/576 at all? I am assuming that the descending limb would start at the first age or length with your code and that users do not have a way to adjust it to zero out age 0 and age 1? 

Also, we chatted about the potential for not needing the descending logistic because you can fix the ascending slope to be Inf and get the same result because
$$(\frac{1}{1+e^{(-(x-x_0)*k)}}) * [1-(\frac{1}{1+e^{(-(x-x_0)*k)}})]$$
becomes
$$(\frac{1}{1+e^{(-(x-x_0)*Inf)}}) * [1-(\frac{1}{1+e^{(-(x-x_0)*k)}})]$$
$$(\frac{1}{1+e^{-Inf}}) * [1-(\frac{1}{1+e^{(-(x-x_0)*k)}})]$$
$$(\frac{1}{1+0}) * [1-(\frac{1}{1+e^{(-(x-x_0)*k)}})]$$
$$[1-(\frac{1}{1+e^{(-(x-x_0)*k)}})]$$
which is what Grant has for the descending logistic equation.
I think that we need to have a larger FIMS discussion about this. Do you see utility in not requiring the user to specify two additional parameters that they do not need, which I think is good but I do not know if it comes at a benefit of the amount of code that was added. I think it would be great if we could do some speed, memory, or other tests to show the benefit of one method or the other. A third method would be to keep the user interface to the descending logistic but fix the parameter behind the scenes for them in the double logistic.

@msupernaw @grantdadams @iantaylor-NOAA @JonBrodziak @MOshima-PIFSC @timjmiller @nathanvaughan-NOAA @Andrea-Havron-NOAA do any of you have opinions about this?

nathanvaughan-NOAA:
Hey @grantdadams I've got some good news and some bad news. The good news is you did a great job digging through FIMS and finding all the files you need to change to add a new module, so you're ready to go for the next one. The bad news is I'm pretty sure that the logistic function with a negative slope value is already the descending logistic function so FIMS can already fit both scenarios with the one function without any extra fixed parameters as mentioned in the earlier comments. Let me know if I'm missing something but otherwise I would not recommend adding in this extra complexity to FIMS.

grantdadams:
Gah, stuck in my `exp(ln_slope)` ways, I didn't even think of that! That makes sense to me, thanks for the review!

kellijohnson-NOAA:
I am going to close this PR because we have a new issue open #1197 for documenting the ability to use a negative slope.

PR #1193, opened 2026-02-12
docs(catch_at_age): add snippets to reuse shared parameters across functions

szu-yun-ko:
# What is the feature?
* This addresses the discussion in PR #1137 about using Doxygen's `\snippet{doc}` function to reuse shared parameters across functions in `catch_at_age.hpp`. 
* Note: this is currently still a draft to test whether the approach works. So far I've only implemented it on `CalculateMortality()` for testing. 

* At the top of `catch_at_age.hpp`, a dictionary block was declared to document all the shared parameters across the functions. Then in each function, simply use `@snippet{doc} this snippet_id` to include the parameter. 

   ```cpp
  /* Dictionary block for shared parameter snippet documentations.
   * Referenced in function docs via @snippet{doc} this snippet_id.
   * [param_population]
   * @param population Shared pointer to the population object.
   * [param_population]
  ...
   */
    
  /**
  @snippet{doc} this param_population
  ...
   */
  void CalculateMortality(
      std::shared_ptr<fims_popdy::Population<Type>> &population,
      size_t i_age_year, size_t year, size_t age)
   ```
* A small change to what was discussed in PR#1137 was that `\noop` cause compilation errors when I tried to compile the documentation. So I didn't use `\noop`, but found that simply using `//` was sufficient for the task.

* No additional files were created. No code changes were made. Might affect the deployed website.

### DIFF
--- a/inst/include/models/functors/catch_at_age.hpp
+++ b/inst/include/models/functors/catch_at_age.hpp
@@ -27,6 +27,15 @@
  * [param_age]
  * @param age Age index.
  * [param_age]
+ * [param_i_agem1_yearm1]
+ * @param i_agem1_yearm1 Dimension folded index for age-1 and year-1.
+ * [param_i_agem1_yearm1]
+ * [param_i_dev]
+ * @param i_dev Index to log_recruit_dev of vector length n_years-1.
+ * [param_i_dev]
+ * [param_other]
+ * @param other The other CatchAtAge object to copy from.
+ * [param_other]
  */
 
 namespace fims_popdy {
@@ -122,7 +131,7 @@ class CatchAtAge : public FisheryModelBase<Type> {
   /**
    * @brief Copy constructor for the CatchAtAge class.
    *
-   * @param other The other CatchAtAge object to copy from.
+   * @snippet{doc} this param_other
    */
   CatchAtAge(const CatchAtAge &other)
       : FisheryModelBase<Type>(other), name_m(other.name_m), ages(other.ages) {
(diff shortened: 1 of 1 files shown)


inst/include/models/functors/catch_at_age.hpp
```diff
@@ -260,8 +260,8 @@ class CatchAtAge : public FisheryModelBase<Type> {
    * N_{a,0} = \exp(\log N_{a,0})
    * \f]
    *
-   * @param population Shared pointer to the population object.
-   * @param i_age_year Dimension folded index for age and year.
+   * @snippet{doc} this param_population
+   * @snippet{doc} this param_i_age_year
    * @param a Age index.
```
szu-yun-ko: Here the age parameter is aliased with `a` instead of `age` so I kept the original `@param` for now to avoid making changes to the code. If a change is preferred here please let me know and I'll incorporate that.

szu-yun-ko:
> Would it work if we used snippet for those params that are failing even if we are not sharing the code across multiple functions right now?
> 
> Also, go ahead and change the parameter names for consistency where you think that they need changed.

Yes. I think it would work if snippets were used for all params (it worked earlier when only `CalculateMortality()` was implemented). 

> Second, can you add the following formulas? CalculateNumbersAA needs this formula needs added to the plus group calculation
> 
> ```
>    * N_{A,y} = N_{A-1,y-1} \exp(-Z_{A-1,y-1}) + N_{A,y-1} \exp(-Z_{A,y-1})
> ```
> 
> CalculateUnfishedNumbersAA needs this formula needs added to the plus group calculation
> 
> ```
>    * N^U_{A,y} = N^U_{A-1,y-1} \exp(-M_{A-1,y-1}) + N^U_{A,y-1} \exp(-M_{A,y-1})
> ```

Sure. I can make these changes within these two days.

szu-yun-ko:
@kellijohnson-NOAA Now that `\snippet` is used for all parameters, the run-doxygen check passed successfully! I've also included the expanded equations for `CalculateNumbersAA` and `CalculateUnfishedNumbersAA`. This is ready for review whenever you have time. Thanks!

kellijohnson-NOAA:
Great, I see that two pull requests are failing because they just don't work on forks and one is failing b/c we have yet to figure out globally how to fix it 🤣 so I will review the code and hopefully get it merged in before this next release goes out. Thank you for your contributions and we look forward to working with you again.
Side note, I pattern matched the snippet code that you wrote and used it successfully elsewhere in the code in the dev-teamwork branch. So, thank you again for investigating this.

szu-yun-ko:
That's great news to hear, glad that the code turned out to be helpful elsewhere too! Thanks for taking the time to review this. Please feel free to reach out if any help is needed going forward. I’m happy to keep helping with related work or other tasks as well!

PR #1204, opened 2026-02-18
add stockplotr from R universe

Bai-Li-NOAA:
# What is the feature?
* Fix the failed GHA run that occurs when R-Universe builds the vignettes

* Added the toolbox R Universe to `AdditionalRepositories` so that {stockplotr} can be installed from R Universe during the build process. 

* No

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -104,6 +104,8 @@ Remotes:
     github::nmfs-ost/stockplotr,
     github::ThinkR-open/lozen,
     github::ThinkR-open/testdown
+Additional_repositories: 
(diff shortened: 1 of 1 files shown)


kellijohnson-NOAA:
I just regenerated the FIMS-PAT a few days ago. And, it is linked to nmfs-ost so we should be good.

Also, in response to {pak} not using the `Additional Repositories` section, see [this comment](https://github.com/r-lib/remotes/issues/618#issuecomment-2139245222) on the remotes package that verifies that the "Additional Repositories" field does NOTHING 😢.

PR #1217, opened 2026-02-20
Dev fix memory issue

### LINKED ISSUE #1139: [Developer Issue]: Models hitting memory limits
### Description

During the development of the projections feature branch, FIMS runs were taking up large amounts of memory (10's of GB's) even with relatively simple models 30 year time series and 30 year projections. This was enough to crash the GitHub tests for macOS that have a cap of 16GB. Ideally, projections would be 100+ years, and many assessments in the SE have very long historic time series (100's years), which would be impractical to run with this amount of memory usage. A first guess is that eliminating redundancy in reporting finalize/sdreport could minimize some of this issue, but that seems like it would only be a 2-3x improvement at best. We ideally need an order of magnitude or more reduction for FIMS to be scalable to much more complex scenarios.

msupernaw:
# What is the feature?
Adds FreeAdFun to get_report

* Adds FreeAdFun to get_report

* 
no, reduces memory consumption when running multiple model simultaneously.


kellijohnson-NOAA (approved):

inst/include/interface/rcpp/rcpp_objects/rcpp_models.hpp
```diff
@@ -779,7 +808,7 @@ class CatchAtAgeInterface : public FisheryModelInterfaceBase {
       grouped_out = Rcpp::List::create();
       // first_est is already initialized to 0.0
     }
-
+    freeADFun(obj);
```
kellijohnson-NOAA:
```suggestion
    FreeADFun(obj);
```

kellijohnson-NOAA (changes requested):
I approved this but it appears to be failing the R CMD check and testthat is not providing any output why, which leads me to believe we have segfault problems somewhere.

PR #1221, opened 2026-02-22
Work from 2026-02-04 FIMS Implementation Team Meeting

kellijohnson-NOAA:
# What is the feature?
* documentation and clean up of the code base to get ready for the CIE Review

* Broke into several groups to complete the work. Looked at R CMD check to see additional tasks that needed to be worked on. Tried to remove the remaining compiler errors on Windows but a few still remain.

*


kellijohnson-NOAA:
@szu-yun-ko do you have time to look at why my use of snippets did not work in [‎inst/include/interface/rcpp/rcpp_interface.hpp](https://github.com/NOAA-FIMS/FIMS/pull/1221/changes/e2f2e53a5d9b4a6e4f7aa21635c0dede6805801f#diff-76ce540f3f4a0d140385d4a51dcb7bceff3b20586fe98beb242115baefa8e098) within e2f2e53a5d9b4a6e4f7aa21635c0dede6805801f, which is a commit in this PR. I was getting weird warnings when I had * in front of each line. The commit passed the doxygen rendering on Windows but then when I pushed the linux action was failing. So, then I changed to using @page in this commit which passes on Windows and Linux. I would rather use your snippet method to keep things consistent but I just did not get it to work on Linux for this application. Any suggestions would be greatly appreciated.

szu-yun-ko:
@kellijohnson-NOAA Congrats!! I had no idea that asterisks shouldn’t be used in snippets on Linux (and also had no idea how that worked on CI for the other two PRs— learned something new today 🫢). Glad it worked out! Thanks for updating my earlier changes as well, this will probably make the CI more stable too.

PR #1257, opened 2026-03-01
fix: assign unlist() result in dimension_folded_to_tibble()

### LINKED ISSUE #1256: [Bug]: Potential infinite loop in dimension_folded_to_tibble() due to unassigned unlist() result
#Description:
In R/reshape_output.R (lines 301-302), the while loop in dimension_folded_to_tibble() never assigns the result of unlist() back to section:

while (length(section) == 1) {
  unlist(section, recursive = FALSE)  # result is discarded
}
Since section is never modified inside the loop, if length(section) == 1 is true on entry, the loop runs forever.

#Expected Behavior:
The result should be assigned back:

while (length(section) == 1) {
  section <- unlist(section, recursive = FALSE)
}

#Steps to Reproduce:

Call dimension_folded_to_tibble() with a list of length 1 — the function will hang indefinitely.

Eshaan-byte:
#Description:
The while loop called unlist(section, recursive = FALSE) without assigning the result back to section, causing an infinite loop when length(section) == 1.

#What is the feature?
Bug fix: The while loop in dimension_folded_to_tibble() (R/reshape_output.R, line 302) discarded the return value of unlist(), so section was never modified and the loop would hang indefinitely when passed a length-1 list.

#How have you implemented the solution?
Added the assignment section <- so the result of unlist(section, recursive = FALSE) is stored back into section, allowing the loop to terminate.

#Does the PR impact any other area of the project, maybe another repo?
No. This is a single-line fix scoped to dimension_folded_to_tibble() in R/reshape_output.R. No other files or repos are affected.

Close #1256

### DIFF
--- a/R/reshape_output.R
+++ b/R/reshape_output.R
@@ -299,7 +299,7 @@ dimension_folded_to_tibble <- function(section) {
     return(NA)
   }
   while (length(section) == 1) {
-    unlist(section, recursive = FALSE)
+    section <- unlist(section, recursive = FALSE)
   }
   temp <- dimensions_to_tibble(section[["dimensionality"]]) |>
     dplyr::mutate(name = section[["name"]])


kellijohnson-NOAA (approved):
I wish I had some example to test this on but all I have is my memory from like two years ago when I wrote it that I wanted the code to act like how you have fixed it to. So, we will go with that. Thanks again!

PR #1272, opened 2026-03-06
Dev removes C++ calculation of sdreport

### LINKED ISSUE #1266: [Developer Issue]: Remove all traces of getting sdreport within C++ code
### Description

We need to remove the duplication of getting the sdreport. Right now, v.0.8.1, there is code in the C++ to get the sdreport and code in R. We want to keep the code in R `sdreport()` and `reshape_tmb_estimates()` but we want to remove the code from C++. We also should remove anything related to uncertainty and gradient in the json file as this information will come from `reshape_tmb_estimates()`.
One part that I am uncertain about is log_f_multiplier and spawning_biomass_ratio. @nathanvaughan-NOAA will we need to add these to the derived quantity list or are they classified as parameters because of the desire to set them in the forecast? I just know that they were added after the addition of the sdreport in the C++ so I am not sure if the pattern matching that happened there will be correct or not.

kellijohnson-NOAA:
# What is the feature?
* Removes, hopefully, everything related to internally (within the C++) calculating the uncertainty of parameters and derived quantities. Also removes `calculate_reference_points()` function in C++ because it was not being used anywhere.

* A helper function in R, `add_unique_id ()`, was added to facilitate merging the tmb and json output together, it assumes that all of the derived quantities within a type, e.g., biomass, are in the same order in the tmb as they are in the json.
* Duplicated code inside of `FIMSFit()` was also removed because we were not reporting the estimates from FIMSFit as a slot, instead there is a method, `get_estimates()` that uses information in the slots to get the estimates.
* `min`, `max`, and `uncertainty` were removed from the json output because they are no longer needed, well min and max were never needed because we do not have bounds. A test was changed to count the number of NA values instead of the number of 0 values for counting the correct number of constant parameters.
* Removes Uncertainty words from the word list
* Removes C++ functions:
  * `create_report_vectors()`
  * `get_report_vector_count()`
  * `create_report_vectors_map()`
  * `calculate_reference_points()`
  * `fill_min()`
  * `fill_max()`
  * `set_uncertainty()`
  * `get_se_values()`
  * `get_report()`
* Removes C++ structures:
  * UncertaintyReportInfo
  * UncertaintyReportInfoMap
  * GetFleetUncertaintyReportInfo
  * GetPopulationUncertaintyReportInfo
* Removes `so_sd_report` argument from C++ function `get_output()`
* Removes sd logic from `to_json()`
* Removes all parameters from the TMB reporting because they are automatically reported when they are parameters and having them there was causing things like selectivity and log_m to be reported twice when estimated
* Commented out tests regarding recruitment deviations being zero until we revamp how those can be estimated, this PR does not actually change anything except they are no longer reported in the sd report to be able to check if they are in fact zero.
* Adds "weight_at_age" as the name in the json for weight-at-age so the label is propagated in get_estimates
* Removes some code that was commented out in inst/include/models/functors/catch_at_age.hpp regarding uncertainty

* Some of the case studies might be affected if setting min or max somewhere but this should not be the case
* If a power user in the case study was getting information out of the report in a way that is now removed we will need to move to get_estimates() output instead

___

Hello reviewer, thanks for taking the time to review this PR!

- Please use this checklist during your review, checking off items that you have verified are complete, but feel free to skip over items that are not relevant!
- Please use conventions in the [guidelines for conventional commit messages](https://www.conventionalcommits.org/) for both commit messages and comments.
- Engage with the developer when they respond to comments and check off additional boxes as they become complete so the PR can be merged in when all the tasks are fulfilled. Make it clear when the PR is approved by selecting the approved status, and potentially by commenting on the PR with something like `This PR is now ready to be merged, no additional changes are needed`.

## Checklist

- [ ] The PR is requested to be merged into the appropriate branch (typically dev).
- [ ] The code is well-designed.
- [ ] The functionality is good for the users of the code.
- [ ] Code coverage remains high, indicating the new code is tested.
- [ ] The code is commented and the comments are clear, useful, and mostly explain why instead of what.
- [ ] Code is appropriately documented (doxygen and roxygen).

Closes #1266

### DIFF
--- a/R/fimsfit.R
+++ b/R/fimsfit.R
@@ -222,7 +222,7 @@ methods::setMethod(
               seq_len(dplyr::n()),
               parameter_id
             ),
-            sep =  "_"
+            sep = "_"
           )
         ) |>
         dplyr::ungroup()
--- a/tests/testthat/test-projections-looped.R
+++ b/tests/testthat/test-projections-looped.R
@@ -654,11 +654,11 @@ estimation_error <- max(abs(sdr_fixed_5_year_project_catch_low[-c(33:37), "Estim
 
 sd_error <- max(abs(sdr_fixed_5_year_project_catch_low[-c(33:37), "Std. Error"] - sdr_fixed_no_project[, "Std. Error"]) / abs(sdr_fixed_no_project[, "Std. Error"]))
 
-#rec_devs <- sdr_report_5_year_project_catch_low[row.names(sdr_report_5_year_project_catch_low) == "log_devs", ][30:34, ]
+# rec_devs <- sdr_report_5_year_project_catch_low[row.names(sdr_report_5_year_project_catch_low) == "log_devs", ][30:34, ]
 
 test_that("projections with low catch data achieve same estimates and no projection model run", {
   #' @description Test that rec devs were fixed at zero in projection.
-  #expect_equal(sum(rec_devs[, "Estimate"]), 0)
+  # expect_equal(sum(rec_devs[, "Estimate"]), 0)
 
   #' @description Test that the maximum parameter estimate difference between a low catch projection run and no projection run is less than 10%.
   expect_lt(estimation_error, 0.1)
--- a/vignettes/fims-logging.Rmd
+++ b/vignettes/fims-logging.Rmd
@@ -95,7 +95,6 @@ log_warnings_str <- get_log_warnings()
 
 # get log errors only
 log_errors_str <- get_log_errors()
-
 ```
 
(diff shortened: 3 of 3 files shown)


Andrea-Havron-NOAA:
@kellijohnson-NOAA, I was looking over the code and I don't think the json output and TMB output will be in the same order. The order of reported values from TMB can be found in catch_at_age in the [ADREPORT section](https://github.com/NOAA-FIMS/FIMS/blob/e1f58085a5b4c7e4ad55a296da3f5ac995becf36/inst/include/models/functors/catch_at_age.hpp#L1385), while *I think* the order in json will be based on the order of how the derived quantities are intialized [here](https://github.com/NOAA-FIMS/FIMS/blob/e1f58085a5b4c7e4ad55a296da3f5ac995becf36/inst/include/interface/rcpp/rcpp_objects/rcpp_models.hpp#L1204)

PR #1297, opened 2026-03-13
change log_devs default to random_effects

nathanvaughan-NOAA:
# What is the feature?
* Change default model setup from fixed effects recruitment deviations to random effects recruitment deviations. For now log_sd is still fixed not estimated.

* Simple change in the default configuration. We will work on cleaning up the parameters tibble in a future branch.

* No


nathanvaughan-NOAA:
@kellijohnson-NOAA  it's not really a test itself, but the original tests were failing because of the random effects being on (i.e. tests of pars$p were the wrong length because parameters had moved to pars$re) let me look and see if there is something I can add to do the same thing intentionally.

PR #1321, opened 2026-03-18
Fix scalar log_M handling in catch-at-age output

### LINKED ISSUE #1084: [Bug:] input and output dimensions of log_M do not always match
### Describe the bug

I recently set up a test using helper functions and accidently added log_M twice to my tibble, increasing its dimension to 720 instead of 360. Since log_M is a fixed parameter by default, the model was able to run, but crashed when calling reshape_json_estimates(). This happened because the model's log_M parameter size is defined by the size of the input ([here](https://github.com/NOAA-FIMS/FIMS/blob/1276b24e7aca707bd366adbc8329fda4747f4d09/inst/include/interface/rcpp/rcpp_objects/rcpp_population.hpp#L339))  but the output size is structured by nyears and nages defined in population ([here](https://github.com/NOAA-FIMS/FIMS/blob/1276b24e7aca707bd366adbc8329fda4747f4d09/inst/include/interface/rcpp/rcpp_objects/rcpp_models.hpp#L276)). This was a challenging bug to track down as it is currently not possible to look at fixed parameter inputs to the model via the helper functions.

This setup raises a few issues:

- [ ] FIMS needs to dimension check all parameters, fixed, estimated, and random effects
- [ ] Is there a way to print a summary of all fixed, estimated, and random parameters from the model, along with their dimensions before the model is run?
- [ ] log_M is not always nyears x nages. Users may want to estimate a single log_M value for the entire population
    - [ ] log_M[ i_age_year] in [catch_at_age.hpp](https://github.com/NOAA-FIMS/FIMS/blob/1276b24e7aca707bd366adbc8329fda4747f4d09/inst/include/models/functors/catch_at_age.hpp#L175C42-L175C59) needs to be updated to log_M.get_force_scalar(i_age_year)
    - [ ] The dimensions of the output should be defined by the dimensions of the input. 
- [ ] Other parameters in the model should be checked for similar issues 

### Describe how to reproduce the bug

```
devtools::load_all()
load(test_path("fixtures", "integration_test_data.RData"))
# Load sample data
data("data1")
# Prepare data for FIMS model
data_4_model <- FIMSFrame(data1)

# Create parameters
default_parameters <- data_4_model |>
  create_default_configurations() |>
  create_default_parameters(data = data_4_model)

parameters_4_model <- default_parameters |>
  tidyr::unnest(cols = data) |>
  # Update log_obs_sd initial values for Fleet1
  dplyr::filter(
    !(fleet_name == "fleet1" & label == "log_sd" & time > 1)
  ) |>
  dplyr::bind_rows(
    default_parameters |>
      tidyr::unnest(cols = data) |>
      dplyr::filter(is.na(fleet_name))
  )

test_fit <- parameters_4_model |>
  initialize_fims(data = data_4_model) |>
  fit_fims(optimize = FALSE)
```

### Screenshots

_No response_

Linux

_No response_

dev branch

### Additional context

_No response_

awilnoaa:
# What is the feature?
* Fixes handling of scalar `log_M` in catch-at-age model output.

* Updated catch-at-age calculations to use `get_force_scalar()` when consuming population natural mortality values.
* Updated `log_M` and related mortality output dimensions to reflect the actual `log_M` input shape instead of always reporting full age-year dimensions.
* Fixed JSON header quoting and type conversion issues needed to support the new scalar dimension reporting.

* No expected impact outside of FIMS catch-at-age `log_M` handling and its JSON/model output.

Addresses Issue #1084 

___

## Checklist

### DIFF
--- a/R/create_default_parameters.R
+++ b/R/create_default_parameters.R
@@ -304,15 +304,11 @@ create_default_Population <- function(
   init_naa[n_ages] <- init_naa[n_ages] / M_value # sum of infinite series
 
   # Create a list of default parameters
-  default <- create_default_parameters_template(
-    n_parameters = n_years * n_ages
-  ) |>
+  default <- create_default_parameters_template() |>
     # Add the module type, label, value, and estimation type
     dplyr::mutate(
       label = "log_M",
       value = log(M_value),
-      age = rep(ages, n_years),
-      time = rep(years, each = n_ages),
       estimation_type = "constant"
     ) |>
     dplyr::add_row(
--- a/inst/include/interface/rcpp/rcpp_objects/rcpp_models.hpp
+++ b/inst/include/interface/rcpp/rcpp_objects/rcpp_models.hpp
@@ -365,9 +365,17 @@ class CatchAtAgeInterface : public FisheryModelInterfaceBase {
       ss << " \"id\":" << population_interface->log_M.id_m << ",\n";
       ss << " \"type\": \"vector\",\n";
       ss << " \"dimensionality\": {\n";
-      ss << "  \"header\": [\"na\"],\n";
-      ss << "  \"dimensions\": [" << population_interface->log_M.size()
-         << "]\n},\n";
+      if (population_interface->log_M.size() ==
+        population_interface->n_years.get() * population_interface->n_ages.get()) {
+        ss << " \"header\": [\"n_years\", \"n_ages\"],\n";
+        ss << " \"dimensions\": [" << population_interface->n_years.get() << ", "
+        << population_interface->n_ages.get() << "]\n";
+      } else {
+        ss << " \"header\": [\"scalar\"],\n";
+        ss << " \"dimensions\": [" << population_interface->log_M.size()
+        << "]\n";
+      }
+      ss << "},\n";
       ss << " \"values\": " << population_interface->log_M << "\n\n";
       ss << "},\n";
 
@@ -1006,7 +1014,7 @@ class CatchAtAgeInterface : public FisheryModelInterfaceBase {
           fims::Vector<std::string>{"n_years", "n_ages"});
 
       derived_quantities["mortality_M"] = fims::Vector<Type>(
-          population->n_years.get() * population->n_ages.get());
+          this->log_M.size());
       fims::Vector<std::string> dim_names;
       if (population->log_M.size() ==
           population->n_years.get() * population->n_ages.get()) {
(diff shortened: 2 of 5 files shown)


nathanvaughan-NOAA (changes requested):
Assuming that we want the scalar log_M because it speeds things up, should we also change the default parameter setup to create a scalar parameter rather than a replicated n_years*n_ages vector? The setup is on lines 277:287 in create_default_parameters.R. That would also help test/confirm that the changes don't have any bugs.

Andrea-Havron-NOAA:
> @nathanvaughan-NOAA we went back and forth a bit while Andrea and Adrianne were working on this to try and decide if M should be reported for the entire time x age series if only a single value is estimated and because selectivity we thought that it should match what is done for selectivity where only a single value is reported if it does not vary with time. I am willing to rethink this globally for all parameters but we cannot get that rethink done before the CIE Review. So, I propose that we have M be reported as a single value when only a single value is estimated. @Andrea-Havron-NOAA I think that Nathan still found some changes that need to be made though. Do you have time today to edit this PR?

@kellijohnson-NOAA, yes I can work on this today.

nathanvaughan-NOAA:
Sounds good to me @kellijohnson-NOAA , unless I made any mistakes I think you should be able to just copy in the code chunks in my comments @Andrea-Havron-NOAA and this will be good to go.

Andrea-Havron-NOAA:
> Sounds good to me @kellijohnson-NOAA , unless I made any mistakes I think you should be able to just copy in the code chunks in my comments @Andrea-Havron-NOAA and this will be good to go.

I copied the code over and just fixed some minor bugs. Should I go ahead and change the default log_M to be a scalar in the parameter tibble?

nathanvaughan-NOAA (approved):

kellijohnson-NOAA (changes requested):
@awilnoaa @nathanvaughan-NOAA @Andrea-Havron-NOAA thank you for these changes. I am going to move this PR to draft status and update the issue to better reflect what we need. In reviewing this PR I realized that we have a few fundamental changes that we need to make to the codebase beyond just log_M. As we make everything time-varying and potentially age-varying (i.e., random walk at age for selectivity) we need to be consistent in how we do this. I think get_force_scalar() needs some work and we need to be consistent in how all values and their dimensions are reported in the json.
A fundamental question that I have is, do we always want to report things at their full scale, i.e., report every parameter for every time step, but just have a new label for estimation_type when it is essentially mirrored rather than freely estimated. For example, if we have a single M, what would be the estimation type if we report that parameter for every combination of age and year but it is only estimated once. Comparison between models where you estimate it on the annual time step versus a single value would be easier for users if the values are always reported the same. This was a design choice that we talked about a long time ago but it is not being implemented that way right now.
Second, what do we do for transformed parameters when we want their uncertainty on the untransformed scale. This is something @Andrea-Havron-NOAA is thinking a lot about right now.
Third, the dimensions of all inputs need to be checked, not just log_M, and in a systematic way.
So, I am going to make sure that the original issue reflects these needs. I am going to change this to a draft PR so we can pick it up again later, and I look forward to the future conversations that we will have trying to solve all of these problems.

PR #1336, opened 2026-03-21
Optimizing the build memory and to surface compilation performance metrics

### LINKED ISSUE #1113: [Developer Issue]: Create health metrics about compiling FIMS
### Description

We need to start capturing metrics of the build health, which was mentioned in #981. Things like compilation time, memory use, template bloat, etc. Below are some ideas for metrics to track and options for computing them. Please propose additional metrics if you wish or ways to track, store, and compare this information. 

Here are the specific tasks that should be accomplished before closing this issue
- [ ] Define which metrics should be tracked (e.g., compile time, ram usage, etc., look to other projects for ideas)
- [ ] Determine how to store the information (e.g., artifacts, logs, md files, pull request comments)
- [ ] Create a method to track the information over time, e.g., commit to commit
- [ ] Compute the metrics using a GitHub action in .github/workflows/build-metrics.yml
- [ ] Write code to run this on PRs to main

## Proposed metrics

- Compilation time (total and per file)
- Peak memory usage during build
- Binary size (library, object files)
- Number of template instantiations (template bloat)
- Number of warnings/errors

## Options for actions

- use CMake’s built-in timing (e.g., time cmake --build .)
- wrap build commands with /usr/bin/time -v or PowerShell’s Measure-Command.
- Use du, ls -lh, or PowerShell’s Get-Item for file sizes after build.
- Record .dll/.so, .a/.lib, and .o/.obj sizes.
- Enable verbose template instantiation reporting (e.g., -ftemplate-backtrace-limit, -ftime-report for GCC/Clang).
- Parse compiler output for template instantiations.
- Optionally, use tools like Bloaty McBloatface for binary analysis.
- Capture and parse build logs for warnings/errors.
- Fail CI on new warnings or increased count.

Rohith-1717:
This closes #1113 

I've added .github/workflows/build-metrics.yml which runs on PRs to dev and captures build health metrics. The workflow models the setup steps from the existing run-googletest.yml to make sure the environment is consistent. It wraps cmake --build with /usr/bin/time -v to capture compilation time and peak RAM, passes --coverage -ftime-report -ftemplate-backtrace-limit=0 to capture template instantiation time. It uses du and ls for binary sizes, and it also greps the build log for warning and error counts.
I've also added R/build_metrics.R with two functions. parse_build_metrics() basically reads the build log and it writes all the metrics to metrics.json. compare_build_metrics() compares the current run against the previous one using GitHub Actions cache and it logs any regressions to metrics_comparison.log. 
Both of the files are uploaded as artifacts on every run.

So the wall clock time was 35.67 seconds, peak RAM was 358.11 MB, template instantiation time came out at 225.25 seconds, the build had 0 warnings and 0 errors.

___

## Checklist


Rohith-1717:
Thank you for testing this :)

Yep the /usr/bin/time -v is linux specific, so when running it locally on a Windows the -v flag needs to be removed from the command. I have made a comment in the code to make this clear.

And getting the same metrics in windows is kinda tricky as there is no direct equivalent to `/usr/bin/time -v` for detailed resource usage like peak memory. We can make use of something like powershell’s measure command for timing, but even that won’t give the same level of detail. So right now, it provides full metrics in CI.

Right now the metrics are measuring the cmake --build step, so they shpw the actual compilation and not just setting up the build folder. That considers everything thats being built like FIMS, GoogleTest, google benchmark, and test binaries, not only FIMS. FIMS only would give a good view of changes to the core code, while full build metrics are useful for overall CI performance. I can shift it depending on what you prefer.

And also right now it doesn’t fail the action when the it determines that FIMS takes longer to compile. It compares the current metrics with the previous run and logs any increases in metrics_comparison.log, so the information is there in the artifacts and logs for the PR.

Rohith-1717:
Right now the metrics are measuring only the cmake --build step, so they correspond to the compilation time and doesnt include test execution. So the timing should not be affected by how long the tests themselves take but only by what gets compiled.

For regression handling, currently it logs any increases in metrics_comparison.log, but it doesn’t fail the action. 
I'll update the workflow to fail when regressions are detected so it’s more visible.
And yep, once a PR is merged, the cached metrics get updated  so the future runs will compare against the new baseline instead of the old one.

Yep also ill update the WORDLIST with the entries that u mentioned.

msupernaw:
@Rohith-1717 This is great! Probably the most important metric is the RAM. This will come in handy as we work on translational units. Thanks!

Rohith-1717:
Thanks :D  

Yep RAM is pretty interesting here. It's pretty useful for the translation units work.

Maybe I could add some hardware assesment work asw?

Rohith-1717:
Hey @kellijohnson-NOAA 

Ive replaced R/build_metrics.R with shell steps directly in the yml with awk and python3, and process tree sampling, and also the Codecov coverage issue should also be fixed too since there is no R code left to cover. I've also tried resorting to artifact based instead of cache based on my local. And I've also set the tolerance at 15 percent for wall clock time and RAM so tiny CI fluctuations don't trigger false failures :)

kellijohnson-NOAA:
@Rohith-1717 thank you for this PR. I just wanted to let you know that we have a three-day external review of our software this week so I will be slow to look at PRs until that review is over.

.github/workflows/build-metrics.yml
```diff
@@ -0,0 +1,257 @@
...
+          path: previous_metrics.json
+          key: build-metrics-previous
+
+      - name: Configure
+        run: cmake -S . -B build -DBUILD_DOC=OFF -G Ninja -DCMAKE_CXX_FLAGS="--coverage -ftime-report -ftemplate-backtrace-limit=0"
+
+      # /usr/bin/time -v is Linux only
+      # on macOS use /usr/bin/time -l, on Windows remove the prefix entirely
+      - name: Build and capture metrics
+        run: /usr/bin/time -v cmake --build build --parallel 16 2>&1 | tee build.log
```
kellijohnson-NOAA: This line will lead to moving on to the next step even if the build fails because failures are only returned for the last step, which is `tee` here. A suggestion from copilot is to run the step with `set -euo pipefail`, so the pipeline fails if any command in it fails.

.github/workflows/build-metrics.yml
```diff
@@ -0,0 +1,257 @@
...
+              if [ "$CURR_WARN" -gt "$PREV_WARN" ] 2>/dev/null; then
+                REGRESSIONS="${REGRESSIONS}WARNING: warning count increased from $PREV_WARN to $CURR_WARN\n"
+              fi
+
+              if [ "$CURR_ERR" -gt "$PREV_ERR" ] 2>/dev/null; then
+                REGRESSIONS="${REGRESSIONS}WARNING: error count increased from $PREV_ERR to $CURR_ERR\n"
+              fi
+
+              if [ "$PREV_WALL" != "null" ] && [ "$CURR_WALL" != "null" ]; then
+                WALL_REGRESSED=$(awk "BEGIN {print (($CURR_WALL - $PREV_WALL) / $PREV_WALL * 100) > 15 ? 1 : 0}")
```
kellijohnson-NOAA: This call to `awk` and the similar call on line 206 was identified as fragile by copilot because it is easy for `awk` to parse incorrectly because print expr > 15 can be interpreted as a redirection form instead of a boolean comparison.

A minimal shell-safe version would be:
```
WALL_REGRESSED=$(awk "BEGIN { pct=(($CURR_WALL - $PREV_WALL) / $PREV_WALL) * 100; print (pct > 15) ? 1 : 0 }")
```
That extra pct= assignment makes the parse much clearer.

You should also guard against divide-by-zero if an old baseline somehow contains 0:
```
if [ "$PREV_WALL" != "null" ] && [ "$CURR_WALL" != "null" ] && [ "$PREV_WALL" != "0" ]; then
```
Same comment applies to RAM and template comparisons.

.github/workflows/build-metrics.yml
```diff
@@ -0,0 +1,257 @@
...
+
+      - name: Setup for gtest
+        run: FIMS:::setup_gtest()
+        shell: Rscript {0}
+
+      - name: Restore previous metrics from cache
+        uses: actions/cache/restore@v4
+        with:
+          path: previous_metrics.json
+          key: build-metrics-previous
```
kellijohnson-NOAA: Should we be using cache or should we be pulling artifacts from previous builds, like the last build to main?

kellijohnson-NOAA (changes requested):
@Rohith-1717 this is great. I have a few comments regarding some code that was identified as fragile by copilot and one more substantial comment regarding the cache storing. I think it would be best to pull information from a previous artifact file rather than using a cache. What do you think?

kellijohnson-NOAA:
I should also add that I rebased your PR while reviewing it so I could see what files were actually changed. Please let me know if you need help checking out this version of the branch or rebasing locally. I am happy to help.

msupernaw:
@Rohith-1717 Great job putting this together. it's very important  I spent some time digging into the build metrics and compiler memory behavior related to the translation unit changes, and I wanted to summarize a few observations.

One important point is that peak memory usage can vary substantially depending on the compiler and platform being used. The local measurements I collected on macOS/Clang showed a larger reduction in peak resident memory than what we are currently observing in the Ubuntu/GCC GitHub Action environment.

The GitHub Actions workflow is currently building with g++ on Ubuntu, and GCC appears to manage template-heavy Eigen/TMB compilation differently than Clang. Because of that, I would not necessarily expect identical memory reductions across environments, even when the underlying code changes are the same.

To make the measurements more meaningful long term, I updated the workflow to explicitly track:

Wall-clock build/install time
Maximum resident set size (Max RSS)
Warning/error counts
Template compilation timing
Binary/build directory sizes

using /usr/bin/time -v around both the R CMD INSTALL path and the CMake/Ninja build path.

The main metric I think we should focus on for memory scalability is maximum resident set size (Max RSS), since that reflects actual peak resident RAM pressure during compilation and is the metric most closely tied to CI memory exhaustion risk.

Even in the Ubuntu/GCC environment, the translation unit restructuring still reduced peak RAM usage by roughly 25%. I do not think that is a paltry improvement. For large template-heavy scientific codebases, a 25% reduction in peak compiler residency is meaningful from a build scalability and CI stability standpoint, particularly when working near runner memory limits.

The tradeoff appears to be consistent with what we would expect architecturally:

lower peak memory residency,
somewhat higher aggregate CPU work,
and slightly longer build times due to increased translation-unit decomposition and repeated parsing.

Overall, I think the updated metrics workflow will help us track these tradeoffs more consistently over time and make future compiler/build-system decisions more data driven.

Attached is the version I settled on. If you think of any improvements, please feel free to make modifications.

[build-metrics.yml](https://github.com/user-attachments/files/27645822/build-metrics.yml)

kellijohnson-NOAA:
@msupernaw I simplified the workflow to install the minimal dependency set and then time just the install of FIMS. I also removed the creation of the build folder entirely. Below is the summary that I am getting between "main" and "translational units"
image
So, as you said the install time is longer and Max RSS from time -v decreased, but shared size increased 45%. Are these the kind of metrics you think we should keep track of? Do you have anything that we should add?

Rohith-1717:
Hey @kellijohnson-NOAA , I was a bit busy cuz of my final exams for the last 2 weekss

The 45 percent shared size increase is kinda interesting, it could be worth tracking that separately and maybe flagging it in the regression check since a big leap there might indicate something unexpected wouldve crept in.

kellijohnson-NOAA:
No worries at all @Rohith-1717.
@msupernaw and I have just been experimenting with what you wrote and looking a few other metrics (see my fork at kellijohnson-noaa/fims) to compare main to dev-TU-RTMB-Style, a branch that @msupernaw has been working on to try and reduce the amount of memory that is used when compiling the package. One major limitation we have right now is that the package (main branch) will not compile on a small computer so when we do things in the cloud we have to pay for larger machines to compile the code. We are trying to decrease the size and memory use to allow compiling on smaller machines. So, your work has been really great to get formal metrics of that.

msupernaw:
> Hey @kellijohnson-NOAA , I was a bit busy cuz of my final exams for the last 2 weekss
> 
> The 45 percent shared size increase is kinda interesting, it could be worth tracking that separately and maybe flagging it in the regression check since a big leap there might indicate something unexpected crept in.

@Rohith-1717 The increase in the shared object size is from duplicate template instantiations in the translational units. This is expected behavior. Still, I agree this should be tracked also.

kellijohnson-NOAA:
Update: We have three more metrics that @msupernaw will be adding regarding how small of a computer could be used in theory to build FIMS given the resources that were used while building it on ubuntu in the cloud. Once those are added tomorrow we will merge in this PR. @Rohith-1717 we are planning on attributing the PR and commit to you even though we have done some modifications. We did not design the original issue with enough guidance to get to where we ended up and that is not your fault so we want to make sure you get credit for starting this and giving us so many great ideas.

msupernaw:
> Update: We have three more metrics that @msupernaw will be adding regarding how small of a computer could be used in theory to build FIMS given the resources that were used while building it on ubuntu in the cloud. Once those are added tomorrow we will merge in this PR. @Rohith-1717 we are planning on attributing the PR and commit to you even though we have done some modifications. We did not design the original issue with enough guidance to get to where we ended up and that is not your fault so we want to make sure you get credit for starting this and giving us so many great ideas.

@Rohith-1717 Attached is the updated yaml file with "Estimated Minimum System Requirements" added. Please let me know if you experience any issues. 

[build-metrics.yml](https://github.com/user-attachments/files/27733346/build-metrics.yml)

Rohith-1717:
@msupernaw 
The additions are neat. The artifact based baseline is defo cleaner than the cache approach.

kellijohnson-NOAA:
@Rohith-1717 thank you for offering to continue to help on this PR but I think we got it. We fully realize that this turned into a much bigger project than a "good first issue" as it was originally labeled. So, we appreciate all of your work but did not want to put too much on you regarding our repeated idea creep. Feel free to assign yourself to other issues that interest you though or even propose your own issues to work on. We really enjoyed your ability to think outside of the box and propose things that we hadn't even thought about. Great work!

Here is what the current action looks like when comparing main to dev-TU-RTMB-Style. Note that the computer size did not go down as much for my version of dev-TU-RTMB-Style as what @msupernaw showed. 

image

Note that the action is not currently running for this PR because the PR is to dev and not main, I changed the action to run only on PRs to main because we are going to remove the dev branch shortly.

Issue #1291, opened 2026-03-11
[Developer Issue]: Referencing derived quantities from R to setup likelihood distributions


nathanvaughan-NOAA:
### Description

As part of the ongoing likelihood distribution refactor we would like to be able to set pointers for observed or expected values in a density component to derived quantities within FIMS. Currently only parameters exposed through rcpp can be set as pointers. This has caused confusion when derived quantities have been setup as ParameterVectors in rcpp to make them accessible despite not being estimable parameters. 

https://github.com/NOAA-FIMS/FIMS/blob/0df8c46fed5aef325236ebb626ad837dc1fa7824/inst/include/interface/rcpp/rcpp_objects/rcpp_population.hpp#L132-L135

Derived quantities currently have a string linked to their map that should be able to be used to reference them without an ID which doesn't get set until model initialization. However, these quantities are currently housed within individual modules (Fleet, Population, Catch-at-age, etc) which makes it difficult to reference them from R. 

https://github.com/NOAA-FIMS/FIMS/blob/0df8c46fed5aef325236ebb626ad837dc1fa7824/inst/include/models/functors/catch_at_age.hpp#L68-L74

To simplify this process I propose developing a derived quantities module that stores all derived quantities, and setup individual models so that they add quantities as needed. This will hopefully simplify the location for referencing as well as reporting out of the values. 

This may also overlap with the idea of unifying derived quantities and parameters which will be described in another issue.

nathanvaughan-NOAA:
Our recent likelihood refactor as changed ParameterVector to VariableVector and now includes estimated parameters and derived quantities so that all of them have id's and are accessible from R this avoided needing to use string names and the associated overhead. Thanks for the work on this though @codeafridi, we may end up adapting some of this in the future (and will credit you if we do) if we discover we need names vs integer id's for other reasons.

PR #1357, opened 2026-03-25
Add population derived quantity lookup by name in rcpp

codeafridi:
proof of concept for issue #1291 
# What is the feature?
This PR adds a small proof of concept for referencing a population derived quantity from R by name. in issue #1291, the main problem is that derived quantities are created during model setup but do not have a consistent R facing access path like parameters do so I focused on a minimal population side example.

I added a small registry path in the CatchAtAge Rcpp interface for population derived quantities during model initialization.

The change does the following:
- registers population derived quantities into the existing model setup path
- stores a stable ID for each quantity by name
- exposes an Rcpp method to retrieve the ID from R using the population id and quantity name

For the proof of concept, I used "spawning_biomass".

I also added a focused test in "test-initialize_modules.R" to check that:
- a population derived quantity can be referenced from R by name after model initialization
- an unknown derived quantity returns an error

No this change is only within FIMS


kellijohnson-NOAA:
@nathanvaughan-NOAA and @Andrea-Havron-NOAA are you using parts of this PR for your refactor on random effects? If yes, I would like to ensure that @codeafridi gets credit by being a commit author. We can modify this PR if that would be helpful. Please let me know how to proceed. Also, @msupernaw are there parts of the PR that you think should be brought into the codebase?

nathanvaughan-NOAA:
Not sure at the moment @Kelli Johnson - NOAA Federal
***@***.***> but we will make sure to attribute it correctly if
we do.

*Nathan R. Vaughan Ph.D.*

Research Scientist, Contractor/Affiliate with Vaughan Analytics in support
of
NOAA Fisheries - Southeast Fisheries Science Center |  U.S. Department of
Commerce

Mobile: (305) 733-6643

www.fisheries.noaa.gov

kellijohnson-NOAA:
Thank you @codeafridi for this Proof of Concept PR. I am going to close it for now because it really is not planned for it to move into main. If we end up using the concepts that were outlined here we will be sure to give you attribution for your ideas by making you a coauthor of the commit. Thank you for your contributions.

Issue #1370, opened 2026-03-26
[Developer Issue]: Enhance convergence checks


Andrea-Havron-NOAA:
### Description

I propose two improvements to the current convergence tests in `fit_fims`:
 
1. Currently there are no checks catching if `nlminb` fails with `NA`. When this situation happens, there is no output from `nlminb` to pass on to the user. Ideally, `nlminb` should be wrapped in `tryCatch` and if this returns an error, `fit_fims` should stop and throw a warning along with a return of the TMB object to help users with diagnostics. The tricky part is that nlminb is called multiple time in a loop to improve the gradient. Should each call be wrapped in `tryCatch` or the entire code segment from the first call to the end of the loop?
2. When `fit_fims` fails a convergence test, it prints the error to the screen as a warning but still returns the `FIMSFit` output. This is done so that users can access the output at the state when the model failed in order to diagnose the issue. The returned `FIMSFit` object, however, does not contain the error messages. I propose we add two new field to `FIMSFit`: **convergence** and **message** where **convergence** is 0 if the model passes all convergence checks (following `nlminb` convention) and 1 if it does not. Any error/warnings messges can be appended to the **message** field with the messge: _model passed all convergence tests_ if all tests pass.

awilnoaa:
I’ve been running into an issue with 2-loop vs 3-loop convergence behavior and wanted to suggest a related feature to this fix.

Right now, if a later nlminb() restart returns a worse convergence message, that can end up determining the final fit status even when an earlier restart had already converged cleanly. In the case I saw, the 2-loop and 3-loop fits had essentially the same parameter estimates and objective value, but the later restart reported false convergence (8) and a slightly larger gradient.

One possible improvement would be to store optimizer diagnostics across the restart loops, including at least:
- convergence code
- message
- objective
- max gradient

Then, if a later restart fails or reports worse convergence, fit_fims() could check whether an earlier restart completed successfully and, if so, fall back to the last successful fit instead of treating the final restart as the only result.

This would also make restart-loop behavior much easier to diagnose, since users could inspect the full restart history rather than only the final nlminb() result.

Not sure if that fits within the scope of this issue, but it seemed closely related to the convergence/message handling being discussed here.

aman-raj-srivastva:
@awilnoaa Thanks for the detailed suggestion!
[...]

PR #1385, opened 2026-03-30
Improve convergence handling in fit_fims by wrapping nlminb calls with tryCatch

aman-raj-srivastva:
This PR improves convergence handling in fit_fims by wrapping nlminb calls in tryCatch blocks.

Changes include:
- Wrapped initial nlminb call with tryCatch
- Wrapped nlminb calls inside optimizer restart loop
- Added graceful handling of optimization failures
- Ensured valid FIMSFit object is returned even when optimization fails
- Added informative warnings for debugging

This addresses issue #1370 and improves robustness and user feedback during optimization.

___

## Checklist

### DIFF
--- a/R/fimsfit.R
+++ b/R/fimsfit.R
@@ -601,7 +601,8 @@ fit_fims <- function(input,
       opt = list(
         par = obj[["par"]],
         objective = NA_real_,
-        convergence = 1L
+        convergence = 1L,
+        message = "Optimization failed"
       ),
       sdreport = list(),
       timing = timing
--- a/R/fimsfit.R
+++ b/R/fimsfit.R
@@ -566,15 +566,49 @@ fit_fims <- function(input,
   ## optimize and compare
   cli::cli_inform(c("v" = "Starting optimization ..."))
   t0 <- Sys.time()
-  opt <- with(
-    obj,
-    nlminb(
-      start = par,
-      objective = fn,
-      gradient = gr,
-      control = control
-    )
+  opt <- tryCatch(
+    {
+      with(
+        obj,
+        nlminb(
+          start = par,
+          objective = fn,
+          gradient = gr,
+          control = control
+        )
+      )
+    },
+    error = function(e) {
+      cli::cli_warn(c(
+        "!" = paste("nlminb failed:", e$message)
+      ))
+      return(NULL)
+    }
   )
+
+  if (is.null(opt)) {
+    cli::cli_warn("Optimization failed. Returning partial results.")
+    
+    timing <- c(
+      time_optimization = Sys.time() - t0,
+      time_sdreport = as.difftime(0, units = "secs"),
+      time_total = Sys.time() - t0
+    )
+  
+    fit <- FIMSFit(
+      input = input,
+      obj = obj,
+      opt = list(
+        par = obj[["par"]],
+        objective = NA_real_,
+        convergence = 1
+      ),
+      sdreport = list(),
+      timing = timing
+    )
+    
+    return(fit)
+  }
   maxgrad0 <- maxgrad <- max(abs(obj$gr(opt[["par"]])))
   if (number_of_loops > 0) {
     cli::cli_inform(c(
@@ -585,15 +619,30 @@ fit_fims <- function(input,
       # differences in values printed out using control$trace will be
       # negligible between these different runs and is not worth printing
       control$trace <- 0
-      opt <- with(
-        obj,
-        nlminb(
-          start = opt[["par"]],
-          objective = fn,
-          gradient = gr,
-          control = control
-        )
+      opt <- tryCatch(
+        {
+          with(
+            obj,
+            nlminb(
+              start = opt[["par"]],
+              objective = fn,
+              gradient = gr,
+              control = control
+            )
+          )
+        },
+        error = function(e) {
+          cli::cli_warn(c(
+            "!" = paste("nlminb failed during loop:", e$message)
+          ))
+          return(NULL)
+        }
       )
+
+      if (is.null(opt)) {
+        break
+      }
+      
       maxgrad <- max(abs(obj[["gr"]](opt[["par"]])))
     }
     div_digit <- cli::cli_div(theme = list(.val = list(digits = 5)))
--- a/R/fimsfit.R
+++ b/R/fimsfit.R
@@ -566,123 +566,49 @@ fit_fims <- function(input,
   ## optimize and compare
   cli::cli_inform(c("v" = "Starting optimization ..."))
   t0 <- Sys.time()
-  opt <- tryCatch(
-    {
-      with(
-        obj,
-        nlminb(
-          start = par,
-          objective = fn,
-          gradient = gr,
-          control = control
-        )
-      )
-    },
-    error = function(e) {
-      cli::cli_rule(left = "nlminb failure detail")
-      cli::cli_warn(c(
-        "x" = "nlminb failed: {e$message}",
-        "i" = "Returning partial results."
-      ))
-      return(NULL)
-    }
-  )
-
-  # Handle soft failures where nlminb returns an object but the objective
-  # value is not finite (e.g., Inf or NaN). In such cases, convergence and
-  # message fields may be unreliable, so treat this as a failure and trigger
-  # the fallback logic by setting opt to NULL.
-  if (!is.null(opt) && !is.finite(opt$objective)) {
-    cli::cli_rule(left = "nlminb failure detail")
-    cli::cli_warn(c(
-      "x" = "Optimization returned a non-finite objective value.",
-      "i" = "Returning partial results."
-    ))
-    opt <- NULL
-  }
+  opt <- try_nlminb(object = obj, control_list = control)
 
   if (is.null(opt)) {
-
-    # Record timing information
-    timing <- c(
-      time_optimization = Sys.time() - t0,
-      time_sdreport = as.difftime(0, units = "secs"),
-      time_total = Sys.time() - t0
-    )
-
-    # Create a fallback optimizer result to maintain consistent structure
-    # convergence = 1L indicates non-convergence (integer as expected by nlminb)
-    opt <- list(
-      par = obj[["par"]],
-      objective = NA_real_,
-      convergence = 1L,
-      message = "Optimization failed"
-    )
-
-    # Construct a valid FIMSFit object using the standard constructor
-    # This ensures all required slots are properly populated
-    fit <- FIMSFit(
+    failed_nlminb_object <- return_failed_nlminb(obj, t0)
+    failed_fit <-   fit <- FIMSFit(
       input = input,
       obj = obj,
-      opt = opt,
+      opt = failed_nlminb_object[["opt"]],
       sdreport = list(),
-      timing = timing
+      timing = failed_nlminb_object[["timing"]]
     )
-
-    return(fit)
+    return(failed_fit)
   }
 
   maxgrad0 <- maxgrad <- max(abs(obj$gr(opt[["par"]])))
   if (number_of_loops > 0) {
     cli::cli_inform(c(
-      "i" = "Restarting optimizer {number_of_loops} times to improve gradient."
+      "i" = "Restarting optimizer {number_of_loops} times{?s} to improve
+            gradient."
     ))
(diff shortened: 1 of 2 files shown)


kellijohnson-NOAA (changes requested):
Thanks for the speedy PR. I have some inline comments and I am also wondering about the location of "convergence", where I am assuming this is a typical item in opt so we just recreate that but with a non-converged value of 1.0? Also, should it be `1L` to force it to be an integer?

kellijohnson-NOAA:
Sorry for the delay in reviewing this. I will review it this week.

R/fimsfit.R
```diff
@@ -585,15 +619,30 @@ fit_fims <- function(input,
...
+        error = function(e) {
+          cli::cli_warn(c(
+            "!" = paste("nlminb failed during loop:", e$message)
+          ))
+          return(NULL)
+        }
       )
+
+      if (is.null(opt)) {
+        break
```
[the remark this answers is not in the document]
aman-raj-srivastva: Following up on this, I have updated the loop logic to retain the last successful optimization result (`prev_opt`) and use it as a fallback when a failure occurs, instead of returning NULL.

Specifically:

```r
if (is.null(opt)) {
  opt <- prev_opt
  break
}
```

This ensures the function exits the loop with a valid result and avoids ambiguity in the returned object.

kellijohnson-NOAA:
@aman-raj-srivastva thank you for this PR. I just wanted to let you know that we have a three-day external review of our software this week so I will be slow to look at PRs until that review is over.

aman-raj-srivastva:
Hi @kellijohnson-NOAA, just a gentle follow-up on this PR. I understand you were busy with the external review last week, hope that went well!

I’ve addressed the earlier comments and am happy to make any further changes if needed. Looking forward to your feedback whenever you get a chance. Thanks!

R/fimsfit.R
```diff
@@ -566,15 +566,50 @@ fit_fims <- function(input,
...
+      input = input,
+      obj = obj,
+      opt = list(
+        par = obj[["par"]],
+        objective = NA_real_,
+        convergence = 1L,
+        message = "Optimization failed"
+      ),
+      sdreport = list(),
+      timing = timing
```
kellijohnson-NOAA: I actually do not think this will work without returning all of the slots, e.g., for a converged model we return
```
  fit <- methods::new(
    "FIMSFit",
    input = input,
    obj = obj,
    opt = opt,
    max_gradient = max_gradient,
    report = report,
    sdreport = sdreport,
    number_of_parameters = number_of_parameters,
    timing = timing,
    version = version,
    model_output = model_output
  )
```

R/fimsfit.R
```diff
@@ -566,15 +566,50 @@ fit_fims <- function(input,
...
+          start = par,
+          objective = fn,
+          gradient = gr,
+          control = control
+        )
+      )
+    },
+    error = function(e) {
+      cli::cli_warn(c(
+        "!" = "nlminb failed: {e$message}",
```
kellijohnson-NOAA: I think this should be an "x" instead of "!", where the latter means a warning rather than the former, which means an error.
```suggestion
        "x" = "nlminb failed: {e$message}",
```

R/fimsfit.R
```diff
@@ -585,15 +620,36 @@ fit_fims <- function(input,
...
+              start = opt[["par"]],
+              objective = fn,
+              gradient = gr,
+              control = control
+            )
+          )
+        },
+        error = function(e) {
+          cli::cli_warn(c(
+            "!" = "nlminb failed during loop: {e$message}",
```
kellijohnson-NOAA:
```suggestion
            "!" = "{.fn nlminb} failed during loop {ii}: {e$message}",
```

kellijohnson-NOAA (changes requested):
@aman-raj-srivastva thank you for this PR and sorry it took me so long to review it. I suggested a few minor changes within the use of {cli} and one bigger change regarding the need to return all entries within the FIMSFit class, it needs to have all of the objects to successfully run. With that, we should test that we get the appropriate warnings when a malformed model is passed to nlmimb. Please let me know if you need help creating the tests.

R/fimsfit.R
aman-raj-srivastva: Thanks for the suggestion! I updated the fallback case to return a complete `FIMSFit` object using the existing constructor so that all required slots are populated consistently. The constructor also automatically handles slots like `version` and `model_output` which helps ensure consistency and avoids missing fields compared to a manual `methods::new()` call.

Please let me know if you would prefer explicitly constructing it with `methods::new()` to avoid any implicit behavior.

kellijohnson-NOAA:
@Andrea-Havron-NOAA do you have a suggestion for some code we could use to test nlmimb failing, i.e., how do we quickly set up code for a model that we know should fail to pass nlmimb but will get past initial FIMS checks?

Andrea-Havron-NOAA:
> @Andrea-Havron-NOAA do you have a suggestion for some code we could use to test nlmimb failing, i.e., how do we quickly set up code for a model that we know should fail to pass nlmimb but will get past initial FIMS checks?

My initial thought was to change agecomp data so all the age comp values for a single year are equal to 0. Then I realized we need to add an error check in FIMS to catch this error for age/length comps ([here](https://github.com/NOAA-FIMS/FIMS/blob/f1020161b95bc74c2d7dc470125c81a6c55aff3c/inst/include/models/functors/catch_at_age.hpp#L931) and [here](https://github.com/NOAA-FIMS/FIMS/blob/f1020161b95bc74c2d7dc470125c81a6c55aff3c/inst/include/models/functors/catch_at_age.hpp#L996)) 

My second thought was to set a parameter initial value to -Inf. You can set up a default model and set the first `log_init_naa = -Inf`. The model fails pretty quickly.

kellijohnson-NOAA:
Okay, I can try this. I tried
```
obj$fn <- function(x) return(NaN)
```
and it led to the following warnings being printed to the screen
```
Warning messages:
1: In nlminb(start = par, objective = fn, gradient = gr, control = control) :
  NA/NaN function evaluation
2: In nlminb(start = par, objective = fn, gradient = gr, control = control) :
  NA/NaN function evaluation
3: In nlminb(start = par, objective = fn, gradient = gr, control = control) :
  NA/NaN function evaluation
```
but I still got an object back from `nlmimb()` instead of NULL so it did not get passed to the `error` part of `tryCatch()`, which I guess makes sense because it is a warning. 
Should we also be looking to see if parts of `opt` are "positive" to see if we should move on with the optimization? Below are parts of what is in the nlmimb() object that is returned when the NA warnings are given.
```
$objective
[1] Inf
$convergence
[1] 0
$iterations
[1] 1
$evaluations
function gradient 
       2        1 
$message
[1] "relative convergence (4)"
```
Should we move onto another round of optimization here or stop?

Andrea-Havron-NOAA:
> Okay, I can try this. I tried
> 
> ```
> obj$fn <- function(x) return(NaN)
> ```
> 
> and it led to the following warnings being printed to the screen
> 
> ```
> Warning messages:
> 1: In nlminb(start = par, objective = fn, gradient = gr, control = control) :
>   NA/NaN function evaluation
> 2: In nlminb(start = par, objective = fn, gradient = gr, control = control) :
>   NA/NaN function evaluation
> 3: In nlminb(start = par, objective = fn, gradient = gr, control = control) :
>   NA/NaN function evaluation
> ```
> 
> but I still got an object back from `nlmimb()` instead of NULL so it did not get passed to the `error` part of `tryCatch()`, which I guess makes sense because it is a warning. Should we also be looking to see if parts of `opt` are "positive" to see if we should move on with the optimization? Below are parts of what is in the nlmimb() object that is returned when the NA warnings are given.
> 
> ```
> $objective
> [1] Inf
> $convergence
> [1] 0
> $iterations
> [1] 1
> $evaluations
> function gradient 
>        2        1 
> $message
> [1] "relative convergence (4)"
> ```
> 
> Should we move onto another round of optimization here or stop?

If the objective value (`$objective`) is not a real number, optimization should stop. I suspect since nlminb errored out, the other fields like `convergence` or `message` never updated. I don't think these fields are reliable. TMB can and does recover from testing parameter values that result in invalid log-likelihoods. If nlminb fails, it's because the model has a serious problem.

Andrea-Havron-NOAA:
If you need to run a FIMS model to generate the error, an even simpler approach would be to set up the default configurations and parameters and then run:
```
init_parms = parameters_4_model |>
  initialize_fims(data = data_4_model)
init_parms$parameters$p <- rep(-Inf,length(init_parm$parameters$p)
fit_fims(input = init_parms, optimize = TRUE)
```

kellijohnson-NOAA:
Thanks @Andrea-Havron-NOAA for the suggestion, it worked and I got
```
Warning messages:
1: In nlminb(start = par, objective = fn, gradient = gr, control = control) :
  NA/NaN function evaluation
2: x nlminb failed: NA/NaN gradient evaluation
i Returning partial results. 
```
So, we just need to figure out how to get a return after the warning message before the message from FIMS is printed because right now they are mixed together.

aman-raj-srivastva:
I’ve been following the discussion here, thanks for working through these cases.

Based on the points raised, I suggest the following updates in the PR:

1. Treat non-finite objective values as failure:

```r
if (!is.null(opt) && !is.finite(opt$objective)) {
  opt <- NULL
}
```

2. Suppress `nlminb()` warnings so they don’t mix with the `{cli}` messages:

```r
opt <- tryCatch(
  {
    suppressWarnings(
      with(
        obj,
        nlminb(
          start = par,
          objective = fn,
          gradient = gr,
          control = control
        )
      )
    )
  },
  error = function(e) {
    cli::cli_warn(c(
      "x" = "nlminb failed: {e$message}",
      "i" = "Returning partial results."
    ))
    return(NULL)
  }
)
```

This way we:

* handle both hard failures and soft failures (`Inf` / `NaN` objective), and
* keep the output clean by only showing the `{cli}` messages.

Happy to implement these changes if this approach looks good.

kellijohnson-NOAA:
I would like to keep the error message from nlmimb but because it is FORTRAN it is hard to work with. Gemini came up with the following
```
    error = function(e) {
        # 1. Force a visual break from any previous stderr output
        cli::cli_rule(left = "nlminb failure detail")
        
        # 2. Use semantic bullets
        cli::cli_abort(c(
          "x" = "The optimizer encountered a critical error.",
          "i" = "Details: {.emph {e$message}}",
          "h" = "Check your TMB {.code data} and {.code parameters} for {.val NaN} or {.val Inf} values."
        ), call = NULL) 
        # Note: Use cli_abort if you want to stop, or cli_warn + return(NULL) as you had.
      }
```
which leads to
image

aman-raj-srivastva:
I like the approach of keeping the original nlminb output while improving readability with `{cli}`, I’ll update the PR to add a clear separator (e.g., `cli_rule()`) and structured messaging while preserving the warning output, and continue using `cli_warn()` with `return(NULL)` so the fallback FIMSFit behavior is maintained (happy to switch to `cli_abort()` if preferred).

kellijohnson-NOAA:
@aman-raj-srivastva sorry, I should have noticed that Gemini changed it to abort. We want to keep it with the `cli_warn()` behavior as you had it. That way if the object is NULL then a FIMSFit object can be returned early, where `cli_abort()` won't let that happen.

aman-raj-srivastva:
Got it, thanks for the clarification! I’ll keep the `cli_warn()` approach so the fallback `FIMSFit` object can still be returned, and just improve the formatting around the message.

aman-raj-srivastva:
@kellijohnson-NOAA I’ve updated the handling for both hard and soft failures in the initial optimization and within the restart loop. Initially, I handled the soft failure case only inside the restart loop, but then I realized that if the very first `nlminb()` call returns a non-finite objective (e.g., `Inf`/`NaN), will still proceed into the loop with an invalid state. To avoid that I added the same check right after the initial optimization as well.
Please let me know if everything looks good.

kellijohnson-NOAA:
@aman-raj-srivastva thank you for all of this great work. I have one failing test and I am working with @Andrea-Havron-NOAA to debug it. After that test is debugged and I get a code coverage metric back, I will merge this in. So, you can consider your part done and leave the rest up to me. This was a great PR to review and we appreciate your hard work.

PR #1403, opened 2026-04-03
refactor: migrate spell checking from hunspell to cspell

### LINKED ISSUE #1361: [Developer Issue]: change spelling to cspell
### Description

The spell check workflow is not "working" great in my opinion for our codebase because we use a combination of PascalCase and snake_case. I asked Gemini what other options are available for FIMS.

> nmfs-ost/ghactions4r action relies on {spelling} which is built on hunspell, and treats PascalCase words as single unknown strings because they aren't in the standard dictionary. To accommodate Google C++ Style (PascalCase) without manual WORDLIST entries, you can use one of the following strategies:

1. The "Cleanest" Fix: Use a Regex-based Word Splitter
The spelling package allows you to ignore words matching a certain pattern. Instead of adding every word to the WORDLIST, you can modify your CI workflow (usually in .github/workflows/spelling.yml) to filter out or split PascalCase words before they reach the engine. If you are calling spelling::spell_check_package() in your own R script within the action, you can use a regex to exclude PascalCase identifiers:
```r
# In your R script that runs the spell check:
# This identifies words with a leading capital and at least one internal capital (PascalCase)
# and adds them to the 'ignore' list dynamically.
all_files <- list.files(recursive = TRUE, pattern = "\\.(R|cpp|h)$")
text <- unlist(lapply(all_files, readLines))
pascal_words <- grep("\\b[A-Z][a-z]+([A-Z][a-z]+)+\\b", unlist(strsplit(text, "[^a-zA-Z]")), value = TRUE)
spelling::spell_check_package(ignore = c(pascal_words))
```

2. Configure PySpelling (If using rojopolis/spellcheck-github-actions)
Many NOAA/NMFS repositories use a .spellcheck.yml configuration. If your workflow uses PySpelling (the engine behind many generic GitHub spellcheckers), you can enable "CamelCase" splitting. This breaks PascalCaseWord into Pascal, Case, and Word before checking them. The following would need to be added to a GitHub action: 

```
YAML
matrix:
- name: cpp-files
  sources:
  - 'src/**/*.cpp'
  - 'src/**/*.h'
  aspell:
    lang: en
  pipeline:
  - pyspelling.filters.cpp:
  - pyspelling.filters.context:
      context_visible_first: true
      # This allows the spellchecker to split CamelCase and PascalCase
      delimiters: |
        (?x)
        (?P<camel>(?<=[a-z])(?=[A-Z])) |
        (?P<pascal>(?<=[A-Z])(?=[A-Z][a-z]))
```

3. Use cspell (Recommended for R + C++ Projects)
It is designed specifically for code and handles CamelCase/PascalCase by default by splitting them into component words. This workflow is faster than {spelling} because you do not need to install an R package, which can take up to 3 minutes, you just need a simple command-line tool.
Action: streetsidesoftware/cspell-action
Benefit: You won't need an inst/WORDLIST for DataProcessor because it will check Data and Processor independently.

    1. Create the GitHub Action, .github/workflows/cspell.yml. This action will run on every pull request to catch spelling errors before they are merged.

```
name: "Spelling Check"
on:
  push:
    branches: [main]
  pull_request:
    branches: [main]

jobs:
  spelling:
    name: Check Spelling
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      
      - name: Run CSpell
        uses: streetsidesoftware/cspell-action@v6
        with:
          # Define which files to check (globs)
          files: "**/*.{cpp,h,R,md}"
          # Use incremental checking to only check changed files in PRs
          incremental_files_only: true
          # Path to your configuration file
          config: "cspell.json"
```

    2. Configure cspell.json, instead of an inst/WORDLIST, in your root directory. This is where you specify global ignores and project-specific settings.

```
{
  "version": "0.2",
  "language": "en",
  "words": [
    "FIMS",
    "NOAA",
    "cppit",
    "Rcpp"
  ],
  "ignorePaths": [
    "inst/WORDLIST",
    "**/node_modules/**",
    "*.Rproj",
    ".github/workflows/**"
  ],
  "dictionaries": [
    "cpp",
    "en_US",
    "softwareTerms",
    "r"
  ],
  "ignoreRegExpList": [
    "/[a-z0-9]{32,}/g" 
  ]
}
```

    3. Key Configuration Features: Specifying files uses standard glob patterns. `src/**/*.cpp` — only C++ source files. `**/*.{R,cpp,h}` — all R and C++ related files. Ignoring Words (words): Add acronyms or project names (like FIMS) to the "words" array in cspell.json. Ignoring Patterns (ignoreRegExpList): If you have specific strings (like hex codes, long IDs, or specialized math notation) that aren't words, you can use Regular Expressions to tell the checker to skip them entirely. Inline Ignores: If you have a one-off word in a file you don't want to add to the global list, you can add a comment directly in the code: `C++: // cspell:ignore MyStrangeVariable` and `R: # cspell:ignore MyStrangeVariable`

    4. Transitioning from inst/WORDLIST is seamless because we already have a WORDLIST, we can simply copy its contents into a text file (e.g., fims-dict.txt) and point cspell.json to it as a custom dictionary. 

4. Strategic inst/WORDLIST Management
If you must stay with the current R-based system, you can automate the maintenance of inst/WORDLIST so developers don't have to do it manually. You can add a pre-commit hook or a local utility function to your package:

```r
# A helper function for FIMS developers
update_fims_wordlist <- function() {
  bad_words <- spelling::spell_check_package()
  # Filter out words that look like valid PascalCase
  true_errors <- bad_words[!grepl("^[A-Z][a-z]+[A-Z]", bad_words$word), ]
  # Automatically add the rest to WORDLIST
  spelling::update_wordlist(confirm = FALSE)
}
```

I think we should choose Option 3 of using cspell. And we can remove all PascalCase words that are combinations of real words from our WORDLIST before converting it over to the json file.

Eshaan-byte:
# What is the feature?
Replaces the R-based {spelling} workflow with cspell (streetsidesoftware/cspell-action@v6). cspell handles PascalCase/camelCase splitting natively, so words like CalculateBiomass no longer need manual WORDLIST entries.

-Added cspell.json config pointing to fims-dict.txt as a custom dictionary
-Added .github/workflows/cspell.yml using streetsidesoftware/cspell-action@v6
-Removed .github/workflows/call-spell-check.yml
-Added !cspell.json exception to .gitignore (which ignores *.json)
-Cleaned the WORDLIST from 1263 to 878 entries, removing PascalCase compounds that cspell splits automatically while keeping acronyms, names, and domain terms

No longer depends on nmfs-ost/ghactions4r for spell checking.

Fixes #1361

___

## Checklist


e-perl-NOAA:
@Eshaan-byte thank you for working on this! I do have a several comments/suggestions:

1. Since the call-spell-check.yml is removed, I believe that we can remove the WORDLIST file completely (and thus remove it from the cspell.json file) since it's no longer calling `spelling` (@kellijohnson-NOAA feel free to fact check me on this).
2. Another thing is that I'm not sure if "inst/include/interface/rcpp/**" should be in the ignore files portion. @kellijohnson-NOAA, should that still be spell checked?
3. I don't know if you did this already, but running cspell without the fims-dict.txt file shows that even more words can be removed given that cspell handle case differences (eg bool, BOOL, and Bool are all treated the same) and that it can handle a lot of the other words together.

@kellijohnson-NOAA I'm not sure if having the cspell.json file and the fims-dict.txt should go in the root directory or elsewhere. I could see them going in the .github folder but also I could see them going elsewhere too.

kellijohnson-NOAA:
Thanks @Eshaan-byte for the great PR and @e-perl-NOAA for the comments. I agree that we can remove WORDLIST if we are going to have fims-dict.txt but we could just point cspell.json to a modified inst/WORDLIST rather than creating a new file. Given that we are an R package, I think that it might be best to do it that way. Regarding the location of the cspell.json file, I think that we should leave it where it is in the home/base directory. That way we are operating similar to other packages in a standardized way and so VS Code can use it. We will have to add it to Rbuildignore if that was not already done.

Eshaan-byte:
Thanks @e-perl-NOAA and @kellijohnson-NOAA for the feedback! A few things before I push updates:

On reusing inst/WORDLIST — makes sense, so I'll just point cspell.json at it directly and drop the fims-dict.txt file. Will also add cspell.json to .Rbuildignore.

About inst/include/interface/rcpp/** being in the ignore list, I added that since those files are mostly auto-generated Rcpp boilerplate, but no strong opinion. @kellijohnson-NOAA any preference?

For trimming the list further, yeah I can do another pass to drop the case duplicates (bool/BOOL/Bool) and anything cspell's built-in dictionaries already cover.

Also, I checked and spelling is still listed in Suggests in DESCRIPTION (line 106), and there's a Language: en-US field (line 130) that the spelling package uses. Should I remove both as part of this PR, or would you prefer I open a separate PR to keep the cleanup isolated?

kellijohnson-NOAA:
I think we should keep the boilerplate rcpp files for right now.

We can get rid of spelling from the DESCRIPTION but not the language field from the file. You can add that to this PR.

CONTRIBUTING.md
```diff
@@ -69,7 +69,7 @@ Along the development process it is important to add tests and ensure that the c
...
 * Format C++ code using clang-format version 18.0.0 with Google style: `clang-format -i --style="{BasedOnStyle: Google, SortIncludes: false}" $(find ./inst/include ./src ./tests/gtest -name "*.hpp" -o -name "*.cpp")`. If you are contributing from a fork, the automated formatting workflow will add a comment to your pull request with detailed instructions on how to format your code locally.
 * Run clang tidy to check C++ for common mistakes using TODO: document how to run clang-tidy.
-* Spell check the package using `spelling::spell_check_files(list.files(c("R", "tests", file.path("inst", "include")), recursive = TRUE, full.names = TRUE, pattern = "\\.cpp|\\.hpp|\\.md|\\.R$|\\.Rmd|\\.txt"), ignore = spelling::get_wordlist())`. Remove all the .md files in the vignettes directory and then spell check the package using `spelling::spell_check_package(pkg = ".", use_wordlist = TRUE)`.
+* Spell check the package using [cspell](https://cspell.org/). Install with `npm install -g cspell` and run `cspell "**/*.{cpp,hpp,h,R,Rmd,md}"` from the package root. Add any project-specific terms to `inst/WORDLIST` (used as the cspell dictionary).
```
kellijohnson-NOAA, line 72: This PR is great, I had to find a way to install this on a LOCKED down Windows machine but I was able to do it by installing the zipped version. So, I think the instructions are fine.

Though, I did get errors from running cspell given the instructions here. 
```
CSpell: Files checked: 161, Issues found: 277 in 41 files.
```
The message was the same even after I passed it the cspell.json file, e.g.,
```
cspell --config ./cspell.json "**/*.{cpp,hpp,h,R,Rmd,md}"
```
I am not sure what I am doing wrong.

cspell.json
```diff
@@ -0,0 +1,27 @@
...
+      "name": "fims-wordlist",
+      "path": "./inst/WORDLIST",
+      "addWords": true
+    }
+  ],
+  "dictionaries": [
+    "fims-wordlist",
+    "cpp",
+    "r",
+    "softwareTerms"
```
kellijohnson-NOAA, line 15: What are these dictionaries?

cspell.json
```diff
@@ -0,0 +1,27 @@
...
+  ],
+  "ignorePaths": [
+    "**/node_modules/**",
+    "*.Rproj",
+    ".github/workflows/**",
+    "inst/include/interface/rcpp/**",
```
kellijohnson-NOAA:
```suggestion
```
We want to include these files even though they are boilerplate files, at least for right now.

kellijohnson-NOAA (changes requested):
I think that this PR is a great step forward for FIMS but I have a few concerns about not knowing how to run it locally. I love how it shows line by line in the PR the words that are spelled wrong or not in the dictionary and I like the output from the command line. So, I definitely think we should move forward with it but I think we need to tweak a few things. There are two words to add to the WORDLIST as indicated here and then we need to ensure that the instructions for how to run it via command line are correct.

kellijohnson-NOAA:
@Eshaan-byte do you mind replying to my comments on your PR so we can move forward with this one?

cspell.json
Eshaan-byte: Good question,these are built-in dictionaries that come bundled with cspell (https://cspell.org/docs/dictionaries/):

#cpp: covers C/C++ keywords and stdlib stuff like nullptr, size_t, constexpr, etc.
#r: covers R keywords, base functions, and common packages like dplyr, ggplot2, roxygen, etc.
#softwareTerms: general dev/CS vocabulary like API, boolean, refactor, regex, etc.

Basically they keep our WORDLIST from being cluttered with standard programming terms so it only has FIMS-specific words.

CONTRIBUTING.md
Eshaan-byte: You weren't doing anything wrong, the issue was on my end. When I switched back to inst/WORDLIST from fims-dict.txt, the original untrimmed list came back with a bunch of missing words. I've cleaned it up now and it should run clean after the next push.

cspell.json
Eshaan-byte: Done, removed the rcpp ignore. Those files will be spell-checked now.

Eshaan-byte:
> I think that this PR is a great step forward for FIMS but I have a few concerns about not knowing how to run it locally. I love how it shows line by line in the PR the words that are spelled wrong or not in the dictionary and I like the output from the command line. So, I definitely think we should move forward with it but I think we need to tweak a few things. There are two words to add to the WORDLIST as indicated here and then we need to ensure that the instructions for how to run it via command line are correct.

Thanks for the review! I've added gtests and logit to the WORDLIST and cleaned up the list so it runs clean locally. Will push the updates shortly.

Eshaan-byte:
> @Eshaan-byte do you mind replying to my comments on your PR so we can move forward with this one?

Sorry for the delay! Replied to everything and working on the fixes now.

kellijohnson-NOAA:
Thank you @Eshaan-byte for these changes. I pushed a few more changes where I was able to decrease the number of words in the WORDLIST because of duplicates in the default dictionaries that you added, removing words that were the same but just different in their case, and removing words that did not exist in the code base. Once the tests pass I will merge this into dev. THANKS AGAIN! We really appreciate this.

PR #1406, opened 2026-04-03
Fixes the inability to map parameters

### LINKED ISSUE #1404: TMB's mapping function makes `set_fixed()` and `get_estimates()` fail

I am working on the Pacific Hake case study where I want to have time-varying selectivity but I don't want it to be different every year. So, I am trying to map some of the selectivity parameters together using the map argument of `TMB::MakeADFun()`. The model ran fine but when I went to use `set_fixed()` the R session crashed. So, I reran the model and just bypassed `set_fixed()` and went on my merry way. Then, when I tried to run `get_estimates()` the function failed because the names of the parameters did not match the length of the fixed effect parameter vector.

### Additional context

Working on this in dev-par-length based on changes that were suggested from copilot.

Windows

v0.9.2

kellijohnson-NOAA:
# What is the feature?
* Allows users to map parameters to other parameters with the argument map to `TMB::MakeADFun()`
* Adds warning messages when the object passed to `set_fixed()` is not the same length as what is expected because before R would crash. 

* Uses `obj[["env"]]$parList(opt[["par"]])[["p"]]` instead of `opt[["par"]]`

* No 😁 

___

## Checklist

### DIFF
--- a/R/reshape_output.R
+++ b/R/reshape_output.R
@@ -218,6 +218,8 @@ reshape_tmb_estimates <- function(obj,
                                   sdreport = NULL,
                                   opt = NULL,
                                   parameter_names) {
+  n_fixed_parameters <- length(parameter_names)
+
   # Outline for the estimates table
   estimates_outline <- tibble::tibble(
     # The FIMS Rcpp module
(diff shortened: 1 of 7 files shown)


kellijohnson-NOAA:
According to Google the difference between `opt[["par"]]` and `obj[["env"]]$parList(opt[["par"]])[["p"]]`

>In TMB (Template Model Builder), the difference between opt[["par"]] and obj[["env"]]$parList(opt[["par"]])[["p"]] lies in whether the parameters are presented as a vector of all parameters (as used by the optimizer) or as a structured list of parameters (mapped back to the structure defined in your C++ code).Here is the breakdown of the comparison:opt[["par"]]: This is the output from an optimizer like nlminb(). It is a simple, named numeric vector containing all estimated parameters (both fixed and random, depending on how MakeADFun was called) concatenated together.obj[["env"]]$parList(opt[["par"]])[["p"]]: This uses the TMB environment function parList to convert that flat numeric vector opt[["par"]] back into a structured list of parameters. The [["p"]] at the end typically selects a specific parameter vector or matrix, for example parList() returns a list containing parameters defined in C++ as PARAMETER(p).

R/reshape_output.R
```diff
@@ -243,33 +245,53 @@ reshape_tmb_estimates <- function(obj,
...
+    if (nrow(std) < n_fixed_parameters) {
+      cli::cli_abort(
+        "The number of rows in `summary(sdreport)` ({nrow(std)}) is less than the number of fixed parameters ({n_fixed_parameters})."
+      )
+    }
+
     # Number of rows for derived quantities: based on the difference
     # between the total number of rows in std and the length of parameter_names.
-    derived_quantity_nrow <- nrow(std) - length(parameter_names)
+    derived_quantity_nrow <- nrow(std) - n_fixed_parameters
```
Andrea-Havron-NOAA, line 257: It does not look like the code here is set up to handle random effects models. 

I would use instead: ‘derived_quantity_nrow <- summary(std, “report”) |> nrow()’

Then: 
‘random_nrow <- summary(std, “random”) |>  row()’

tests/testthat/helper-integration-tests-setup-function.R
```diff
@@ -461,7 +461,8 @@ setup_and_run_FIMS_without_wrappers <- function(iter_id,
     opt <- stats::nlminb(obj[["par"]], obj[["fn"]], obj[["gr"]],
       control = list(eval.max = 10000, iter.max = 10000, trace = 0)
     )
-    FIMS::set_fixed(opt$par)
+    expanded_fixed <- obj[["env"]]$parList(opt[["par"]])[["p"]]
+    FIMS::set_fixed(expanded_fixed)
```
Andrea-Havron-NOAA, line 465: I don’t understand the purpose of set_fixed. It looks like it is only used just before sdreport, but that function defines par within the function, so par is being redefined by sdreport. I am also very uncomfortable giving users the ability to overwrite info->fixed_effects_parameters and info->random_effects_parameters. It seems like doing so could cause a lot of unintended consequences in optimization an SE calculation. TMB’s obj manages parameter updates the parameters in obj should point to info->fixed_effects_parameters and info->random_effects_parameters

Andrea-Havron-NOAA:
When map is invoked, I think we need to fix the mismatch between obj$par and info->fixed_effects_parameters and info->random_effects_parameters immediately after MakeADFun because any run with cause a mismatch in parameters [here](https://github.com/NOAA-FIMS/FIMS/blob/5666d0bcff5493a6cd185284c6e869cdb9f06e50/src/FIMS.cpp#L33) which will cause errors. Also the model will be trying to optimize parameters that are not linked to anything in the model, which will lead to convergence errors. 

I think we should pull out map IDs from obj somehow and use these to update fixed and random effects vectors in info before doing anything else. We will still need to track what has been mapped to link back up with obj$env$parList().

Andrea-Havron-NOAA:
> When map is invoked, I think we need to fix the mismatch between obj$par and info->fixed_effects_parameters and info->random_effects_parameters immediately after MakeADFun because any run with cause a mismatch in parameters [here](https://github.com/NOAA-FIMS/FIMS/blob/5666d0bcff5493a6cd185284c6e869cdb9f06e50/src/FIMS.cpp#L33) which will cause errors. Also the model will be trying to optimize parameters that are not linked to anything in the model, which will lead to convergence errors.
> 
> I think we should pull out map IDs from obj somehow and use these to update fixed and random effects vectors in info before doing anything else. We will still need to track what has been mapped to link back up with obj$env$parList().

I dug into this a bit more and wanted to correct my comment above. When using the map argument, TMB does not modify the length of `PARAMETER_VECTOR(p)` in FIMS.cpp, so we do not need to modify the length of `info->fixed_effects_parameters` and `info->random_effects_parameters`. I still think we should remove `set_fixed` and `set_random`, but other than that this PR is correct in trying to fix the output so dimensionality lines up.

PR #1494, opened 2026-05-29
Expose Population proportion_female with default 0.5

### LINKED ISSUE #521: [Feature]: prop_female should be modifiable to change sex fraction
### Describe the bug

Users can set the proportion of females in the model although this value is hard coded to 0.5 in the Population class. 

We should either remove the iterface or allow this value to be updated by the user.

prop_female in FIMS:
[interface](https://github.com/NOAA-FIMS/FIMS/blob/fe84d2f717cb0752beed8cefce23df18b9faba1a/inst/include/interface/rcpp/rcpp_objects/rcpp_population.hpp#L66)
[population class](https://github.com/NOAA-FIMS/FIMS/blob/fe84d2f717cb0752beed8cefce23df18b9faba1a/inst/include/population_dynamics/population/population.hpp#L47)

### To Reproduce

_No response_

### Screenshots

_No response_

_No response_

_No response_

_No response_

### Additional Context

_No response_

szu-yun-ko:
# What is the feature?
* Addresses issue #521 

### User / R layer
- [x] `create_default_parameters()`: added a single `Population` row (`proportion_female = 0.5`, `estimation_type = "constant"`) so users can edit it with the usual dplyr/tidyr workflow.
- [x] Exposed `proportion_female` on the `Population` Rcpp module (`src/rcpp_population.cpp`). Direct access uses the `ParameterVector` API: `pop$proportion_female[1]$value <- 0.35`.

### Rcpp
- [x] Added `proportion_female` as a `ParameterVector` on `PopulationInterface`.
- [x] Constructor sets slot `[0]` to `0.5` with `estimation_type = "constant"`, so integration tests and manual builds that skip the tibble do not inherit `Parameter()`'s default of 0.
- [x] In `add_to_fims_tmb_internal()`, if `proportion_female` has invalid size (not 1 or `n_ages`), a warning is logged and the value is reset to a scalar 0.5. Values out of `[0, 1]` also log warnings. Valid sizes are 1 (scalar, recycled across ages) or `n_ages` (per-age values). Values are copied to the TMB `Population` vector without broadcasting to `n_ages`.
- [x] `finalize()` syncs `proportion_female` `final_value_m` like other population parameters.
- [x] `population_to_json()` in `rcpp_models.hpp` includes `proportion_female` in standard model output.

### C++ layer
- [x] `proportion_female` remains a `fims::Vector<Type>` on `fims_popdy::Population`, defaulting to length 1 with value 0.5.
- [x] Spawning biomass and related calculations use `get_force_scalar(age)` so a length-1 vector is recycled across all ages, while a length-`n_ages` vector supports per-age values.
- [x] Removed the `Prepare()` loop that forced `proportion_female[age] = 0.5` on every run, which previously overwrote user-provided values.
- [x] `Initialize()` no longer resizes `proportion_female` to `n_ages`. It only ensures a default exists if the vector is empty.

### Testing this PR

End-to-end test: verify that the value reaches the TMB model, not only the R interface.
```r
library(FIMS)
library(dplyr)
data("data_big")
data_4_model <- FIMSFrame(data_big)
base_params <- data_4_model |>
  create_default_configurations() |>
  create_default_parameters(data = data_4_model) |>
  tidyr::unnest(cols = data)
low_female_params <- base_params |>
  mutate(value = if_else(label == "proportion_female", 0.35, value))
clear()
fit_default <- initialize_fims(base_params, data_4_model) |>
  fit_fims(optimize = FALSE, get_sd = FALSE)
clear()
fit_low <- initialize_fims(low_female_params, data_4_model) |>
  fit_fims(optimize = FALSE, get_sd = FALSE)

fit_default@report[["spawning_biomass"]][[1]]
fit_low@report[["spawning_biomass"]][[1]]
```
* No.

___

## Checklist


szu-yun-ko:
Hi @kellijohnson-NOAA, @nathanvaughan-NOAA, @Andrea-Havron-NOAA. I ran into some implementation questions when implementing this feature and I was hoping if you can point me in the correct direction:

The issue I hit is that `PopulationInterface` used to register a **copy** in `live_objects` (`make_shared(*this)`), while R keeps the original `Population`. That is fine for `ParameterVector` and `SharedInt` (shared storage on copy) but not for plain `double`: the tibble / `$proportion_female` updates hit the R object, while `add_to_fims_tmb_internal()` on the registered copy still saw 0.5. This PR works around that by registering `this` instead of a copy (`rcpp_population.hpp` lines 161–169). 

Question: For a scalar `proportion_female` functional requirement in #521, which approach should I use?
1. **`double` + register `this` (current draft)**: scalar R API (`pop$proportion_female <- 0.35`), but less consistent with the other declarations.
2. **`ParameterVector` length 1**: matches FIMS (`log_M`, `inflection_point`, etc.), no constructor fix needed. R access is `pop$proportion_female[1]$value`.
3. **Use `SharedReal`**: `SharedReal` today seems to use the same implementation as `SharedInt` in [`rcpp_shared_primitive.hpp`](https://github.com/NOAA-FIMS/FIMS/blob/main/inst/include/interface/rcpp/rcpp_objects/rcpp_shared_primitive.hpp#L784-L805), so it cannot represent `proportion_female`. 

I am still getting familiar with the TMB architecture and FIMS primitive data types, so please correct me if I am misunderstanding anything. I would appreciate any guidance on which of these approaches best fits the intended design, or whether there is another approach you would prefer.

Thanks!

nathanvaughan-NOAA:
Hey @szu-yun-ko looks like you are making good progress digging into the architecture. Changing proportion_female to a ParameterVector to match the other inputs seems like the best approach. I would suggest setting it up similar to other parameters that default to a scalar value using but can be extended to be time-varying. This shouldn't be any more difficult than using year instead of 1 when referencing the value because fims vectors include a get_force_scalar function to recycle scalar values for vectors. If you set a default 0.5 value that will enable the user to not specify the value if they choose and also allow all the existing examples/tests to function the same. You can look at the setup for log_f_multiplier which is filled with default values if you hit any snags, I'm not sure if we have any other parameters with internal default values.

szu-yun-ko:
@kellijohnson-NOAA @awilnoaa I think this is ready for review now. At one point I ran into an issue that seemed to be the root cause of most failing CIs, which I want to briefly document in this comment: 

On main, `Prepare()` forced `proportion_female[age] = 0.5` every evaluation, which masked missing initialization. Once we removed that loop so user values are respected, manual builds and integration tests that call `methods::new(Population)` without a parameters tibble were left with `Parameter()`’s default `initial_value_m` of `0`, which propagated into spawning biomass calculations as `NaNs`. This was fixed by setting a default in the `PopulationInterface` constructor:
```cpp
this->proportion_female[0].initial_value_m = static_cast<double>(0.5);
this->proportion_female[0].estimation_type_m.set("constant");
```
That gives integration tests and manual builds a safe fallback (0.5) without reintroducing the `Prepare()` overwrite. The standard `initialize_fims()` path still overwrites this via `set_param_vector()` when the user edits the parameters tibble. After the fix most CI passed. 

Please let me know if there's anything I'm missing to address, and thanks for the review!

inst/include/interface/rcpp/rcpp_objects/rcpp_population.hpp
```diff
@@ -413,6 +430,43 @@ class PopulationInterface : public PopulationInterfaceBase {
     }
     info->variable_map[this->log_init_naa.id_m] = &(population)->log_init_naa;
 
+    if (this->proportion_female.size() == 1 ||
```
awilnoaa, line 438: Since `proportion_female` is now user-defined, should invalid lengths hard-error here instead of warning and falling back to `0.5`? That seems like it might be easier for users to catch and debug.

inst/include/interface/rcpp/rcpp_objects/rcpp_population.hpp
```diff
@@ -413,6 +430,43 @@ class PopulationInterface : public PopulationInterfaceBase {
...
+      FIMS_WARNING_LOG(
+          "The proportion_female vector is not of size 1 or n_ages. Filling "
+          "with 0.5.");
+      this->proportion_female.resize(1);
+      this->proportion_female[0].initial_value_m = static_cast<double>(0.5);
+      this->proportion_female[0].estimation_type_m.set("constant");
+      population->proportion_female.resize(this->proportion_female.size());
+    }
+
+    for (size_t i = 0; i < this->proportion_female.size(); i++) {
```
awilnoaa, line 452: Correct me if I am wrong, but should `proportion_female` be bounded to `[0, 1]` here or earlier in the wrapper? I think with this setup users can pass values outside of `[0, 1]`?

awilnoaa (commented):
Eveything looks good, I left two comments on `rcpp_population.hpp`. I also wanted to ask if  `proportion_female` should be added to `rcpp_models.hpp` so it shows up in standard output for users to inspect after fitting?

inst/include/interface/rcpp/rcpp_objects/rcpp_population.hpp
awilnoaa: Hi @nathanvaughan-NOAA,  I was reviewing this and wanted to get your thoughts. If proportion_female can support fixed and random effects, should it also be bounded to [0,1] throughout estimation, in addition to checking the initial value?

inst/include/interface/rcpp/rcpp_objects/rcpp_population.hpp
```diff
@@ -135,6 +135,10 @@ class PopulationInterface : public PopulationInterfaceBase {
    * @brief The natural log of the initial numbers at age.
    */
   ParameterVector log_init_naa;
+  /**
+   * @brief Proportion female in the population.
```
kellijohnson-NOAA: More information here would be great.

inst/include/interface/rcpp/rcpp_objects/rcpp_population.hpp
```diff
@@ -283,6 +290,16 @@ class PopulationInterface : public PopulationInterfaceBase {
...
+
+      for (size_t i = 0; i < this->proportion_female.size(); i++) {
+        if (this->proportion_female[i].estimation_type_m.get() == "constant") {
+          this->proportion_female[i].final_value_m =
+              this->proportion_female[i].initial_value_m;
+        } else {
+          this->proportion_female[i].final_value_m =
+              pop->proportion_female.get_force_scalar(i);
+        }
+      }
```
kellijohnson-NOAA, line 307: I do not think this logic is right. "constant" means not estimated in the code. It does not mean that it is a single value across the vector. This is something that others have gotten confused about as well and we need to do some renaming eventually to alleviate that.

inst/include/interface/rcpp/rcpp_objects/rcpp_population.hpp
kellijohnson-NOAA: I think we want to error out if a user assigns it to a fixed effect or a random effect because we have no data to estimate the value right now.

kellijohnson-NOAA (changes requested):
I added some more comments. Additionally, I think we need a test in the testthat tests, demonstrating that if we reduce the proportion female that spawning biomass decreases compared to using a higher value or the default.

inst/include/interface/rcpp/rcpp_objects/rcpp_population.hpp
nathanvaughan-NOAA: Hey @awilnoaa the [0,1] bounding should be accomplished by using a logit transform of the input value, the same way we use a log transform for things like F_mort to constrain them as positive. @kellijohnson I feel like it's probably more future-proof to leave the potential for estimability than to set a rigid error, there are plenty of things like M or the length at age matrix that aren't practically estimable but are still technically allowed to be estimated.

inst/include/interface/rcpp/rcpp_objects/rcpp_population.hpp
nathanvaughan-NOAA: Hey @kellijohnson-NOAA I think this logic is ok. The "constant" here is actually referring to non-estimated not a single value. If "constant" it is just pushing the initial value into the reported final value, otherwise it is pulling the estimated value from population to fill the reported final value. The get_force_scalar is just letting us have a vector reported even if a single value was estimated.

kellijohnson-NOAA:
There are two failed tests in the testthat tests, see the [output here](https://github.com/NOAA-FIMS/FIMS/actions/runs/27292597830/job/80616842091). Let us know if you need help fixing them.

szu-yun-ko:
Thanks for the review! I should be able to address the comments and implement fixes to pass the test before next Monday.

inst/include/interface/rcpp/rcpp_objects/rcpp_population.hpp
szu-yun-ko: I've replaced the hard-errors with warnings for `proportion_female`'s bounding range.

szu-yun-ko:
I think this is ready for review again! I've made some changes to address the comments, and the `testthat` tests now passes. I'm not sure why `build-metrics`, and the `call-r-cmd-check` jobs failed, but from the error logs I feel like it might not be caused by changes in this PR. I've also added a `testthat` test to check that if we reduce the proportion female then spawning biomass decreases compared to using a higher value or the default.

awilnoaa (commented):
This PR looks good to me. A later follow-up or potential issue could be made to add a test that confirms that FIMS gives a clear error if `proportion_female` is less than `0` or greater than `1` or if its length is not `1` or `n_ages`.

kellijohnson-NOAA:
I will merge this in as soon as the tests finish. I had to rebase to main and force push so all the tests are rerunning.

PR #1495, opened 2026-05-30
Prior transformation with jacobian

### LINKED ISSUE #431: [Feature]: Parameter prior specification
### Is your feature request related to a problem? Please describe.

The user will need to be able to specify priors for fixed effects, and include a Jacobian adjustment when using MCMC when a transformation is used. 

### Describe the solution you would like.

The user should be able to specify diverse priors from the R side, and let FIMS handle the rest internally. So for instance one might put a normal prior on natural mortality, or a gamma prior on a process error like sigmaR. In these cases the parameter is (probably) declared and optimized in log space (logM), but the prior is applied in natural space like M~N(.2,.1). An automatic Jacobian (`nll+=logM`) needs to be done only when doing MCMC so that the prior information is correctly incorporated. The same goes for logit parameters and other transformations, although these will be much less common.

So on the user it it would be like the following, assuming these are Parameter class objects.

````
logM.priortype='normal'
logM.priorpars=c(.2,.1)
````
We'd need internal checks that priorpars and priortype are compatible and to have a default of 'none' so priors are opt-in. 

### Describe alternatives you have considered

N\A

### Statistical validity, if applicable

Parameter priors are a fundamental part of Bayesian assessments, but also serve as "penalties" in frequentist approaches. These are widely used, e.g., each SS3 parameter has a prior option.

### Describe if this is needed for a management application

N\A

### Additional context

This is related to Parameter bound issue https://github.com/NOAA-FIMS/FIMS/issues/410

### PR DESCRIPTION
**Pull request by Andrea-Havron-NOAA**
# What is the feature?
* Adds an automatic Jacobian adjustment for prior distributions
* Address issue #431 

* Sets up an interface that allows users to input priors using a formula interface
* Stores input and prior transformation metadata in a modified variable_map struct
* Adds new functions in the interface to parse the formula and pass prior and transformation information to the backend
* Adds new functions in a fims_transformation.hpp files that applies transformations and evaluates the Jacobian adjustment
* Adds a Prepare function to distribution_base to evaluate the prior transformation
* Add code to model.hpp to apply the Jacobian adjustment when the do_mcmc flag is true

* restructures variable_map which affects the Rcpp interface

___

## Checklist

### CODE REVIEW

**Comment by Andrea-Havron-NOAA** [comment]
See #431 for remaining tasks on draft PR

File: inst/include/common/fims_transformations.hpp [hunk 1]
```diff
@@ -0,0 +1,323 @@
...
+ * transformations to parameters, transforming parameters between input
+ * and prior spaces, and computing log Jacobian adjustments for
+ * change-of-variables corrections in MCMC sampling.
+ *
+ * The transformation system supports the following transformations:
+ * - `identity`: No transformation, parameter is on the natural scale.
+ * - `exp`: Exponential transformation, parameter is on the log scale.
+ * - `log`: Log transformation, parameter is on the natural scale.
+ * - `logit`: Logit transformation with optional bounds.
+ * - `square`: Square transformation.
```
nathanvaughan-NOAA, line 16: We could replace the square and sqrt transform options with a power option that takes an input value for the power square=2, sqrt=1/2 and this is then extendable to other powers.

inst/include/common/fims_transformations.hpp
```diff
@@ -0,0 +1,323 @@
...
+ * specified by the transformation argument. The forward transformations
+ * and are:
+ *
+ * | Label    | Forward          |
+ * |----------|------------------|
+ * | identity | x                |
+ * | log      | log(x)           |
+ * | exp      | exp(x)           |
+ * | logit    | logit(x, lo, hi) |
+ * | square   | x^2              |
```
nathanvaughan-NOAA, line 44: | power | power(x, power_value) |

inst/include/common/fims_transformations.hpp
```diff
@@ -0,0 +1,323 @@
...
+namespace fims_transformations {
+
+/**
+ * @brief Applies a forward transformation to a scalar value.
+ *
+ * @details Maps a value from natural scale to the transformed scale
+ * specified by the transformation argument. The forward transformations
+ * and are:
+ *
+ * | Label    | Forward          |
```
nathanvaughan-NOAA, line 38: I use x in the comment below for consistency with the other functions but we shouldn't use 'x' or any other similarly small variable names in the codebase as it makes interpretation and future refactors really difficult. Particularly with something like x being used to mean different things throughout the codebase.

inst/include/common/fims_transformations.hpp
```diff
@@ -0,0 +1,323 @@
...
+ * returned unchanged.
+ *
+ * @tparam Type The numeric type (e.g. double, TMBad::ad_aug).
+ * @param input_value A vector of parameter values in input space.
+ * @param input The transformation applied to the parameter in input space
+ * (e.g. log if the parameter is stored as log(sd)).
+ * @param prior The transformation applied to the parameter in prior space
+ * (e.g. square if the prior is on variance = sd^2).
+ * @return A vector of parameter values transformed to prior space.
+ */
```
nathanvaughan-NOAA, line 162: This could be extended to be more generalized and save compute time if it stores the input, natural, and prior transformed values somewhere for when they are needed throughout the code rather than calculating every time.

inst/include/common/fims_transformations.hpp
```diff
@@ -0,0 +1,323 @@
...
+template <typename Type>
```
nathanvaughan-NOAA, line 163: It would also make sense to extend this to simply a likelihood transformation rather than just limiting it to priors. For that we would need to tie the transformed values to the likelihood component of interest.

inst/include/common/information.hpp
```diff
@@ -139,10 +139,54 @@ class Information {
       uint32_t, std::shared_ptr<fims_popdy::FisheryModelBase<Type>>>::iterator
       model_map_iterator; /**< iterator for variable map>*/
```
nathanvaughan-NOAA, line 141: To extend this to all likelihood components we would have to have one map linked to the input variables like this that holds the input transformation type and the input and natural values. It could also be helpful here to have a reporting transformation and value specified so that use gets things out in the desired way. 

We would then need a second map linked to the likelihood components that specifies the transformation type and transformed values needed for each likelihood component. This allows parameters to have multiple transformations such as one for a prior and one for fit to observed data.

This approach would replace the manual log and natural parameters we currently specify manually in FIMS like log_F_mort and F_mort we would just specify which transformation is used in a specific internal calculation.

inst/include/common/types.hpp
```diff
@@ -0,0 +1,101 @@
...
+      return "logit";
+    case fims::Transformation::Label::square:
+      return "square";
+    case fims::Transformation::Label::sqrt:
+      return "sqrt";
+    default:
+      return "unknown";
+  }
+}
+
```
nathanvaughan-NOAA, line 72: This section on the distribution functions doesn't seem to fit with the rest of the refactor.

nathanvaughan-NOAA (commented):
Some initial comments. Don't need to address this stuff for this pull request but thoughts for how to improve on this in the future.

msupernaw:
@Andrea-Havron-NOAA  it looks like there is an issue with FIMS::fit_fims and do_mcmc thats causing R CMD CHECK to fail.

kellijohnson-NOAA:
@Andrea-Havron-NOAA this is a friendly reminder that in the Requirements spreadsheet we put a deadline of FY26Q4 on this PR, which is next week.

PR #1523, opened 2026-06-08
Update build-metrics.yml

msupernaw:
adds a threshold and only tracks max rss for regression.

# What is the feature?
*This PR updates the build metrics workflow so that workflow failures are based solely on kernel-reported peak memory usage during FIMS installation. 

* Added a configurable MAX_RSS_THRESHOLD_MB environment variable.
* Removed regression-based workflow failures tied to percentage increases in reported metrics.
* Retained reporting of all existing metrics, including:
    * Install time
    * Sampled process-tree RSS
    * Process count
    * Package size
    * Shared object size
    * Warning and error counts
* Updated workflow gating so failures occur only when:
    package_install_max_rss_mb > MAX_RSS_THRESHOLD_MB

*  no

___

## Checklist

### DIFF
--- a/.github/workflows/build-metrics.yml
+++ b/.github/workflows/build-metrics.yml
@@ -24,7 +24,7 @@ jobs:
       GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
       BASELINE_REPOSITORY: ${{ github.event.pull_request.base.repo.full_name || github.repository }}
       CURRENT_COMMIT_SHA: ${{ github.event.pull_request.head.sha || github.sha }}
-      REGRESSION_PCT: 15
+      MAX_RSS_THRESHOLD_MB: 12000
 
     steps:
       - uses: actions/checkout@v6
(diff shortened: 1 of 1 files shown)


kellijohnson-NOAA:
@msupernaw I don't think that we should rely on a single metric. Also, it is unclear what the RSS metric actually is.

msupernaw:
> @msupernaw I don't think that we should rely on a single metric. Also, it is unclear what the RSS metric actually is.

@kellijohnson-NOAA sampled RSS, process counts, package size, and install time are all useful things to monitor. But if the concern is memory exhaustion on build, then it's max RSS we should use as pass/fail because it’s the kernel-reported maximum resident memory observed during execution.

kellijohnson-NOAA:
We won't "monitor" something though unless it fails when there is a change so reporting them without having a pass-fail metric is pointless. And, we care about package size, install time, and memory usage. So removing all of the threshold checks except for RSS essentially means that only memory will be looked at, which is not what we want.

kellijohnson-NOAA:
I am going to close this PR because, as we talked about this morning, it is important to monitor all metrics. But, maybe in the future we can change the threshold of a few of them.

PR #1531, opened 2026-06-10
chore: R universe and GitHub actions

### LINKED ISSUE #1462: Increase testing on forks

Only select actions are initiated to run on forks and as our developer pool increases, I would like the number of tests to be similar on forks as it is on internal PRs, especially when we are merging straight to main instead of dev. See below for a screenshot that shows the tests currently being ran on a fork.

### Additional context

Image

_No response_

_No response_

kellijohnson-NOAA:
# What is the feature?
* Use FIT R-universe links because the noaa-fims r-universe is being deleted
* Enable R CMD Check on PR to main
* Enable Google Tests on PR to main

* Changed links from `https://noaa-fims.r-universe.dev` to `https://noaa-fisheries-integrated-toolbox.r-universe.dev`
* Added on PR to main to the R CMD Check and Google Test workflows, as well as added some paths to ignore

* No

### DIFF
--- a/.github/workflows/call-r-cmd-check.yml
+++ b/.github/workflows/call-r-cmd-check.yml
@@ -3,6 +3,18 @@ name: call-r-cmd-check
 
 on:
   push:
+  pull_request:
+    branches:
+      - main
+    paths-ignore:
+      - .devcontainer
+      - .github
+      - '.gitignore'
+      - '.Rbuildignore'
+      - 'CONTRIBUTING.md'
+      - 'LICENSE'
+      - 'man'
+      - 'README.md'
   workflow_dispatch:
 
 jobs:
--- a/.github/workflows/run-googletest.yml
+++ b/.github/workflows/run-googletest.yml
@@ -3,6 +3,18 @@ name: run-googletest
 
 on:
   push:
+  pull_request:
+    branches:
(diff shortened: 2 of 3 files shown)


.github/workflows/call-r-cmd-check.yml
```diff
@@ -3,6 +3,18 @@ name: call-r-cmd-check
...
+    branches:
+      - main
+    paths-ignore:
+      - .devcontainer
+      - .github
+      - '.gitignore'
+      - '.Rbuildignore'
+      - 'CONTRIBUTING.md'
+      - 'LICENSE'
+      - 'man'
```
Bai-Li-NOAA, line 16: Maybe remove "man"? If something changed in the documentation, does that mean the authors also changed something in R?

Bai-Li-NOAA (approved):
Looks good! I just had a minor question about the `man` folder.

.github/workflows/call-r-cmd-check.yml
kellijohnson-NOAA: Sometimes updating the man folder is forgotten about and then the formatting PR catches it, the formatting PR being merged in then leads to the actions being run again. So, I think I will leave it.

PR #1554, opened 2026-06-18
fix memory leak

### LINKED ISSUE #1553: Fix memory leak from not clearing pointers

Running R's profvis on select functions in `prepare_test_function()` demonstrated memory leaks in R. I suspect this is coming from pointers that are not being properly cleared.

### Additional context

_No response_

Linux

0.9.3

Andrea-Havron-NOAA:
# What is the feature?
* Fixes memory leak in R caused by not clearing pointers

- add recruitment shared pointers to clear function
- add functionality in FIMSObject to track objects for memory leak test and identify source module for memory leak
- add memory leak test

* No

I was able to track down which pointers were not being cleared using a new memory test that adds module registration for memory tracking purposes. The clear function now will throw an error if there is an uncleared pointer and will list the base class and module id. For example, I used the following warning to track down the uncleared pointers were coming from RecruitmentBase:

  ⚠️  WARNING: FIMS Memory Leak Detected after clear()!
  --------------------------------------------------
  The following 4 C++ heap object(s) were NOT destroyed:
    ❌ Leaked: fims_popdy::RecruitmentBase<TMBad::global::ad_aug> [id=0]
    ❌ Leaked: fims_popdy::RecruitmentBase<TMBad::global::ad_aug> [id=1]
    ❌ Leaked: fims_popdy::RecruitmentBase<double> [id=0]
    ❌ Leaked: fims_popdy::RecruitmentBase<double> [id=1]
  --------------------------------------------------
  Ensure all std::shared_ptrs to these types are being reset.

___

## Checklist


Andrea-Havron-NOAA:
I noticed in the build metrics reports: `sampled process tree RSS peak increased by 22.3%`. @msupernaw, would this increase be due to my include statemens like `#include <cxxabi.h>`? About half of the code added in this PR is to help produce readable error messages to help track down the location of the pointer that is not being cleared. But thinking about it more, I'm not sure all this extra code is neccessary. It was useful to find the right modules in the case where I had no idea where to look, but moving forward, this error will only be thrown when new pointers are added that haven't been cleared. It will be obvious to the developer which pointers to check. Should I strip the PR of the extra code to produce readable error messages and see if the build metrics improve?

kellijohnson-NOAA:
@Andrea-Havron-NOAA the build metric has been highly variable for that one, with very small changes leading to 25% differences so I think it is just the instability of the runner. We need to change the threshold for that metric as Matthew pointed out to me before. But, I am curious to hear what @msupernaw thinks about the code to help debug that is in this PR.

msupernaw:
@Andrea-Havron-NOAA and @kellijohnson-NOAA That metric is vulnerable to all kinds of influence like system state. It only becomes informative when peak RSS increases, which is reported directly from the kernel. I would not expect peak RSS to move much unless we create a TU larger than TMB.o.

msupernaw:
@Andrea-Havron-NOAA Good catch! Approved after WORDLIST is updated.

kellijohnson-NOAA:
@Andrea-Havron-NOAA can you let me know when this is ready. I see that you commented out some C++ code and now the google tests are failing. Also, I am not sure if we need everything that was in this original PR as you noted above and @msupernaw forgot to comment on.

Andrea-Havron-NOAA:
There is an issue with using `std::multiset` that I am working on resolving. If it looks like it will take too long or is not an easy fix, I will just simplify the PR as I mentioned.

kellijohnson-NOAA:
@Andrea-Havron-NOAA I was surprised that `devtools::test()` did not pass. But, that workflow is a non-blocker. Is this ready to merge in?

Andrea-Havron-NOAA:
> @Andrea-Havron-NOAA I was surprised that `devtools::test()` did not pass. But, that workflow is a non-blocker. Is this ready to merge in?

`devtools::test()` is also failing on main - it looks like there's an issue with `test-rcpp-fims.R`

This is ready to be merged into main after commits are squashed.

tests/testthat/test-clear.R
kellijohnson-NOAA, line 1: @Andrea-Havron-NOAA it seems like we could just add the two lines (23--24) to a different test that initializes a module instead of having a dedicated file for this. @Bai-Li-NOAA what are your thoughts?

kellijohnson-NOAA (approved):
I made one comment, once that is discussed I will merge this in.

tests/testthat/test-clear.R
Bai-Li-NOAA: Either approach works. I actually like keeping it as a separate file since it's specifically testing `clear()`, especially if we want to check the leak detection message.

Non-blocking suggestion: just from looking at the test file, I wasn't sure what `clear_test()` meant. My first impression was that it was clearing a test. After reading the PR description and looking through the other files, I understood the intent. Maybe we could rename it to `test_clear()` or `test_clear_with_leak_check()` ?

Andrea-Havron-NOAA: @Bai-Li-NOAA, thanks for the suggestion, I can implement that change. One of my motivations for putting this test in its own file was to ensure this test runs first because having a dangling pointer could make other tests fail or crash in ways that would be hard to detect.

Andrea-Havron-NOAA:
> > @Andrea-Havron-NOAA I was surprised that `devtools::test()` did not pass. But, that workflow is a non-blocker. Is this ready to merge in?
> 
> `devtools::test()` is also failing on main - it looks like there's an issue with `test-rcpp-fims.R`
> 
> This is ready to be merged into main after commits are squashed.

The call `methods::new(Parameter, "a")` is crashing the R session instead of just returning the error message. I've fixed this before - I can look into what has changed.

Andrea-Havron-NOAA:
> I made one comment, once that is discussed I will merge this in.

the test name is changed, I think this is ready to merge into main now.

Issue #1557, opened 2026-06-20
Implement partition infrastructure for sex-structured catch-at-age models


szu-yun-ko:
### Summary
This issue addresses the need to introduce a partition infrastructure layer in FIMS that generalizes how population structure (initially sex) is represented, indexed, and stored, without changing population dynamics or likelihood behavior yet.

### Current Status
Today, FIMS treats abundance as a single pooled quantity $N_{a,y}$ and folds it as `i_age_year = year × n_ages + age`. Sex enters only through `Population::proportion_female`, which affects spawning biomass but not landings or index expectations. I looked through the codebase and finds that it already anticipates partitions in documentation (ref: [population.hpp](https://github.com/NOAA-FIMS/FIMS/blob/main/inst/include/population_dynamics/population/population.hpp#L20-L23)), but nothing is wired so far. 
```cpp
/**
 * @brief Population class. Contains subpopulations
 * that are divided into generic partitions (e.g., sex, area).
 */
```

### Implementation Details
Detailed design of the data infrastructure can be found in:  [Google Summer of Code's Project Proposal #114](https://github.com/user-attachments/files/29163056/IOOS.Project.Number.114.Proposal.pdf). The proposal separates structure (how the population is partitioned), indexing (how flat vectors map to year/age/stratum), and storage (where pooled vs. partitioned quantities live). This issue intends to tackle the structure, indexing and storage structure mentioned in the proposal, whereas model dynamics and likelihood wiring come later.

#### Representation of Sex
> Recap: Population structure is described through axes, which are inherently named dimensions with discrete levels. Sex is the first axis, including two levels: female and male. The combination of levels from different axes forms strata (for instance, if we have a sex axis and a region axis, then `{sex = "female", region = "south"}` would be one strata). A `GroupSelector` picks one or more strata and supports wildcards, so a single request can mean e.g. “all sexes” or “female only.” Each stratum has a split factor, at this stage the split factor for sex is simply `Population::proportion_female` parameter.
- [x] Define `Axis`, `GroupSelector`, `PartitionSpec` with `n_strata()`, `stratum_id()`, `levels_from_stratum()`, `expand_group_to_strata()`
- [x] Initialize default sex axis on each `Population` in `CatchAtAge::Initialize()`

#### Indexing Layout
[...]

PR #1581, opened 2026-06-25
Add partition spec and index layout infrastructure for sex strata

szu-yun-ko:
# What is the feature?
This PR adds the first layer of partition infrastructure for population dynamics. It addresses parts 1–2 of [issue #1557](https://github.com/NOAA-FIMS/FIMS/issues/1557) only (representation of sex and indexing layout). Model dynamics, partitioned derived quantities, and likelihood wiring are intentionally deferred to follow-up PRs.

- Flesh out `subpopulation.hpp` with `Axis`, `GroupSelector`, `PartitionSpec`, `IndexLayout`, and `MakeDefaultSexPartitionSpec()` (female/male sex axis).
- Add `partition_spec` and `index_layout` members to `Population`.
- Initialize the default sex partition and index layout for each population in `CatchAtAge::Initialize()`.
- Add gtests for partition/indexing utilities and for partition initialization via `CatchAtAge::Initialize()`.

### Design Notes
- Partition types live in `subpopulation.hpp`, matching existing documentation that subpopulations represent generic partitions.
- `IndexLayout::i_age_year()` preserves the existing pooled (year, age) layout; `i_stratum_age_year()` adds stratum as the leading dimension for future partitioned containers (n_strata × n_years × n_ages).
- Sex is the first (and only) axis for now; GroupSelector supports wildcards so the same API can extend to additional axes (e.g., area) later.

* No.

___

## Checklist


szu-yun-ko:
@awilnoaa @nathanvaughan-NOAA @Andrea-Havron-NOAA This PR should be ready for an initial review. This PR addresses the tasks mentioned in issue #1557, after the review comments are addressed, I will continue tackling tasks in issue #1587 in this same PR. 

### Partition representation: 
- `Axis`: a partition dimension (e.g. sex) with a vector of named levels. The size of an Axis is inferred from the length of its level vector.
- `GroupSelector`: a partial selection pattern that can be mapped to one stratum or many strata, where each stratum is a combination of a level from each axis. 
- `PartitionSpec`: the schema of the partition grid, it includes a vector of axes (each having its own name and levels). Member functions include `n_strata()`, `stratum_id()`, `levels_from_stratum()`, and `expand_group_to_strata()`. These member functions are primarily used for bridging user-facing partition concepts and flat storage indices. 
   - A stratum is the atomic storage unit (one level chosen on every axis). 
   - A group (`GroupSelector`) is a looser selection that may match one or many strata. 
   - `expand_group_to_strata()` resolves a group to the strata that should be read or written.
   - `stratum_id()` and `levels_from_stratum()` convert between those strata and the integer indices that appear in partitioned vectors. 
   - `n_strata()` sizes those containers (e.g. n_strata × n_years × n_ages for fleet at-age DQs). 
- `IndexLayout`: provides folded index helpers for pooled (year, age) and partitioned (stratum, year, age) storage. 
- `MakeDefaultSexPartitionSpec()`: a function for setting up the current (default) partition, which is to include one axis (sex) with two levels (female and male). 

### Wiring into population: 
- `Population` gains `partition_spec` and `index_layout` members (population.hpp).
- `CatchAtAge::Initialize()` sets the default sex partition and populates `index_layout` from `n_years`, `n_ages`, and `partition_spec.n_strata()` (catch_at_age.hpp).

### Partitioned derived-quantity storage:
- Registers four fleet at-age `*_by_partition` DQs in rcpp_models.hpp:
  - `landings_numbers_at_age_by_partition`
  - `landings_weight_at_age_by_partition`
  - `index_numbers_at_age_by_partition`
  - `index_weight_at_age_by_partition`
- Each is sized `n_strata × n_years × n_ages` with `DimensionInfo {"n_strata", "n_years", "n_ages"}`
- Note that in this PR we implement the storage only. These containers are not filled during Calculate* yet (vectors are allocated and zeroed on `Prepare()`). This will be addressed later on (ref: issue #1587) along with the demand-based calculation routing logic. 

### Tests added:
**GTest**
- `test_population_Subpopulation.cpp`: default sex specifications, stratum encoding and decoding, stratum group conversion (include wildcard and single-stratum cases), folded indices.
- `test_population_CatchAtAge_InitializePartition.cpp`: `Initialize()` sets partition state on `Population`.
- `test_fleet_Fleet_InitializePrepare.cpp`:  Check that for each fleet, the four `*_by_partition` vectors have size `n_strata × n_years × n_ages`, while pooled at-age DQs remain `n_years × n_ages`. Also check that after `Prepare()`, the same partitioned vectors are all zeros. This confirms storage is allocated and reset even though `Calculate*` does not fill partitioned values yet.

**testthat**
- Column names: `expected_colnames` now includes `stratum_i`, the R-side name for the `n_strata` dimension after reshaping JSON model output.
- Regenerated because registering partitioned fleet DQs adds rows to `get_estimates()` on the integration fixtures (deterministic and estimation runs).

inst/include/population_dynamics/population/subpopulation.hpp
```diff
@@ -1,21 +1,167 @@
...
+   */
+  size_t stratum_id(const std::vector<size_t> &levels) const {
+    size_t id = 0;
+    size_t multiplier = 1;
+    for (int i = static_cast<int>(axes.size()) - 1; i >= 0; --i) {
+      id += levels[i] * multiplier;
+      multiplier *= axes[i].size();
+    }
+    return id;
+  }
```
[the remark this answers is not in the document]
szu-yun-ko: I added validation before encoding so that `levels.size()` must match `axes.size()`, and each `levels[i]` must be < `axes[i].size()`. Invalid input throws `std::invalid_argument` with a descriptive message. I didn't adopt Copilot's suggestion because returning 0 on error will map invalid inputs to stratum 0 (female), which is misleading. I've also added the tests `StratumIdRejectsMismatchedLevelsSize` and `StratumIdRejectsOutOfBoundsLevel` to make sure exceptions are thrown upon invalid inputs.

inst/include/population_dynamics/population/subpopulation.hpp
```diff
@@ -1,21 +1,167 @@
...
+   * @brief Decode a flat stratum index to one level per axis.
+   */
+  std::vector<size_t> levels_from_stratum(size_t stratum) const {
+    std::vector<size_t> levels(axes.size());
+    for (int i = static_cast<int>(axes.size()) - 1; i >= 0; --i) {
+      levels[i] = stratum % axes[i].size();
+      stratum /= axes[i].size();
+    }
+    return levels;
+  }
```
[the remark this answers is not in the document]
szu-yun-ko: Added a bounds check so that `stratum >= n_strata()` throws `std::invalid_argument` before decode.  I did not add a zero-size axis guard here because axes are defined internally via `MakeDefaultSexPartitionSpec()` for now rather than being user-defined. The test `LevelsFromStratumRejectsOutOfBoundsStratum` was also added to make sure exceptions are thrown upon out of bounds stratum ID.

inst/include/population_dynamics/population/subpopulation.hpp
```diff
@@ -1,21 +1,167 @@
...
+    std::vector<size_t> strata;
+    if (axes.empty()) {
+      strata.push_back(0);
+      return strata;
+    }
+
+    std::vector<size_t> current_levels(axes.size(), 0);
+    expand_group_to_strata_recursive(group, 0, &current_levels, &strata);
+    return strata;
+  }
```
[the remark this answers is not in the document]
szu-yun-ko: Before recursion, the function now validates that `group.level.size() == axes.size()` and that each non-wildcard entry is in `[0, axes[i].size())`. Invalid selectors throw `std::invalid_argument`.

szu-yun-ko:
This PR already adds partition structure (`PartitionSpec`, `IndexLayout`) and storage (fleet `*_by_partition` DQs). I've addressed the Copilot comments and added the split math we’ll need when writing into those partitioned containers (but `Calculate*` does not call any of this yet). The wiring will be addressed after the infrastructure is reviewed so that I can make sure I'm not building on unsolid grounds. The two helpers added in the latest commits include:

`SexStratumSplitFactors(spec, proportion_female)`: 
 - For the default sex-only partition only (`MakeDefaultSexPartitionSpec()`, 2 strata). It returns `{p_female, 1 - p_female}`. `proportion_female` comes from `Population::proportion_female.get_force_scalar(age)` at evaluation time and it is not stored on PartitionSpec. It's not wired at the moment but see below for a sketch of how it will work later on. 

`PartitionSpec::stratum_split_factor(stratum, split_factors)`:
- This is a generic lookup for `split_factors[stratum]` with validation. `split_factors` is passed in by the caller and it is not a field on `PartitionSpec`. The spec defines how many strata exist, the caller supplies how much goes to each.

```cpp
// Pooled write (unchanged)
const Type pooled = (Fmort * f_multiplier * selectivity) / Z
                    * N_aa * (1 - exp(-Z));
fdq_["landings_numbers_at_age"][i_age_year] += pooled;

// Partitioned write (future; sex-only Model 1)
const auto split_factors = SexStratumSplitFactors(
    population->partition_spec,
    population->proportion_female.get_force_scalar(age));

for (size_t s = 0; s < population->partition_spec.n_strata(); ++s) {
  const size_t i_part =
      population->index_layout.i_stratum_age_year(s, year, age);
  fdq_["landings_numbers_at_age_by_partition"][i_part] +=
      pooled * population->partition_spec.stratum_split_factor(s, split_factors);
}
```
The changes in this PR so far addresses all tasks mentioned in issue #1557 and I'll proceed to work on #1587 after the current work is reviewed.

inst/include/population_dynamics/population/subpopulation.hpp
```diff
@@ -1,21 +1,271 @@
...
+    }
+    size_t n = 1;
+    for (const Axis &axis : axes) {
+      n *= axis.size();
+    }
+    return n;
+  }
+
+  /**
+   * @brief Encode one level per axis as a flat stratum index.
```
Andrea-Havron-NOAA, line 111: Add a more detailed description here using the example of male/female for sex. Do the same for all functions in subpopulations.hpp that are missing a `@details` section.

inst/include/interface/rcpp/rcpp_objects/rcpp_models.hpp
```diff
@@ -1196,6 +1201,17 @@ class CatchAtAgeInterface : public FisheryModelInterfaceBase {
       info->variable_map[fleet_interface->landings_numbers_at_age.id_m] =
           &derived_quantities["landings_numbers_at_age"];
 
+      // partitioned landings and index at age (storage only; not filled yet)
+      derived_quantities["landings_numbers_at_age_by_partition"] =
```
nathanvaughan-NOAA: It doesn't look like any of these partitioned derived quantities are being pushed back to info->variable_map. In order to do that I think they will also need to be initialized as VariableVectors in the fleet and population .hpp files. You could look at how their respective non-partitioned quantities are handled to figure out which files are appropriate.

inst/include/population_dynamics/population/subpopulation.hpp
```diff
@@ -1,21 +1,271 @@
...
+  }
+};
+
+/**
+ * @brief Folded indices for pooled and partitioned derived quantities.
+ *
+ * @details Pooled quantities use i_age_year(). Partitioned quantities add
+ * stratum as the leading dimension via i_stratum_age_year().
+ */
+struct IndexLayout {
```
nathanvaughan-NOAA, line 253: The code here looks to be limited to a single new axis layer with hard coded years/ages axes. Do you have a plan for extending this to nested axes such as sex and area?

szu-yun-ko: Great question! Multi-axis nesting (e.g. sex × area) is handled by `PartitionSpec`, not by nesting more dimensions into `IndexLayout`. `PartitionSpec` holds a vector of axes, and `n_strata()` is the product of each axis’s level count (ref. subpopulation.hpp around line 47). Adding area (or season) increases `n_strata` and it does not require nested helpers like `i_sex_area_year_age`. The current wiring pulls `n_strata` from the spec in `CatchAtAge::Initialize()` (ref. catch_at_age.hpp around line 160):

```cpp
      this->populations[p]->partition_spec = MakeDefaultSexPartitionSpec();
      this->populations[p]->index_layout.n_years = this->populations[p]->n_years;
      this->populations[p]->index_layout.n_ages = this->populations[p]->n_ages;
      this->populations[p]->index_layout.n_strata =
          this->populations[p]->partition_spec.n_strata();
```

Andrea-Havron-NOAA:
@szu-yun-ko, @nathanvaughan-NOAA will be out for three weeks so I will be taking over primary review responsibilities while he is gone. I noticed you recently pushed changes to this PR. Is this ready for review now, or are you planning to add more commits first?

szu-yun-ko:
@Andrea-Havron-NOAA No problem, thanks for letting me know. The current pushed commits are mainly for addressing the previous review comments. I’ve made more changes locally that addresses tasks mentioned in issue #1587, but I’m still making some final tests and fixes. It will be pushed for review in a few days! Thanks for checking in!

szu-yun-ko:
@Andrea-Havron-NOAA @awilnoaa I've pushed recent changes that mainly address issue #1587. The only part left would be the R-interface, but I thought we could make sure the backend behaves as we intended before adding the interface. As the PR is getting larger, below is a brief description of every commit, mainly why we need the changes and how the changes were implemented. Please let me know if anything else is needed during the review process. 

---
#### **1. `e35456eb` — Demand as a named-axis request**
Need a way to say “I want female only / male / both / nothing” that can grow to multi-axis later, without a sex-only enum.

**How it works:** The demand can be given as a C++ list (e.g. `list(sex = "female")`), which maps to stratum ids (e.g. `{0}`). An empty demand corresponds to no strata, therefore no partitioned results would be written. 

```cpp
struct AxisLevelSelection {
  std::string axis_name;                 /*!< axis name (e.g. "sex") */
  std::vector<std::string> level_names; /*!< requested level labels */
};

/**
 * @brief User demand for which partition strata to materialize in output.
 *
 * @details Empty selections means pooled output only (no partitioned write) —
 * the backward-compatible default. ...
 */
struct PartitionDemand {
  std::vector<AxisLevelSelection> selections; /*!< empty => pooled */

  /** @brief True when no partitioned strata are requested. */
  bool is_pooled() const { return selections.empty(); }
};
```

- Considering extendability, we need to reframe from using sex-specific enums for declaring demand (e.g. `kFemale`, `kMale`…etc.). We need a way for the user to specify axis + level as a list, so in the future the user can be specific about what level they want on each axis (e.g. `list(sex = “female”, area = “north”)`).
- In order to do that we declare a struct called `AxisLevelSelection`, which is the combination of an axis + a level on that axis. Then the `PartitionDemand` struct will simply consist of a vector of `AxisLevelSelection` (i.e. a vector of axis + level combinations).
    - PartitionDemand also includes a boolean flag to note whether the demand is pooled or not. When no partitioned strata are required, `is_pooled()` will be flagged as true.

```cpp
inline std::vector<size_t> RequestedStrata(const PartitionSpec &spec,
                                           const PartitionDemand &demand) {
  if (demand.is_pooled()) {
    return {};
  }
  return spec.expand_group_to_strata(MakeGroupSelectorFromDemand(spec, demand));
}
```

- Given a `PartitionSpec` and a `PartitionDemand`, we can resolve which strata is requested via the `RequestedStrata` function, which first creates a group from the demand and then expand the group to strata IDs.

---

#### **2. `acd83586` — Hang demand on the population**

Evaluation must read demand from the population `Population::partition_demand`. Default must stay backward-compatible (pooled): `Initialize()` sets `MakePooledPartitionDemand()`.

**How it works:** Every run starts pooled unless something (later: R interface; today: tests) replaces demand after init.

```cpp
  // partition infrastructure (see subpopulation.hpp)
  PartitionSpec partition_spec; /*!< axes and strata for partitioned structure */
  IndexLayout index_layout;     /*!< folded indices for partitioned quantities */
  PartitionDemand partition_demand =
      MakePooledPartitionDemand(); /*!< which strata to materialize; empty = pooled */
```

```cpp
      this->populations[p]->partition_spec = MakeDefaultSexPartitionSpec();
      this->populations[p]->partition_demand = MakePooledPartitionDemand();
      this->populations[p]->index_layout.n_years = this->populations[p]->n_years;
      this->populations[p]->index_layout.n_ages = this->populations[p]->n_ages;
      this->populations[p]->index_layout.n_strata =
          this->populations[p]->partition_spec.n_strata();
```

---

#### **3. `b6217b9a` — One write helper for split-on-write**

Landings and index should share the same demand + sex-ratio logic, not duplicate it four times. So we need one place for splitting pooled cell onto sex-specific containers/planes: `WritePartitionedQuantityAtAge(...)`.

**How it works:**

1. `RequestedStrata`: if empty, return.
2. Else build `{p, 1−p}` via `SexStratumSplitFactors`.
3. For each requested stratum only:`partitioned[i_stratum_age_year] += pooled_value × factor`.

**Scope note:** *This helper is for the sex-ratio at age model only: split pooled value × `proportion_female` onto demanded sex strata. Explicit two-sex later should use the same demand/storage, but compute per sex in `Calculate*` rather than extending this splitter. Later on when the explicit two-sex model is also implemented, we can branch at evaluation to decide which path to use. Today, the code roughly looks like this (pseudo-code):*

```cpp
Evaluate()
  for each year, age:
    CalculateLandingsNumbersAA  → pooled landings + WritePartitioned… (Model 1 split)
    CalculateIndexNumbersAA     → pooled index   + WritePartitioned… (Model 1 split)
    ...
```

*Later on the branching might look like:*

```cpp
Evaluate() / Calculate*
  if model_type == sex_ratio_at_age:     // Model 1
      compute pooled catch
      write pooled DQ
      WritePartitionedQuantityAtAge(...)   // split by proportion_female

  else if model_type == explicit_two_sex: // Model 3
      compute female catch from N^f
      compute male catch from N^m
      write into *_by_partition per stratum
      optionally pooled = female + male
```

---

#### **4. `882cca93` — Plug the helper into evaluation**

After each pooled landings/index at-age update, call the helper with the same cell value so that partitioned DQs actually gets filled when the demand asks for them. 

**How it works:**

- Pooled path unchanged (`landings_*_at_age`, `index_*_at_age`).
- Same increment/weight also offered to the helper.
- With default pooled demand → `_by_partition` stays zero.
- If demand were female → female slot = pooled × p, male stays 0.

```cpp
      // Baranov Catch Equation
      Type landings_naa =
          (population->fleets[fleet_]->Fmort[year] *
           ...
      fdq_["landings_numbers_at_age"][i_age_year] += landings_naa;
      fims_popdy::WritePartitionedQuantityAtAge(
          fdq_["landings_numbers_at_age_by_partition"], landings_naa,
          population->partition_spec, population->index_layout,
          population->partition_demand,
          population->proportion_female.get_force_scalar(age), year, age);
```

**Testing:** Tests are added to make sure the demand-based computations work.
```cpp
    TEST_F(CAAEvaluateTestFixture, FemaleDemandWritesLandingsByPartition)
    {
        ...
        population->partition_demand =
            fims_popdy::MakeSexPartitionDemand({"female"});
        ...
        catch_at_age_model->CalculateLandingsNumbersAA(...);
        ...
            EXPECT_DOUBLE_EQ(
                dq_fleet["landings_numbers_at_age_by_partition"][female_idx],
                dq_fleet["landings_numbers_at_age"][i_age_year] * p_female);
            EXPECT_DOUBLE_EQ(
                dq_fleet["landings_numbers_at_age_by_partition"][male_idx], 0.0);
```

inst/include/models/functors/catch_at_age.hpp
```diff
@@ -157,6 +157,13 @@ class CatchAtAge : public FisheryModelBase<Type> {
         this->populations[p]->proportion_female[0] = static_cast<Type>(0.5);
       }
 
+      this->populations[p]->partition_spec = MakeDefaultSexPartitionSpec();
+      this->populations[p]->partition_demand = MakePooledPartitionDemand();
```
awilnoaa, line 161: This looks great. When you get to the user-facing choice, it may be helpful to include a test showing that a user request for partitioned output is applied during model setup and is not accidentally reset to pooled during initialization.

inst/include/interface/rcpp/rcpp_objects/rcpp_models.hpp
```diff
@@ -1183,6 +1183,11 @@ class CatchAtAgeInterface : public FisheryModelInterfaceBase {
           &derived_quantities_dim_info =
               model->GetFleetDimensionInfo(fleet_interface->id);
 
+      const size_t n_strata = fims_popdy::MakeDefaultSexPartitionSpec().n_strata();
```
awilnoaa, line 1186: I know the design mentions sex is the only axis for now. This may be outside the scope of this PR, but I am curious, @Andrea-Havron-NOAA: would the DQ storage eventually need to be sized from the active `PartitionSpec` so it stays in sync with `n_strata` changes?

inst/include/population_dynamics/population/subpopulation.hpp
```diff
@@ -1,21 +1,490 @@
...
+    }
+  }
+
+  GroupSelector group;
+  group.level.assign(spec.axes.size(), GroupSelector::kWildcard);
+
+  for (size_t axis_index = 0; axis_index < spec.axes.size(); ++axis_index) {
+    const Axis &axis = spec.axes[axis_index];
+    const AxisLevelSelection *selection = nullptr;
+    for (const AxisLevelSelection &candidate : demand.selections) {
```
awilnoaa, line 364: This looks good to me. When the R interface is added, it may be helpful to make sure users cannot accidentally give two different requests for the same axis, like sex = female and sex = male as separate entries. That way the backend always gets one clear request and does not have to silently choose one.

awilnoaa (commented):
Looks good overall! I had a few comments/questions but most can be addressed when the user-facing side is added.

inst/include/models/functors/catch_at_age.hpp
```diff
@@ -837,11 +855,17 @@ class CatchAtAge : public FisheryModelBase<Type> {
...
           (population->fleets[fleet_]->q.get_force_scalar(year) *
            population->fleets[fleet_]->selectivity->evaluate(
                population->ages[age], year)) *
           pdq_["numbers_at_age"][i_age_year];
+      fdq_["index_numbers_at_age"][i_age_year] += index_naa;
+      fims_popdy::WritePartitionedQuantityAtAge(
+          fdq_["index_numbers_at_age_by_partition"], index_naa,
+          population->partition_spec, population->index_layout,
+          population->partition_demand,
+          population->proportion_female.get_force_scalar(age), year, age);
```
Andrea-Havron-NOAA, line 865: Nice progress on partitioned DQ wiring. One concern: WritePartitionedQuantityAtAge still hard-codes sex behavior via proportion_female_at_age and SexStratumSplitFactors. Could we make the writer generic by using a split policy (or per-stratum weights) and keep sex as one policy implementation in a future PR? That would let area/season/etc. reuse the same write path without changing core indexing code.

inst/include/interface/rcpp/rcpp_objects/rcpp_models.hpp
Andrea-Havron-NOAA: Yes, I think this will need to be made more generic for keeping dimensions in sync. @szu-yun-ko, for the longer term design, will we need a PartitionSpec defined for each partitioning scheme (e.g. AreaPartitionSpec, SeasonPartitionSpec, etc.), or can this be a single generic PartitionSpec build from user-provided axes and levels?

Andrea-Havron-NOAA (approved):
Thanks for the PR — this is useful infrastructure for subpopulation partitioning.

My main takeaway is that the structural pieces (PartitionSpec, IndexLayout, partitioned-at-age write path) are now in place, while some allocation logic is still sex-specific (e.g., SexStratumSplitFactors, proportion_female_at_age). This is great incremental progress, and I'm looking forward to follow-up work that generalized split behavior and sets up dimensions based on user input.

inst/include/population_dynamics/population/subpopulation.hpp
szu-yun-ko: Thanks for catching this, I’ll add that validation (and a gtest) on the demand / R-interface PR.

inst/include/interface/rcpp/rcpp_objects/rcpp_models.hpp
szu-yun-ko: > I know the design mentions sex is the only axis for now. This may be outside the scope of this PR, but I am curious, @Andrea-Havron-NOAA: would the DQ storage eventually need to be sized from the active `PartitionSpec` so it stays in sync with `n_strata` changes?

I do agree that `*_by_partition` sizing should eventually come from the active `PartitionSpec` on the population, not `MakeDefaultSexPartitionSpec()` at registration, but at the moment this should be fine (since users cannot set partition structure at build time yet). 

> Yes, I think this will need to be made more generic for keeping dimensions in sync. @szu-yun-ko, for the longer term design, will we need a PartitionSpec defined for each partitioning scheme (e.g. AreaPartitionSpec, SeasonPartitionSpec, etc.), or can this be a single generic PartitionSpec build from user-provided axes and levels?

Even when we have more partitioning schemes, generic `PartitionSpec` built from user-provided axes and levels (sex, area, season, …) is still sufficient.  We don’t need separate types like `AreaPartitionSpec`, area/season are just more `Axis` entries; `n_strata()` is the product of axis sizes, and demand stays a named list over those axes.

inst/include/models/functors/catch_at_age.hpp
szu-yun-ko: Just to check if I'm getting the idea correct, the proposed structure would be:
- `PartitionDemand` selects which strata to fill
- `PartitionSpec` / `IndexLayout` handle indexing
- A pluggable split policy supplies per-stratum weights so `pooled_value * weight[stratum]` can be accumulated without sex-specific logic in the writer.

In that setup, Model 1 would use a `SexSplitPolicy` driven by `proportion_female`; area/season (or other axes) could either add new policies or use user-supplied stratum weights, reusing the same writer. Model 3 (explicit per-stratum dynamics) may bypass split-on-write and write stratum values directly, but would still use the same partition indexing.

Please let me know if this is a correct representation of your idea. If so, I can update the PR with a generic writer + `SexSplitPolicy` as the first policy implementation (with room for empirical/user weights and other policies later).

szu-yun-ko:
@nathanvaughan-NOAA After this week's meeting, I looked into extending the existing fleet derived quantities rather than adding separate `*_by_partition containers`. My understanding of your proposed implementation is that we would keep the same DQ names and extend their `DimensionInfo` from 2D to 3D when partitioning is active? Something like:
```cpp
// Current (what is implemented)
derived_quantities["landings_numbers_at_age"]
    = Vector(n_years * n_ages);                    // pooled
derived_quantities["landings_numbers_at_age_by_partition"]
    = Vector(n_strata * n_years * n_ages);         // partitioned

// The proposed approach
derived_quantities["landings_numbers_at_age"]
    = Vector(n_strata * n_years * n_ages);
derived_quantities_dim_info["landings_numbers_at_age"] =
    DimensionInfo("landings_numbers_at_age",
                  {n_strata, n_years, n_ages},
                  {"n_strata", "n_years", "n_ages"});

// Writes would use i_stratum_age_year instead of i_age_year
// Pooled/default case: n_strata = 1, so shape collapses to today's behavior
```
I think this is possible, but the main concern is that existing models, gtests, JSON output, and R workflows all assume `*_at_age` is (year × age), so resizing the DQs would a breaking change across the C++ code. And in some cases, for example, the current sex-ratio-at-age model, both pooled and partitioned values needs to be computed. Although we can also set it as another strata (like a "pooled" strata) within the 3D container, that ultimately uses the same memory as separate pooled + partitioned vectors when both are stored. Also, A single 3D DQ would always allocate (n_strata [+ 1]) × year × age even when partitioning is off.

awilnoaa:
> I think this is possible, but the main concern is that existing models, gtests, JSON output, and R workflows all assume `*_at_age` is (year × age), so resizing the DQs would a breaking change across the C++ code. And in some cases, for example, the current sex-ratio-at-age model, both pooled and partitioned values needs to be computed. Although we can also set it as another strata (like a "pooled" strata) within the 3D container, that ultimately uses the same memory as separate pooled + partitioned vectors when both are stored. Also, A single 3D DQ would always allocate (n_strata [+ 1]) × year × age even when partitioning is off.

@Andrea-Havron-NOAA and @nathanvaughan-NOAA If the concern is updating the existing models, gtests, JSON output, and R workflows all assume `*_at_age` is (year × age),we could have a code club focused on updating that piece and get some help moving it along more quickly. That still would not solve the issue in cases where both pooled and partitioned values need to be computed, but I wanted to share it as a potential option if that is the direction we decide to go.

szu-yun-ko:
Also, this PR: [Expose partition demand to R and gate partitioned fleet DQs](https://github.com/szu-yun-ko/FIMS/pull/2) is also ready for review (it's stacked on top of this one).

awilnoaa (approved):

kellijohnson-NOAA:
I plan on rebasing this to #1697 and merging it into main by tomorrow.

inst/include/models/functors/catch_at_age.hpp
nathanvaughan-NOAA: I think the issue @Andrea-Havron-NOAA is getting at is that there is no way to specify the strata-specific weighting. In the current setup it is hard coded to be based on population sex proportion. A generalized setup will need to be able to both set different population segmentation (like you mentioned with per stratum weights) and also to enable fleet specific weights such as fleet 1 fishes only in area 1 but fleet 2 fishes 50% in area 1 and 50% in area 2. Even for sex it could be useful to have fleet specific weighting for species that could feasibly have sex specific targeting (such as releasing berried lobsters).

kellijohnson-NOAA:
@szu-yun-ko I was going to bring this into main but as @nathanvaughan-NOAA brought up, it still needs some work. What I would like to see is
- [ ] merge in https://github.com/szu-yun-ko/FIMS/pull/2
- [ ] rebase to main, I have done this locally and there were very minor conflicts. Please let me know if you need help completing this task.
- [ ] Make this work more generic. The indexing/storage is generic, but the allocation rule is still sex-specific because subpopulation.hpp has `SexStratumSplitFactors()` and `WritePartitionedQuantityAtAge(..., proportion_female_at_age, ...)`. That means every partition is implicitly treated as “female/male,” even though `PartitionSpec` says it could be area, sex, fleet group, etc.

I think for right now we could have users supply a value that indicates how much each partition contributes to reproductive biomass to know if a strata contributes to recruitment but it should not be hard-coded as female. Please let us know if you need more guidance here.

szu-yun-ko:
@kellijohnson-NOAA @nathanvaughan-NOAA Thanks for the review! I’ll look into making the code more generic and come up with a pseudo-code sketch ASAP to align the design directions. I’ll reach out whenever I encounter any questions.

inst/include/models/functors/catch_at_age.hpp
szu-yun-ko: I've added two commits ([6cc8472](https://github.com/NOAA-FIMS/FIMS/pull/1581/commits/6cc8472c7e8eed4b2cc303ecc218596fa2132ec5) and [40ba5e5](https://github.com/NOAA-FIMS/FIMS/pull/1581/commits/40ba5e5ed5aa4759bfb57e3ea4e70651b0a72c24)) to separate write/indexing from how weights are chosen:
- `WritePartitionedQuantityAtAge` now takes an explicit per-stratum `split_factors` vector instead of `proportion_female`. The writer only does `pooled_value * weight[stratum]` for the strata selected by PartitionDemand.
- Catch-at-age call sites now build those weights with `SexStratumSplitFactors(...)` (the current sex policy) and pass them in. So population sex proportion is one policy at the call site, not hard-coded inside the writer.

I will also work on enabling fleet-specific weights and have that incorporated as a stacked PR on top of this.

szu-yun-ko:
Updated status on the checklist:
> 
> * [x]  merge in [Expose partition demand to R and gate partitioned fleet DQs szu-yun-ko/FIMS#2](https://github.com/szu-yun-ko/FIMS/pull/2)
> * [x]  rebase to main, I have done this locally and there were very minor conflicts. Please let me know if you need help completing this task.
> * [x]  Make this work more generic. The indexing/storage is generic, but the allocation rule is still sex-specific because subpopulation.hpp has `SexStratumSplitFactors()` and `WritePartitionedQuantityAtAge(..., proportion_female_at_age, ...)`. That means every partition is implicitly treated as “female/male,” even though `PartitionSpec` says it could be area, sex, fleet group, etc.
> 

On the final item, `WritePartitionedQuantityAtAge` no longer takes `proportion_female`. It takes an explicit per-stratum `split_factors` vector and only does `pooled_value * weight[stratum]` for strata selected by PartitionDemand. Catch-at-age call sites build those weights with `SexStratumSplitFactors(...)` (current Model 1 sex policy) and pass them in, so sex is one policy at the call site rather than inside the writer.

Follow-ups:
- Fleet-specific stratum weights (this addresses Nathan's earlier comment): today every fleet reuses the population sex weights. Next PR can optionally let a fleet supply its own length-n_strata weights and fall back to the population split when unset (e.g. area mix or sex-specific targeting).
- @kellijohnson-NOAA I think I need some additional pointers on “users supply a value for how much each partition contributes to reproductive biomass.” My current understanding is that this is separate from the split-on-write weights: those allocate pooled catch/N into strata for output, while a reproductive-contribution factor would control which strata enter spawning biomass / recruitment (today still via proportion_female). Specifically, we might add a separate per-stratum `reproductive_contribution` used in SB (and thus recruitment), with sex default `{1,0}` instead of hard-coding “female”? Please let me know if that's what you have in mind, and I can also address this in a follow-up PR.

kellijohnson-NOAA:
@szu-yun-ko thank you for these edits. @nathanvaughan-NOAA and @Andrea-Havron-NOAA are going to look at them in more detail today and provide comments. Tomorrow's meeting will be a great time to work through any questions that you have. I am really excited to think about partitioning in general rather than just through the lens of males and females. We talked today about the "users supply a value for how much each partition contributes to reproductive biomass" and we believe this should come from the maturity relationship specified for each partition. In theory, if you had males and females as two separate partitions and you did not want males to contribute to spawning biomass you could say that no males are mature. This would get around needing to say what proportion are males and females.
The bigger question that we would like you to think about, that I did not think about before and we will talk about in our meeting tomorrow, is --- we need to figure out how recruits from a strata are apportioned to each available strata. We think this should be a user-supplied vector for each strata. I don't believe we have talked about this before so it might be a new concept but we think this is the missing link to getting this worked wrapped up and into main.

szu-yun-ko:
@kellijohnson-NOAA I've pushed a few more fixes so that most checks have passed. In yesterday's meeting, we discussed merging this into the [main-gsoc-partition](https://github.com/NOAA-FIMS/FIMS/tree/main-gsoc-partition) branch, so I can retarget the explicit two sex [PR](https://github.com/szu-yun-ko/FIMS/pull/3) to the same branch. Also, after looking into your earlier comment regarding apportioning recruits into different strata, I added https://github.com/NOAA-FIMS/FIMS/issues/1625#issuecomment-5726741537 to propose a more generic approach towards the explicit two sex model. If this aligns with your vision of the explicit two sex work, I can resume working on https://github.com/szu-yun-ko/FIMS/pull/3 to make this happen.

PR #1583, opened 2026-06-25
Fix build metrics

### LINKED ISSUE #1477: build metrics is failing b/c it doesn't have a positive metric to compare to

The build metrics action looks for the last positive metric to compare to on the main branch but because there is not an artifact for this workflow on main where the action actually had a passing score (that is because the action is new and was merged into main on a fail) now subsequent actions are also failing. They should be looking to compare to the last artifact in main, not the last passing artifact.

The issue is that the workflow currently searches for the last successful push run (lines 359-367), but you want it to compare against the latest commit on main, regardless of whether that run succeeded or failed.

### Additional context

The .yml file should be changed to (according to copilot).
```
- name: Download latest main metrics artifact
  env:
    GH_TOKEN: ${{ github.token }}
  run: |
    set -euo pipefail
    echo "Looking for baseline metrics in $BASELINE_REPOSITORY."

    # Get the latest commit on main branch
    LATEST_MAIN_COMMIT=$(git -C $BASELINE_REPOSITORY rev-parse origin/main 2>/dev/null || echo "")
    
    if [ -z "$LATEST_MAIN_COMMIT" ]; then
      echo "Could not determine latest main branch commit."
      exit 0
    fi

    # Find the run for that commit (regardless of status)
    BASELINE_RUN_ID=$(gh run list \
      --repo "$BASELINE_REPOSITORY" \
      --workflow build-metrics.yml \
      --branch main \
      --event push \
      --limit 50 \
      --json headSha,databaseId,conclusion \
      --jq ".[] | select(.headSha == \"$LATEST_MAIN_COMMIT\") | .databaseId" | head -1)

    if [ -z "$BASELINE_RUN_ID" ]; then
      echo "No build-metrics run found for latest main commit $LATEST_MAIN_COMMIT."
      exit 0
    fi

    echo "Found run $BASELINE_RUN_ID for latest main commit $LATEST_MAIN_COMMIT."
    rm -rf baseline-metrics
    mkdir -p baseline-metrics

    if gh run download "$BASELINE_RUN_ID" \
      --repo "$BASELINE_REPOSITORY" \
      --name build-metrics-main \
      --dir baseline-metrics; then
      if [ -f baseline-metrics/metrics.json ]; then
        cp baseline-metrics/metrics.json previous_metrics.json
        echo "Downloaded baseline metrics from main commit $LATEST_MAIN_COMMIT (run $BASELINE_RUN_ID)."
      else
        echo "Run $BASELINE_RUN_ID had the artifact but no metrics.json file."
      fi
    else
      echo "No build-metrics-main artifact found for run $BASELINE_RUN_ID."
    fi

    if [ ! -f previous_metrics.json ]; then
      echo "No usable build-metrics-main artifact found for latest main commit."
    fi
```

_No response_

_No response_

kellijohnson-NOAA:
# What is the feature?
* Compare against the previous commit in main rather than main to main
* Look for warning or error, case agnostic
* Change to 40 % for threshold for RSS  

* Updates .github/workflows/build-metrics.yml so warning/error parsing is case-insensitive and catches R/compiler-style log lines more reliably. It also fixes fallback baseline builds so they sample process tree RSS and process count instead of reusing RSS approximations and hard-coding process count to 0.
ae50bdbb - Changes build metric for RSS to be 40%
* Adds SAMPLED_PROCESS_TREE_RSS_REGRESSION_PCT: 40 and lets each metric carry its own regression threshold. The sampled process tree RSS metric now gates at 40%, while the other gated metrics continue using the default 15%.
9aa36f2d - fix(build-metrics): Compare to previous commit on push to main
* Fixes push-to-main comparisons so they use ${{ github.event.before }} as the baseline commit instead of comparing the pushed commit against itself. PRs still compare against latest main, and fallback baseline builds now fetch the exact baseline SHA.

NO

I tested it for a PR on my fork but not for pushes to main.

___

## Checklist

### DIFF
--- a/.github/workflows/build-metrics.yml
+++ b/.github/workflows/build-metrics.yml
@@ -25,6 +25,11 @@ jobs:
       BASELINE_REPOSITORY: ${{ github.event.pull_request.base.repo.full_name || github.repository }}
       CURRENT_COMMIT_SHA: ${{ github.event.pull_request.head.sha || github.sha }}
       REGRESSION_PCT: 15
+      # Percent threshold for sampled process-tree RSS regressions; sampled peaks are noisier than /usr/bin/time RSS.
+      SAMPLED_PROCESS_TREE_RSS_REGRESSION_PCT: 40
+    defaults:
+      run:
+        shell: bash
 
     steps:
       - uses: actions/checkout@v7
@@ -87,27 +92,32 @@ jobs:
             } > "${label}_cgroup_memory_values.tsv"
           }
 
-          sample_memory_until_done() {
+          capture_memory_sample() {
             local prefix="$1"
             local command_pid="$2"
             local samples_file="${prefix}_memory_samples.tsv"
+            local timestamp_epoch
+            local process_tree_rss_kb=0
+            local process_count=0
+            local pids
 
-            printf "timestamp_epoch\tprocess_tree_rss_kb\tprocess_count\n" > "$samples_file"
+            timestamp_epoch=$(date +%s)
+
+            pids=$(collect_descendant_pids "$command_pid" | sort -n | uniq | awk '{ pids = pids sep $1; sep = "," } END { print pids }')
+            if [ -n "$pids" ]; then
+              process_tree_rss_kb=$(ps -o rss= -p "$pids" 2>/dev/null | awk '{sum += $1} END {print sum + 0}')
+              process_count=$(printf "%s" "$pids" | awk -F, '{print NF}')
+            fi
+
+            printf "%s\t%s\t%s\n" "$timestamp_epoch" "$process_tree_rss_kb" "$process_count" >> "$samples_file"
+          }
+
+          sample_memory_until_done() {
+            local prefix="$1"
+            local command_pid="$2"
+
             while kill -0 "$command_pid" 2>/dev/null; do
-              local timestamp_epoch
-              local process_tree_rss_kb=0
-              local process_count=0
-              local pids
-
-              timestamp_epoch=$(date +%s)
-
-              pids=$(collect_descendant_pids "$command_pid" | sort -n | uniq | awk '{ pids = pids sep $1; sep = "," } END { print pids }')
-              if [ -n "$pids" ]; then
-                process_tree_rss_kb=$(ps -o rss= -p "$pids" 2>/dev/null | awk '{sum += $1} END {print sum + 0}')
-                process_count=$(printf "%s" "$pids" | awk -F, '{print NF}')
-              fi
-
-              printf "%s\t%s\t%s\n" "$timestamp_epoch" "$process_tree_rss_kb" "$process_count" >> "$samples_file"
+              capture_memory_sample "$prefix" "$command_pid"
               sleep 1
             done
           }
@@ -117,6 +127,7 @@ jobs:
             shift
 
             capture_memory_snapshot "${prefix}_before"
+            printf "timestamp_epoch\tprocess_tree_rss_kb\tprocess_count\n" > "${prefix}_memory_samples.tsv"
 
             /usr/bin/time -v "$@" > >(tee "${prefix}.log") 2>&1 &
             local command_pid=$!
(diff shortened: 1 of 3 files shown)


Bai-Li-NOAA (commented):
Well, I need to admit that I don't fully understand the code in this PR. So I asked Copilot (GPT-5-mini) to review it, and here’s what it returned:

---

## Issues & Recommendations (ordered by severity)

### 1. Python: potential crash when baseline-sampling files are missing

**What I saw:**  
The Python script does:

```python
package_sampled_peak_process_tree_rss_kb = read_int("baseline_package_install_sampled_peak_process_tree_rss_kb.txt")
package_sampled_peak_process_tree_rss_mb = round(package_sampled_peak_process_tree_rss_kb / 1024, 2)
```

**Problem:**  
If `read_int` returns `None` (e.g., file missing), dividing `None` by `1024` will raise an exception and the workflow step will fail.

**Why this can happen:**  
The workflow may either:
- download a baseline artifact (no local sample files exist), or  
- build a baseline locally (which produces sample files)

The code must safely handle the “artifact downloaded” case.

**Suggested fix:**

```python
if package_sampled_peak_process_tree_rss_kb is not None:
    package_sampled_peak_process_tree_rss_mb = round(
        package_sampled_peak_process_tree_rss_kb / 1024, 2
    )
else:
    package_sampled_peak_process_tree_rss_mb = None
```

Alternatively, coalesce to `0` if numeric fallback is preferred (though `None` is clearer for missing data).

---

### 2. awk numeric comparison when computing max sampled value

**What I saw:**

```sh
awk -F '\t' 'NR > 1 && max < $2 {max = $2} END {print max + 0}'
```

**Problem:**  
This relies on implicit awk coercion and an uninitialized `max`. It usually works but is fragile due to string vs numeric comparisons.

**Suggested fix:**

```sh
awk -F '\t' '
NR > 1 {
    val = ($2 + 0)
    if (val > max) max = val
}
END {
    print (max + 0)
}' file.tsv
```

---

### 3. git fetch by SHA with `--depth 1` may fail in some setups

**What I saw:**

```sh
git clone --no-checkout --filter=blob:none ...
git -C baseline-main fetch --depth 1 origin "$BASELINE_COMMIT"
```

**Risk:**  
Fetching a single SHA with `--depth 1` can fail depending on server or Git configuration. It often works if the commit is reachable, but not always.

**Suggested improvement:**

```sh
if ! git -C baseline-main fetch --depth 1 origin "$BASELINE_COMMIT"; then
    git -C baseline-main fetch origin "$BASELINE_COMMIT" || {
        echo "couldn't fetch commit"
        return 1
    }
fi
```

Or:

- clone with `--depth 1 --branch main`, then `git checkout "$BASELINE_COMMIT"` (if reachable)

---

### 4. Shell portability & explicit bash usage

**What I saw:**  
The script uses bash features (`local`, arrays, functions, recursion).

**Risk:**  
GitHub Actions typically uses bash on `ubuntu-latest`, but explicitly setting it avoids surprises on other environments.

**Suggested change:**

```yaml
- name: Download baseline metrics artifact
  shell: bash
  run: |
    set -euo pipefail
```

---

### 5. Sampling final value / potential race condition

**What I saw:**  
`sample_memory_until_done` loops while `kill -0 "$command_pid"` and writes samples, then the main script waits and kills the sampler.

**Risk:**  
A final memory spike right after process exit may not be captured, and sampler termination may not flush the last sample.

**Suggestion:**  
After `wait "$command_pid"`, take one final synchronous sample before killing the sampler to ensure peak capture.

---

### 6. Regex counting approach (warnings/errors)

**What I saw:**  
You added `re.IGNORECASE` and filtering for summary lines like `"3 warnings generated"`.

**Suggestion:**  
This is generally good, but consider tightening patterns to avoid false positives (e.g., stack traces or unrelated text containing “warning” or “error”).  
Possible improvement: anchor to typical compiler formats like `warning:` or `error:` if applicable.

---

### 7. Baseline selection for push events

**What I saw:**  
For push events:

```python
BASELINE_COMMIT="${{ github.event.before }}"
```

**Note:**  
This is correct for comparing against the previous commit.

**Good catch already in place:**  
`github.event.before` can be all zeros in some cases (force push / initial commit), and you already handle that case.

---

### 8. Metric tuple change consistency

**What I saw:**  
Metrics tuples expanded from 4 → 5 fields.

**Risk:**  
There may be other parts of the script still assuming 4-tuple structure.

**Suggestion:**  
Double-check the codebase for any remaining unpacking logic.

---

### 9. Small style / maintainability suggestions

- Add a comment explaining `SAMPLED_PROCESS_TREE_RSS_REGRESSION_PCT` (units = percent, rationale for threshold like 40%)
- Consider extracting sampling logic into a separate script or reusable action if it grows further
```

kellijohnson-NOAA:
@Bai-Li-NOAA would you mind reviewing the second commit in this PR? The bash script was failing for Mac so I did some digging as to why and added it to this PR. I also removed some if/else logic in the install script to just use pak for all operating systems unless it doesn't work.

Bai-Li-NOAA:
> @Bai-Li-NOAA would you mind reviewing the second commit in this PR? The bash script was failing for Mac so I did some digging as to why and added it to this PR. I also removed some if/else logic in the install script to just use pak for all operating systems unless it doesn't work.

Looks like the second commit runs fine on macOS now. I'm not sure why it passes on my local machine. I don't remember ever installing OpenMP locally.

I think the bash script is getting harder and harder to maintain. I wonder if we could have Copilot create a FIMSverse package so we can test things out and see if it's a better approach.

PR #1594, opened 2026-06-30
Add gmrf and precision builder functions as precursors for eventually adding DSEM

e-perl-NOAA:
**Disclaimer**
I don't actually know what I'm doing here, but there was enthusiasm for progress being made on getting DSEM into FIMS so I embarked on this journey ⛵ with nothing but Copilot and a dream ☁️.  Please keep that in mind as you make comments and I beg you to explain things in a way that you would see them in those "XXX for Dummies" books. 

# What is the feature?
Getting the behind the scenes functions of gmrf and a precision builder (that can be used with more than just dsem) that are precursors seemed like a good point to start merging things in since myself and copilot/notebookLM then all started to get confused on how to proceed from here (it started going a bit off the rails once we started getting to what needed to be added in the `information.hpp` file) . I don't even think that these functions are "exposed" enough to write tests for? Maybe I'm wrong here and if so, @Bai-Li-NOAA I would DEFINITELY need your help.

Once it seems like there is consensus that these are a good way forward and/or good to go, they can be merged in and I/we can incrementally work on getting DSEM into FIMS in subsequent PRs. Hopefully that method confuses copilot/notebookLM (and honestly, myself) less.

* Add gmrf function
* Add precision builders function - this can eventually take more than just DSEM

___

## Checklist


Bai-Li-NOAA:
>I don't even think that these functions are "exposed" enough to write tests for? Maybe I'm wrong here and if so, @Bai-Li-NOAA I would DEFINITELY need your help.

Wow 😮 , you did it! I haven't taken a close look at the code in those files yet, but for `gmrf.hpp` and `precision_builder.hpp`, you could write C++ unit tests using `FIMS:::use_gtest_template()`. If you'd like, here are some [additional instructions for Fernando](https://github.com/NOAA-FIMS/FIMS/tree/main/tests#readme). 

For the `rcpp_*.hpp` files, yes, they'll need to be added to [`inst/include/interface/rcpp/rcpp_interface.hpp`](https://github.com/NOAA-FIMS/FIMS/blob/main/inst/include/interface/rcpp/rcpp_interface.hpp) and exposed to R. After that, we can write testthat tests for those modules. You can find an example [here](https://github.com/NOAA-FIMS/FIMS/blob/main/tests/testthat/test-rcpp-recruitment-interface.R).

I'm not sure what your timeline is for working on the tests, but I'll be setting aside dedicated time each week to work on testing-related issues. Feel free to join those sessions (starting July 9 and listed on the FIMS calendar) if you have any questions or would like to work through any testing challenges together!

e-perl-NOAA:
@Bai-Li-NOAA Yeah, that sounds great! I won't have time to work on tests this week, but I'll definitely have questions for you when I start working on it next week! Also, I love that you are leaning into me naming my AI/agents Fernando 🤣

kellijohnson-NOAA:
@nathanvaughan-NOAA can you review this PR?

e-perl-NOAA:
@kellijohnson-NOAA I need to add tests, I just haven't had time to work on them 😞

nathanvaughan-NOAA:
No worries @e-perl-NOAA, Fernando is a busy guy I'm sure he'll get to them eventually 😁 I'll just go through the functions.

inst/include/distributions/functors/gmrf.hpp
```diff
@@ -0,0 +1,86 @@
...
+    /** @brief Constructor. */
+    GMRF() : DensityComponentBase<Type>() {}
+
+    /** @brief Destructor. */
+    virtual ~GMRF() {}
+
+    /**
+     * @brief Evaluates the GMRF log probability density function.
+     * @details The evaluation follows: 
+     * \f[ \log(f(x)) = -\frac{1}{2} (x - \mu)^T Q (x - \mu) + \frac{1}{2} \log|Q| \f]
```
Andrea-Havron-NOAA: The formula is missing the constant term: -N\log(\sqrt(2\pi))
N is the size of the latent state.

Andrea-Havron-NOAA:
@e-perl-NOAA, there are two issues:

1. `beta_z` is a parameter vector that is used to both assemble and scale the precision matrix. We can simplify code by keeping `beta_z` a member of the dsem module ONLY and use the value to scale the precision matrix in the GMRF call. 
2. There isn't any code linking the DSEMPrecisinMatrixBuilder to Q in the GMRF. 
 
I started working on a comment that included a bunch of code - I realized it might be easier to set up a co-working session to go over these changes together. I think Fernando has gotten you to a nice spot, but at this point, writing more instructions to AI would probably take more time than implementing the needed changes directly!

inst/include/interface/rcpp/rcpp_objects/rcpp_dsem.hpp
```diff
@@ -0,0 +1,76 @@
...
+            path.type = ram_matrix(r, 0);       // Rho vs Gamma
+            path.from = ram_matrix(r, 1);       // Origin
+            path.to = ram_matrix(r, 2);         // Target
+            path.beta_index = ram_matrix(r, 3); // Index into beta_z
+            path.start = static_cast<Type>(ram_start_values[r]);
+            builder->paths.push_back(path);
+        }
+
+        // Register builder in Information Map (to be accessed by GMRF distribution)
+        info->dsem_builders[this->id] = builder;
```
Andrea-Havron-NOAA: `dsem_builders` does not exist in information. More code is needed to link the precision matrix builder to the GMRF both here and in information.

```

File: inst/include/interface/rcpp/rcpp_objects/rcpp_growth.hpp [hunk 3]
```diff
@@ -285,4 +285,76 @@ class EWAAGrowthInterface : public GrowthInterfaceBase {
 #endif
 };
 
+/**
+ * @brief Rcpp interface for DSEM growth.
```
Andrea-Havron-NOAA: remove `beta_z` from the next four files (rcpp_growth, rcpp_maturity, rcpp_recruitment, and rcpp_selectivity)

inst/include/interface/rcpp/rcpp_objects/wip.cpp
```diff
@@ -0,0 +1,75 @@
+/**
```
Andrea-Havron-NOAA: this file can be removed, beta_z will be linked in a different way

inst/include/distributions/functors/gmrf.hpp
```diff
@@ -0,0 +1,86 @@
...
+        if (static_cast<size_t>(this->precision_matrix_ptr->rows()) != n_x) {
+            throw std::invalid_argument(
+                "GMRF: Dimension mismatch. Precision matrix rows (" + 
+                std::to_string(this->precision_matrix_ptr->rows()) + 
+                ") must match random effect vector size (" + 
+                std::to_string(n_x) + ").");
+        }
+
+        // Centering: x - mu because TMB's GMRF expects input centered around a mean of 0.
+        vector<Type> x_centered(n_x);
```
nathanvaughan-NOAA: I think this needs to be std::vector

inst/include/interface/rcpp/rcpp_objects/rcpp_growth.hpp
```diff
@@ -285,4 +285,76 @@ class EWAAGrowthInterface : public GrowthInterfaceBase {
...
+ */
```
nathanvaughan-NOAA: We should avoid setting this up to create module specific DSEM classes. Everything should be handled in the dsem interface so it can then point to arbitrary variables in any current/future modules without adding development overhead to them.

inst/include/distributions/functors/precision_builders.hpp
```diff
@@ -0,0 +1,159 @@
...
+template <typename Type>
+struct DSEMPrecisionMatrixBuilder : public PrecisionMatrixBuilderBase<Type> {
+    /**
+     * @brief A small container (struct) to hold the "story" of one arrow.
+     */
+    struct RAMPath {
+        int type = 0;         /**< 1 = A path effect (Rho), 2 = Variance (Gamma) */
+        int from = 0;         /**< The variable the arrow starts from */
+        int to = 0;           /**< The variable the arrow points to */
+        int beta_index = 0;   /**< Which parameter in beta_z to use for this arrow */
```
nathanvaughan-NOAA: Am I right that using the index here is basically so you can map off multiple linkages to be estimated together? I'm not sure how common that would be but it may make more sense just to have the value in RAMPath and allow it to be estimable and/or mapped through TMB like other parameters? That basically replaces beta_index and start with beta_z that can be a fixed or estimated variable.

inst/include/distributions/functors/precision_builders.hpp
```diff
@@ -0,0 +1,159 @@
...
+                const size_t b_idx = static_cast<size_t>(this->paths[r].beta_index - 1);
+                if (b_idx >= this->beta_z.size()) {
+                    throw std::invalid_argument(
+                        "DSEMPrecisionMatrixBuilder: beta_index points past beta_z size.");
+                    }
+                value = this->beta_z[b_idx];
+            }
+
+            // Poke the "strength" into the correct grid slot based on arrow type.
+            if (this->paths[r].type == 1) {
```
nathanvaughan-NOAA: When I looked into SparseMatrix it looks like using .coeffRef is supposedly really slow due to memory moving issues and using Eigen::Triplet is much faster. Something like this apparently 

    int rows = 3;
    int cols = 3;
    Eigen::SparseMatrix<double> A(rows, cols);

    std::vector<Eigen::Triplet<double>> tripletList;
    tripletList.reserve(4); 

    tripletList.push_back(Eigen::Triplet<double>(0, 0, 1.5));
    tripletList.push_back(Eigen::Triplet<double>(1, 2, 2.0));
    tripletList.push_back(Eigen::Triplet<double>(2, 1, -3.5));
    tripletList.push_back(Eigen::Triplet<double>(2, 2, 4.0));

    A.setFromTriplets(tripletList.begin(), tripletList.end());

    A.makeCompressed();

inst/include/distributions/functors/gmrf.hpp
nathanvaughan-NOAA: It would also be faster to have

vector<Type> x_centered;
x_centered.reserve(n_x);
for (size_t i = 0; i < n_x; ++i) {
            x_centered.emplace_back(this->get_observed(i) - this->get_expected(i));
        }

this avoids filling all the x_centered values with default first and overwriting them, emplace also avoids creating and then moving the centered value which push_back does.

inst/include/interface/rcpp/rcpp_objects/wip.cpp
e-perl-NOAA: idk why this and my notes file were included, they shouldn't have been.

inst/include/distributions/functors/precision_builders.hpp
e-perl-NOAA: @nathanvaughan-NOAA When Fernando and I try to implement your suggestion it is giving me the following. Does this look right to you?
```
        // 2. Initialize sparse components
        // We create three empty grids that are "sparse"
        Eigen::SparseMatrix<Type> Rho_kk(static_cast<int>(n_k), static_cast<int>(n_k)); // Grid for causal paths (from -> to)
        Eigen::SparseMatrix<Type> Gamma_kk(static_cast<int>(n_k), static_cast<int>(n_k)); // Grid for variances (<->)
        Eigen::SparseMatrix<Type> I_kk(static_cast<int>(n_k), static_cast<int>(n_k));     // "Standard" grid (identity)
        I_kk.setIdentity(); // Fill diagonal with 1s

       // 3. Translate the "arrows" (RAMPath) into triplet lists for efficient sparse matrix construction.
        std::vector<Eigen::Triplet<Type>> rho_triplets;
        std::vector<Eigen::Triplet<Type>> gamma_triplets;
        rho_triplets.reserve(this->paths.size());
        gamma_triplets.reserve(this->paths.size());
        for (size_t r = 0; r < this->paths.size(); ++r) {
            // C++ starts counting at 0, but R starts at 1, so we subtract 1.
            const int from = this->paths[r].from - 1; 
            const int to = this->paths[r].to - 1;

            // Check if the user's arrow points to a slot that doesn't exist.
            if (from < 0 || to < 0 || static_cast<size_t>(from) >= n_k || 
                static_cast<size_t>(to) >= n_k) {
                throw std::invalid_argument(
                    "DSEMPrecisionMatrixBuilder: RAM indices out of bounds.");
            }

            // Determine the "strength" of this arrow. If beta_index is 1 or 
            // more, it's a parameter the model is guessing, otherwise, it's a fixed number.
            Type value = this->paths[r].start;
            if (this->paths[r].beta_index >= 1) {
                const size_t b_idx = static_cast<size_t>(this->paths[r].beta_index - 1);
                if (b_idx >= this->beta_z.size()) {
                    throw std::invalid_argument(
                        "DSEMPrecisionMatrixBuilder: beta_index points past beta_z size.");
                    }
                value = this->beta_z[b_idx];
            }

            // Add the "strength" to the correct triplet list based on arrow type.
            if (this->paths[r].type == 1) {
                rho_triplets.push_back(Eigen::Triplet<Type>(from, to, value));
            } else if (this->paths[r].type == 2) {
                gamma_triplets.push_back(Eigen::Triplet<Type>(from, to, value));
            }
        }

        Rho_kk.setFromTriplets(rho_triplets.begin(), rho_triplets.end());
        Gamma_kk.setFromTriplets(gamma_triplets.begin(), gamma_triplets.end());
```

inst/include/distributions/functors/gmrf.hpp
e-perl-NOAA: @nathanvaughan-NOAA 
This is what implementing your suggestion gave me, does this look right?
```
        // Centering: x - mu. TMB's GMRF expects input centered around a mean of 0.
        // To improve performance, we build a std::vector using emplace_back to avoid
        // default-constructing and then reassigning every element.
        std::vector<Type> x_centered_std;
        x_centered_std.reserve(n_x);
        for (size_t i = 0; i < n_x; ++i) {
            x_centered_std.emplace_back(this->get_observed(i) - this->get_expected(i));
        }

        // Evaluate TMB GMRF and multiply by -1 to convert from negative log-likelihood to log-likelihood.
        this->lpdf = -1.0 * density::GMRF(*(this->precision_matrix_ptr))(Eigen::Map<const vector<Type>>(x_centered_std.data(), n_x));
```

inst/include/distributions/functors/precision_builders.hpp
nathanvaughan-NOAA: Yep this looks good to me, could probably use emplace_back here too to avoid the move but that is also a wider refactor question throughout FIMS.

inst/include/distributions/functors/gmrf.hpp
nathanvaughan-NOAA: Looks good.

inst/include/distributions/functors/precision_builders.hpp
```diff
@@ -0,0 +1,159 @@
...
+            if (b_idx < 0 || static_cast<size_t>(b_idx) >= this->beta_z.size()) {
+                throw std::invalid_argument(
+                    "DSEMPrecisionMatrixBuilder: beta_index out of bounds for beta_z.");
+            }
+            Type value = this->beta_z[b_idx];
+
+            // Add the "strength" to the correct triplet list based on arrow type.
+            if (this->paths[r].type == 1) {
+                rho_triplets.emplace_back(from, to, value);
+            } else if (this->paths[r].type == 2) {
```
e-perl-NOAA: @nathanvaughan-NOAA and @Andrea-Havron-NOAA 
Fernando is now recommending the following: 
```
   fims::Vector<Eigen::Triplet<Type>> rho_triplets;
   fims::Vector<Eigen::Triplet<Type>> v_triplets;
   rho_triplets.reserve(this->paths.size());
   v_triplets.reserve(this->paths.size() * 2);
...
...
} else if (this->paths[r].type == 2) { // Variance or Covariance path
                if (from == to) { // Variance path: value is a standard deviation
                    v_triplets.emplace_back(from, to, value * value);
                } else { // Covariance path: value is a covariance
                    v_triplets.emplace_back(from, to, value);
                    v_triplets.emplace_back(to, from, value); // Ensure V is symmetric
                }
            }
        }

        Rho_kk.setFromTriplets(rho_triplets.begin(), rho_triplets.end());
        
        Eigen::SparseMatrix<Type> V_kk(static_cast<int>(n_k), static_cast<int>(n_k));
        V_kk.setFromTriplets(v_triplets.begin(), v_triplets.end());
```

and removes

```
 Eigen::SparseMatrix<Type> V_kk = Gamma_kk.transpose() * Gamma_kk;
```

@Andrea-Havron-NOAA does this suggestion tackle what you were talking about yesterday in terms of making sure everything is transposed correctly? This approach would also have me removing gamma from the calculations entirely ... at least in name.

e-perl-NOAA: After some additional back and forth, the following is now being recommended to support covariances:
```
/**
     * @brief The engine that builds the matrix.
     */
    virtual Eigen::SparseMatrix<Type> BuildPrecisionMatrixSparse() const override {
        #ifndef TMB_MODEL
            throw std::invalid_argument(
                "DSEMPrecisionMatrixBuilder: BuildPrecisionMatrixSparse() requires "
                "compilation with TMB_MODEL defined.");
        #else
        // n_k is the total number of "slots" (years multiplied by variables).
        const size_t n_k = this->n_time * this->n_variables;
        
        // 1. Safety check: If the grid is size zero, stop and tell the user.
        if (n_k == 0) {
            throw std::invalid_argument(
                "DSEMPrecisionMatrixBuilder: n_time and n_variables must be > 0.");
        }

        // 2. Initialize Identity matrix
        Eigen::SparseMatrix<Type> I_kk(static_cast<int>(n_k), static_cast<int>(n_k));     // "Standard" grid (identity)
        I_kk.setIdentity(); // Fill diagonal with 1s

       // 3. Translate the "arrows" (RAMPath) into triplet lists for efficient sparse matrix construction.
        fims::Vector<Eigen::Triplet<Type>> rho_triplets;
        fims::Vector<Eigen::Triplet<Type>> vinv_triplets;
        rho_triplets.reserve(this->paths.size());
        vinv_triplets.reserve(this->n_time * this->n_variables * this->n_variables);

        // 4. Build Rho (Directed Paths)
        for (size_t r = 0; r < this->paths.size(); ++r) {
            if (this->paths[r].type == 1) { // Directed path (Rho)
                const int from_global = this->paths[r].from - 1; 
                const int to_global = this->paths[r].to - 1;

                if (from_global < 0 || to_global < 0 || static_cast<size_t>(from_global) >= n_k || static_cast<size_t>(to_global) >= n_k) {
                    throw std::invalid_argument("DSEMPrecisionMatrixBuilder: RAM indices out of bounds.");
                }

                const int b_idx = this->paths[r].beta_index - 1;
                if (b_idx < 0 || static_cast<size_t>(b_idx) >= this->beta_z.size()) {
                    throw std::invalid_argument("DSEMPrecisionMatrixBuilder: beta_index out of bounds for beta_z.");
                }
                
                rho_triplets.emplace_back(from_global, to_global, this->beta_z[b_idx]);
            }
        }
        
        // 5. Build V_inv via Block Inversion
        // This assumes global indices are generated as: 
        // global_index = (time_step * n_variables) + variable_index.
        // If your R data is vectorized column-major (time inside, variables outside),
        // adjust the division and modulo math below accordingly.
        for (size_t t = 0; t < this->n_time; ++t) {
            
            Eigen::Matrix<Type, Eigen::Dynamic, Eigen::Dynamic> V_block = 
                Eigen::Matrix<Type, Eigen::Dynamic, Eigen::Dynamic>::Zero(this->n_variables, this->n_variables);
            
            bool has_variance_for_block = false;

            for (size_t r = 0; r < this->paths.size(); ++r) {
                if (this->paths[r].type == 2) { 
                    int from_global = this->paths[r].from - 1; 
                    int to_global = this->paths[r].to - 1;
                    
                    size_t path_t_from = from_global / this->n_variables;
                    size_t path_t_to = to_global / this->n_variables;

                    // Safety check: Block inversion cannot handle cross-year covariances
                    if (path_t_from != path_t_to) {
                        throw std::runtime_error(
                            "DSEMPrecisionMatrixBuilder: Cross-year covariances (Type 2 paths "
                            "between different time steps) are not supported with block inversion.");
                    }

                    if (path_t_from == t) {
                        has_variance_for_block = true;
                        int from_local = from_global % this->n_variables;
                        int to_local = to_global % this->n_variables;
                        
                        const int b_idx = this->paths[r].beta_index - 1;
                        Type value = this->beta_z[b_idx];

                        if (from_local == to_local) {
                            V_block(from_local, to_local) = value * value; // Variance 
                        } else {
                            V_block(from_local, to_local) = value;         // Covariance
                            V_block(to_local, from_local) = value;         // Enforce Symmetry
                        }
                    }
                }
            }
            
            // Invert the dense block and map it back to the global sparse format
            if (has_variance_for_block) {
                Eigen::Matrix<Type, Eigen::Dynamic, Eigen::Dynamic> Vinv_block = V_block.inverse();
                
                for (size_t i = 0; i < this->n_variables; ++i) {
                    for (size_t j = 0; j < this->n_variables; ++j) {
                        if (Vinv_block(i, j) != Type(0.0)) {
                            int global_i = (t * this->n_variables) + i;
                            int global_j = (t * this->n_variables) + j;
                            vinv_triplets.emplace_back(global_i, global_j, Vinv_block(i, j));
                        }
                    }
                }
            }
        }

        // 6. Final Sparse Matrix Assembly
        Eigen::SparseMatrix<Type> Rho_kk(static_cast<int>(n_k), static_cast<int>(n_k));
        Rho_kk.setFromTriplets(rho_triplets.begin(), rho_triplets.end());

        Eigen::SparseMatrix<Type> Vinv_sparse(static_cast<int>(n_k), static_cast<int>(n_k));
        Vinv_sparse.setFromTriplets(vinv_triplets.begin(), vinv_triplets.end());

        // Calculate Q = (I - P)^T * V^-1 * (I - P)
        Eigen::SparseMatrix<Type> IminusRho_kk = I_kk - Rho_kk;
        
        return IminusRho_kk.transpose() * Vinv_sparse * IminusRho_kk;
#endif
    }
```

e-perl-NOAA:
@Andrea-Havron-NOAA Also, gemini is convinced that the following lines that we added yesterday in the rcpp_precision_builders.hpp need to be removed:
```
std::shared_ptr<fims_distributions::GMRF<Type>> gmrf = 
            std::make_shared<fims_distributions::GMRF<Type>>();    

        gmrf->precision_matrix_ptr = builder; 
```
because of the following:
**Incorrect Logic:** The DSEMInterface's responsibility is to create and configure the DSEMPrecisionMatrixBuilder, not the GMRF distribution itself. The gmrf object created here is a local variable that is immediately destroyed, serving no purpose. The builder is correctly registered in the Information map to be used by a GMRF component later.
**Type Mismatch:** The assignment gmrf->precision_matrix_ptr = builder; would cause a compilation error. gmrf->precision_matrix_ptr is a std::shared_ptr<Eigen::SparseMatrix<Type>> (the matrix), but builder is a std::shared_ptr<fims_distributions::DSEMPrecisionMatrixBuilder<Type>> (the object that builds the matrix).

Does this seem right?

kellijohnson-NOAA:
@e-perl-NOAA should we mark this as a draft PR until you are ready for a final review? Or, is there something you currently need help with?

e-perl-NOAA:
@kellijohnson-NOAA I'm still waiting on some feedback from @Andrea-Havron-NOAA and/or @nathanvaughan-NOAA after my last two posts in the thread that contained some questions.

nathanvaughan-NOAA:
Andrea is the expert on the GMRF stats but I’m good with explicitly filling out the symmetrical matrix for clarity.

Andrea-Havron-NOAA:
> @Andrea-Havron-NOAA Also, gemini is convinced that the following lines that we added yesterday in the rcpp_precision_builders.hpp need to be removed:
> 
> ```
> std::shared_ptr<fims_distributions::GMRF<Type>> gmrf = 
>             std::make_shared<fims_distributions::GMRF<Type>>();    
> 
>         gmrf->precision_matrix_ptr = builder; 
> ```
> 
> because of the following: **Incorrect Logic:** The DSEMInterface's responsibility is to create and configure the DSEMPrecisionMatrixBuilder, not the GMRF distribution itself. The gmrf object created here is a local variable that is immediately destroyed, serving no purpose. The builder is correctly registered in the Information map to be used by a GMRF component later. **Type Mismatch:** The assignment gmrf->precision_matrix_ptr = builder; would cause a compilation error. gmrf->precision_matrix_ptr is a std::shared_ptr<Eigen::SparseMatrix> (the matrix), but builder is a std::shared_ptr<fims_distributions::DSEMPrecisionMatrixBuilder> (the object that builds the matrix).
> 
> Does this seem right?

This is a bit philosophical. The GMRF is a distribution that requires a precision matrix. Rather than initialize the GMRF and its precision matrix independently and add a lot of checks to make sure the user set up both correctly and linked them together correctly, I think they should be linked internally. I decided that the user specifies the precision matrix (in this case, dsem), and by default, this initializes GMRF. We could rename the interface to dsemGMRFInterface to be more explicit about what is happening under the hood. 

Given this, there are a couple of pieces missing as Gemini is right in that the gmrf is created locally and then destroyed. We can make the gmrf object persist by adding this code just below ` gmrf->precision_matrix_ptr = builder; `:

```
gmrf->distribution_type = "random_effects";
info->density_components[gmrf->id] = gmrf;
```

Also `precision_builder_ptr` Type in gmrf.hpp should be updated to:
`std::shared_ptr<fims_distributions::PrecisionMatrixBuilderBase<Type>> precision_builder_ptr = nullptr;`

And then in the gmrf.hpp file, add:
`Eigen::SparseMatrix<Type> Q = precision_builder_ptr->BuildPrecisionMatrixSparse();`
and replace `*(this->precision_matrix_ptr)` in the GMRF with Q: `this->lpdf = -1.0 * density::GMRF(Q)(Eigen::Map<const vector<Type>>(x_centered_std.data(), n_x));`

There is still a missing piece, we will need to link the distribution's `observed_values` with the correct parameters. We will probably need to build this into the interface, I need to see what the R output from running the R dsem model looks like so I can think about this a bit more. I think this update can be worked on in the next PR though.

I did want to mention, another design direction would be to create a gmrf interface which creates the DSEM builder instead given a precision type input from the user. I guess my rational for thinking about it the other way is that the parameters are associated with the precision matrix, so it is easier to set them up for registration if the primary class being initialized is the precision matrix which then calls the GMRF rather than the other way around.  

I will look over the suggested changes to precision_builders.hpp tomorrow!

inst/include/distributions/functors/precision_builders.hpp
Andrea-Havron-NOAA: I had AI walk me through what this code above is doing. Apparently, it is a very efficient calculation of Q, but requires the assumption that there is no cross-year covariance (what a moving average assumes, for example, using a moving average model to handle a heat wave that started in the fall of one year and ended in the spring of the next year). Thorson set up his dsem package to be able to handle cross-year correlation, so we should probably also be able to handle it. The recommendation to account for this is to have both approaches for calculating Q. The efficient approach is the default and the less efficient approach is only evaluated when there are cross-year correlations.:

```
bool has_cross_year_cov = false;

// Pre-scan paths to determine the optimal inversion strategy
for (size_t r = 0; r < this->paths.size(); ++r) {
    if (this->paths[r].type == 2) {
        size_t path_t_from = (this->paths[r].from - 1) / this->n_variables;
        size_t path_t_to = (this->paths[r].to - 1) / this->n_variables;
        
        if (path_t_from != path_t_to) {
            has_cross_year_cov = true;
            break; // Stop scanning as soon as we find one
        }
    }
}

Eigen::SparseMatrix<Type> BuildPrecisionMatrixSparse() const override {
    // ... [Initial setup and pre-scan] ...

    if (has_cross_year_cov) {
        // 🔹 Fallback: Slower, but mathematically robust for complex time linkages
        return BuildQ_FullInversion(); 
    } else {
        // ⚡ Fast Path: Highly optimized block-diagonal inversion
        return BuildQ_BlockInversion(); 
    }
}
```

I also noticed in Grant's DSEM code, that x is additionally centered with a `delta_tj`:
```
 jnll_gmrf = GMRF(Q_kk)( x_tj - xhat_tj - delta_tj );
```
The code to calculate delta_tj is here and is based on an estimated parameter, `delta0_j`: 
```
// Calculate effect of initial condition -- SPARSE version
  vector<Type> delta_k( n_k );
  delta_k.setZero();
  if( delta0_j.size() > 0 ){
    // Compute delta_k
    matrix<Type> delta0_k1( n_k, 1 );
    delta0_k1.setZero();
    for(int j=0; j<n_j; j++){
      k = j * n_t;
      delta0_k1(k,0) = delta0_j(j);
    }

    // SPARSE version
    // See C:\Users\James.Thorson\Desktop\Work files\AFSC\2023-06 -- Sparse inverse-product\Kasper example\lu.cpp
    matrix<Type> x = inverseIminusRho_kk.solve(delta0_k1);

    //REPORT( delta0_k1 );
    delta_k = x.array();
  }
  //REPORT( delta_k );

  // Format mu_j
  array<Type> xhat_tj( n_t, n_j );
  array<Type> delta_tj( n_t, n_j );
  for(int j=0; j<n_j; j++){
    for(int t=0; t<n_t; t++){
      k = j*n_t + t;
      xhat_tj(t,j) = mu_j(j);
      delta_tj(t,j) = delta_k(k);
    }}
```

but copilot says this will be faster:
```
// Initialize the offset matrix (n_time x n_variables)
Eigen::Matrix<Type, Eigen::Dynamic, Eigen::Dynamic> delta_tj(this->n_time, this->n_variables);
delta_tj.setZero();

// We only need to do this if initial conditions are actually provided
if (this->delta0_j.size() > 0) {
    
    // Loop forward through time
    for (size_t t = 0; t < this->n_time; ++t) {
        
        // 1. Gather incoming effects from previous time steps
        Eigen::Matrix<Type, Eigen::Dynamic, 1> incoming_delta = 
            Eigen::Matrix<Type, Eigen::Dynamic, 1>::Zero(this->n_variables);
        
        // At t=0, the "incoming" effect is just the user-provided initial condition
        if (t == 0) {
            for (size_t j = 0; j < this->n_variables; ++j) {
                incoming_delta(j) = this->delta0_j(j);
            }
        }
        
        // Add effects propagating from PAST time steps
        for (size_t r = 0; r < this->paths.size(); ++r) {
            if (this->paths[r].type == 1) { // Type 1 = Causal Path (P matrix)
                size_t t_from = (this->paths[r].from - 1) / this->n_variables;
                size_t t_to   = (this->paths[r].to - 1) / this->n_variables;
                
                // If path arrives at current year 't' FROM a previous year
                if (t_to == t && t_from < t) {
                    size_t j_from = (this->paths[r].from - 1) % this->n_variables;
                    size_t j_to   = (this->paths[r].to - 1) % this->n_variables;
                    
                    // Note: You may need to handle fixed vs estimated beta_z here
                    Type path_val = this->beta_z[this->paths[r].beta_index - 1]; 
                    
                    // P * delta_{past}
                    incoming_delta(j_to) += path_val * delta_tj(t_from, j_from);
                }
            }
        }
        
        // 2. Gather WITHIN-YEAR causal paths (creates a local I - P matrix)
        Eigen::Matrix<Type, Eigen::Dynamic, Eigen::Dynamic> I_minus_P_local = 
            Eigen::Matrix<Type, Eigen::Dynamic, Eigen::Dynamic>::Identity(this->n_variables, this->n_variables);
            
        bool has_local_paths = false;
        for (size_t r = 0; r < this->paths.size(); ++r) {
            if (this->paths[r].type == 1) {
                size_t t_from = (this->paths[r].from - 1) / this->n_variables;
                size_t t_to   = (this->paths[r].to - 1) / this->n_variables;
                
                // If path is entirely within the current year 't'
                if (t_to == t && t_from == t) {
                    size_t j_from = (this->paths[r].from - 1) % this->n_variables;
                    size_t j_to   = (this->paths[r].to - 1) % this->n_variables;
                    
                    Type path_val = this->beta_z[this->paths[r].beta_index - 1];
                    I_minus_P_local(j_to, j_from) -= path_val; // Subtract because it's (I - P)
                    has_local_paths = true;
                }
            }
        }
        
        // 3. Solve for this year's final delta
        if (has_local_paths) {
            // If there are simultaneous effects (e.g. j1 -> j2 instantly), solve the local block
            // (I - P_local) * delta_t = incoming_delta
            delta_tj.row(t) = I_minus_P_local.colPivHouseholderQr().solve(incoming_delta).transpose();
        } else {
            // Most common case: no instantaneous within-year paths
            delta_tj.row(t) = incoming_delta.transpose();
        }
    }
}
```

Then you will need to add a new GetMeanOffset function to the base class (where the offset is zero by default):
```
template <typename Type>
struct PrecisionMatrixBuilderBase {
    virtual ~PrecisionMatrixBuilderBase() {}
    
    // Builds the precision matrix Q
    virtual Eigen::SparseMatrix<Type> BuildPrecisionMatrixSparse() const = 0;

    // NEW: Optional mean offset. Defaults to a zero vector.
    virtual Eigen::Matrix<Type, Eigen::Dynamic, 1> GetMeanOffset(size_t n_k) const {
        return Eigen::Matrix<Type, Eigen::Dynamic, 1>::Zero(n_k);
    }
};
```

and then override in the DSEM Builder to compute the DSEM specific initial condition effect

```
virtual Eigen::Matrix<Type, Eigen::Dynamic, 1> GetMeanOffset(size_t n_k) const override {
    if (this->delta0_j.size() == 0) {
        return Eigen::Matrix<Type, Eigen::Dynamic, 1>::Zero(n_k);
    }
    // ... logic to compute delta = (I - P)^-1 * delta_0 ...
    return computed_delta_vector;
}
```

Now the GMRF is clean of the DSEM specific code:
```
Eigen::SparseMatrix<Type> Q = builder->BuildPrecisionMatrixSparse();
Eigen::Matrix<Type, Eigen::Dynamic, 1> offset = builder->GetMeanOffset(n_k);

// Fully generic centering
Eigen::Matrix<Type, Eigen::Dynamic, 1> residual = x - xhat - offset;

// Evaluate density
return GMRF(Q)(residual);
```

Andrea-Havron-NOAA: I also wanted to note next steps for future PRs that I noticed while looking over the code:


Map beta_z to log_sd:
Link the correct index from the beta_z vector (the Type 2 variance path for recruitment) to the log_sd parameter in your existing recruitment distribution module. This ensures the rest of your assessment's penalty functions know what the recruitment standard deviation is.

Add a Safety Catch for Marginal Variance:
Add a quick check in your matrix builder to throw an explicit std::runtime_error if a user passes a model with marginal variance enabled (options(1) > 0). This protects users from silent mathematical failures until you implement the feature fully.

Long term:
Implement Marginal Variance Scaling:
Add the LU-decomposition and matrix scaling logic to your matrix builder to fully support dsem's marginal variance options. This will be necessary if users want to apply life-history priors to $\sigma_R$ in heavily autocorrelated models.

inst/include/distributions/functors/gmrf.hpp
```diff
@@ -32,7 +32,11 @@ struct GMRF : public DensityComponentBase<Type> {
      * @details The matrix is assembled by a separate precision 
      * builder (e.g., the DSEM builder).
      */
-    std::shared_ptr<Eigen::SparseMatrix<Type>> precision_matrix_ptr = nullptr;
+
+    std::shared_ptr<fims_distributions::PrecisionMatrixBuilderBase<Type>> precision_matrix_ptr = nullptr;
+
+    // IS THIS WHERE THIS GOES?
+    Eigen::SparseMatrix<Type> Q = precision_matrix_ptr->BuildPrecisionMatrixSparse();
```
Andrea-Havron-NOAA: This needs to be inside the GMRF evaluate function

inst/include/common/fims_vector.hpp
```diff
@@ -354,7 +354,12 @@ class Vector {
   /**
    * @brief Adds an element to the end.
    */
-  inline void push_back(const Type &&value) { this->vec_m.push_back(value); }
+  inline void push_back(const Type &value) { this->vec_m.push_back(value); }
+
+  /**
+   * @brief Adds an element to the end using move semantics.
+   */
+  inline void push_back(Type &&value) { this->vec_m.push_back(std::move(value)); }
```
Andrea-Havron-NOAA, line 379: gemini's recommendation for updating the documentation for the push_back on line 357:
```
/**
 * @brief Adds a copy of an element to the end of the vector.
 * 
 * @details This function takes a read-only reference to an object and 
 * safely copies its data into the internal container. 
 * 
 * @param value The element to be copied and appended.
 */
```

and on line 362: 
```
/**
 * @brief Moves an element to the end of the vector.
 * 
 * @details This function takes a temporary object (rvalue) and transfers 
 * its internal data directly into the container to avoid an expensive copy. 
 * Note: The original `value` object is left in a valid but unspecified state 
 * after this operation.
 * 
 * @param value The temporary element whose contents will be moved and appended.
 */
```

inst/include/common/information.hpp
```diff
@@ -350,12 +363,30 @@ class Information {
                       fims::to_string(d->key[0]));
         vmit = this->variable_map.find(d->key[0]);
         d->re = (*vmit).second;
-        if (d->key.size() == 2) {
+
+        std::shared_ptr<fims_distributions::GMRF<Type>> gmrf =
```
Andrea-Havron-NOAA: gmrf shouldn't be explicitly specified here. If it is added to `info->density_components` in the interface, it will be one of the objects in the loop. What we need to add for this to work is a pointer to uncertainty in DensityComponentsBase, Then this uncertainty pointer can point to the gmrf->precision_builders. 

Need:
- uncertainty pointer in DensityComponentsBase (`fims::Vector<Type>* uncertainty = NULL;`) - also add to clear in information
- a third argument in GMRFDistributionsInterface for set_distribution_links that passes in the precision_builder_id_m to the key_m vector
- update SetupRandomEffects:
- The key vector in DensityComponentsBase needs to be refactored as a Struct:
```
struct DistributionKey {
    uint32_t observed_id;
    uint32_t expected_id static_cast<uint32_t>(-999);
    uint32_t uncertainty_id static_cast<uint32_t>(-999);
};
```
and in DensityComponentsBase:
`DistributionKey key;`

Then `SetupRandomEffects` can be refactored to:
```
void SetupRandomEffects() {
  for (density_components_iterator it = this->density_components.begin();
       it != this->density_components.end(); ++it) {
    std::shared_ptr<fims_distributions::DensityComponentBase<Type>> d = (*it).second;

    if (d->input_type == "random_effects") {
      FIMS_INFO_LOG("Setup random effects for distribution " + fims::to_string(d->id));
      variable_map_iterator vmit;

      // 1. Link the observed random effect vector
      FIMS_INFO_LOG("Link random effects from distribution " + fims::to_string(d->id) +
                    " to derived value " + fims::to_string(d->key.observed_id));
      vmit = this->variable_map.find(d->key.observed_id);
      if (vmit != this->variable_map.end()) {
        d->re = (*vmit).second;
      } else {
        FIMS_ERROR_LOG("Observed ID not found in variable map for distribution " + fims::to_string(d->id));
      }

      // 2. Link or initialize the expected values
      if (d->key.expected_id.has_value()) {
        vmit = this->variable_map.find(d->key.expected_id.value());
        if (vmit != this->variable_map.end()) {
          d->re_expected_values = (*vmit).second;
        } else {
          FIMS_ERROR_LOG("Expected ID not found in variable map for distribution " + fims::to_string(d->id));
        }
      } else {
        // Fallback if no expected ID is provided (maintains the previous GMRF resizing behavior)
        if (d->re) {
          d->expected_values.resize(d->re->size());
        }
        d->re_expected_values = &d->expected_values;
      }

      // 3. Link the uncertainty (this now generically handles the GMRF precision builder pointer)
      if (d->key.uncertainty_id.has_value()) {
        vmit = this->variable_map.find(d->key.uncertainty_id.value());
        if (vmit != this->variable_map.end()) {
          d->re_uncertainty_values = (*vmit).second; 
        } else {
          FIMS_ERROR_LOG("Uncertainty ID not found in variable map for distribution " + fims::to_string(d->id));
        }
      }

      FIMS_INFO_LOG("Random effect size for distribution " + fims::to_string(d->id) +
                    " is: " + fims::to_string(d->observed_values.size()));
    }
  }
}
```

We might want to pull out this refactor (uncertainty + refactoring key to struct) and bring into main sooner than later for @nathanvaughan-NOAA work

inst/include/common/model_object.hpp
```diff
@@ -60,6 +60,14 @@ struct FIMSObject : public FIMSMemoryTracker {
   std::vector<Type*>
       fixed_effects_parameters; /**< list of fixed effects parameters */
 
+  /**
+     * @brief Vector of path coefficients for DSEM-based modeling.
+     * @details This allows any module inheriting from FIMSObject (Recruitment, 
+     * Growth, Selectivity, Maturity) to serve as an origin or target in a 
+     * structural equation model.
+     */
+    fims::Vector<Type> beta_z;
```
Andrea-Havron-NOAA: I don't think we need to define `beta_z` anywhere. We link beta_z to each module via the `re` pointer in `DensityComponentsBase` 
Steps:
1. `set_distribution_links` sets the expected values key (e.g. id to `log_expected_recruitment` which is already added to the variable map in the recruitment interface)
2. `SetupRandomEffects` links the gmrf expected_values to `log_expected_recruitment` via the `re_expected_values` pointer
3. The GMRF evaluate function calls `this->get_expected(i)`, which retrieves the ith value from `log_expected_recruitment`

* More machinery will be needed to handle the case where `beta_z` points to expected values in more than one module

inst/include/distributions/distributions.hpp
```diff
@@ -11,9 +11,11 @@
 #ifndef FIMS_DISTRIBUTIONS_HPP
 #define FIMS_DISTRIBUTIONS_HPP
 
-#include "functors/density_components_base.hpp"
-#include "functors/lognormal_lpdf.hpp"
-#include "functors/multinomial_lpmf.hpp"
-#include "functors/normal_lpdf.hpp"
+#include "distributions/functors/density_components_base.hpp"
```
Andrea-Havron-NOAA: I'm not sure why this needed to change. Because both `distributions.hpp` and the functors folder are in the distributions folder, this file should look inside the distributions folder for includes.

inst/include/interface/rcpp/rcpp_objects/rcpp_distribution.hpp
```diff
@@ -848,6 +850,139 @@ class DlnormDistributionsInterface : public DistributionsInterfaceBase {
...
+  GMRFDistributionsInterface(const GMRFDistributionsInterface &other)
+      : DistributionsInterfaceBase(other),
+        precision_builder_id_m(other.precision_builder_id_m),
+        lpdf_vec(other.lpdf_vec) {}
+
+  virtual ~GMRFDistributionsInterface() {}
+
+  virtual uint32_t get_id() { return this->id_m; }
+
+  virtual bool set_distribution_links(std::string input_type,
```
Andrea-Havron-NOAA: refactor so this function takes an observation id, expected_values id, and uncertainty_id and remove precision_builder specific code

inst/include/interface/rcpp/rcpp_objects/rcpp_distribution.hpp
```diff
@@ -848,6 +850,139 @@ class DlnormDistributionsInterface : public DistributionsInterfaceBase {
...
+    this->key_m->clear();
+    this->key_m->reserve(ids.size() + 1);
+    for (R_xlen_t i = 0; i < ids.size(); i++) {
+      this->key_m->push_back(ids[i]);
+    }
+    this->key_m->push_back(this->precision_builder_id_m.get());
+    return true;
+  }
+
+  bool set_precision_builder_id(int precision_builder_id) {
```
Andrea-Havron-NOAA: remove this function

e-perl-NOAA:
@Andrea-Havron-NOAA after our recent changes it seems like there is an issue with the rcpp_distribution.hpp file and copilot said the following in response to the failed r-cmd-check:

What’s breaking
The logs show two concrete problems:

DistributionKey is not declared

Fails in DistributionsInterfaceBase constructor:
this->key_m = std::make_shared<DistributionKey>();
But the type being used elsewhere looks like key_m is intended to hold a container of IDs, not a DistributionKey type directly, and the compiler cannot find that name in scope.
key is treated like a vector, but DistributionKey has no resize()

In TMB export code:
distribution->key.resize(this->key_m->size());
distribution->key is a vector-like member on the TMB distribution object.
this->key_m is a pointer to DistributionKey, which appears inconsistent with the usage in this file.


Gemini recommended the following changes:
- In `rcpp_distribution.hpp` change `std::shared_ptr<DistributionKey> key_m;` to `std::shared_ptr<fims_distributions::DistributionKey> key_m;`
- In `information.hpp` change `if (d->key.expected_id.has_value()) {
          vmit = this->variable_map.find(d->key.expected_id.value());` to `if (!d->key.expected_id.empty()) {
          vmit = this->variable_map.find(d->key.expected_id[0]);` and `if (d->key.uncertainty_id.has_value()) {
          vmit = this->variable_map.find(d->key.uncertainty_id.value());` to `if (d->key.uncertainty_id != static_cast<uint32_t>(-999)) {vmit = this->variable_map.find(d->key.uncertainty_id);`

Does any or all of this seem logical to you?

Andrea-Havron-NOAA:
example code for connecting the GMRF and precision builder in R:

```
gmrf = methods::new(GMRFDistributionsInterface)
dsem = method::new(DSEMInterface)
dsem$ram_matrix = # ram output from R's dsem code
dsem$beta_z = # beta_z from R's dsem code

gmrf$set_distribution_links("random_effects", observed_id = c("ids from module parameters"), uncertainty_id = dsem$get_id()) 

```

e-perl-NOAA:
@Andrea-Havron-NOAA I've used up all my copilot credits for the month and gemini isn't as helpful anymore with the errors from the two github actions that are failing. However, it *seems* like this almost is there?!?

kellijohnson-NOAA:
@e-perl-NOAA I am only at 377 of 3,000 if you want to co-work on it next week to use mine just set up a calendar invite.

e-perl-NOAA:
@kellijohnson-NOAA That would actually be great! I was at 63% usage this morning and idk how in just several hours it jumped up to 100% especially when I only used 7% yesterday doing the same thing. I think it might be that you need to start a new chat every once in awhile to erase too much context that it reads every time and make sure to close out of as many files as possible.

Andrea-Havron-NOAA:
There are a couple of changes that might help with compile time. Since `<Eigen/Core>` and `<Eigen/Sparse>` are already included with TMB, you shouldn't need to include them again when TMB is defined. Wrap these with:

```
#ifndef TMB_MODEL
#include <Eigen/Core>
#include <Eigen/Sparse>
#endif
```

It doesn't look like you are using `Eigen/SparseLU` anywhere, I recommend removing this include. 

See if the model still compiles with these changes. Hopefully 🤞 the model compiles faster!

e-perl-NOAA:
@Andrea-Havron-NOAA when I tried to do that it threw me errors such as:
```
 44 | virtual Eigen::SparseMatrix<Type> BuildPrecisionMatrixSparse() const = 0; | ^~~~~ /home/runner/work/FIMS/FIMS/inst/include/population_dynamics/population/../fleet/../../distributions/functors/precision_builders.hpp:53:13: error: ‘Eigen’ does not name a type 53 | virtual Eigen::Matrix<Type, Eigen::Dynamic, 1> Project( | ^~~~~ 
/home/runner/work/FIMS/FIMS/inst/include/population_dynamics/population/../fleet/../../distributions/functors/precision_builders.hpp:66:13: error: ‘Eigen’ does not name a type 66 | virtual Eigen::Matrix<Type, Eigen::Dynamic, 1> GetMeanOffset(size_t n_k) const { | ^~~~~ 
/home/runner/work/FIMS/FIMS/inst/include/population_dynamics/population/../fleet/../../distributions/functors/precision_builders.hpp:91:13: error: ‘Eigen’ does not name a type 91 | virtual Eigen::SparseMatrix<Type> BuildPrecisionMatrixSparse() const override { | ^~~~~
```

Issue #1601, opened 2026-07-02
Input parameter tibble should match the output tibble to the best extent possible


kellijohnson-NOAA:
I think it would be 🥇 if we can get the input parameter tibble to match the output tibble to the best extent possible.
- [ ] module_name entries in output tibble should match the case of the module and entries in the input tibble rather than be all lower case. This change will help the output match the codebase for easier searching.
- [ ] Double check the Growth module, should module_name be Growth and then label is EWAA? Rather than EWAAGrowth?
- [ ] Remove "model_family" from output
- [x] Change "name" to "fleet" in FIMSFrame
- [x] Change "fleet_name" to "fleet" in output
- [x] Change "age_i" to "age" with appropriate values in output
- [ ] Change "length_i" to "length" with the appropriate values in output
- [x] Change "year_i" to "timing" with appropriate values in output
- [x] Change "time" to "timing" in input
- [ ] There is a mismatch between "value" in the parameter tibble and "input" in the output from get_estimates(), where value matches the C++. Need to check with Sam regarding the naming of the output. input_value and estimated_value are more clear but take up more room.

Below is a table of labels by module_name from an example output.
```
dplyr::count(out, module_name, label) |> print(n = 900)
# A tibble: 45 × 3
   module_name label                          n
   <chr>       <chr>                      <int>
 1 fleet       age_to_length_conversion     552
 2 fleet       agecomp_expected             720
 3 fleet       agecomp_proportion           720
 4 fleet       catch_index                   60
 5 fleet       index_expected                60
 6 fleet       index_numbers                 60
 7 fleet       index_numbers_at_age         720
 8 fleet       index_numbers_at_length     1380
 9 fleet       index_weight                  60
10 fleet       index_weight_at_age          720
11 fleet       landings_expected             60
12 fleet       landings_numbers              60
13 fleet       landings_numbers_at_age      720
14 fleet       landings_numbers_at_length  1380
15 fleet       landings_weight               60
16 fleet       landings_weight_at_age       720
17 fleet       lengthcomp_expected         1380
18 fleet       lengthcomp_proportion       1380
19 fleet       log_Fmort                     60
20 fleet       log_index_expected            60
21 fleet       log_landings_expected         60
22 fleet       log_q                          2
23 growth      EWAA                          12
24 maturity    inflection_point               1
25 maturity    slope                          1
26 population  biomass                       31
27 population  expected_recruitment          31
28 population  log_M                        360
[...]

PR #1608, opened 2026-07-08
Fix name error with dplyr

kellijohnson-NOAA:
# What is the feature?
* When we changed to $column_name or [["column_name"]] we did not realize that we could not use .data$name where we had a column in FIMSFrame of "name". This was changed to "fleet" in this PR. We also changed the output to be "fleet" instead of "fleet_name" so they match.

* Use .data$fleet instead of .data$name and .data$fleet_name.
* Uses dplyr::all_of() and dplyr::across(dplyr::all_of()) in non-select functions to avoid the warning from dplyr.
* Couldn't use devtools::document() on windows with the non UTF-8 characters, so that was fixed as well
* Removes {scales} from DESCRIPTION file because it is not used

* [`r4ss::ss3_data_to_fims()`](https://github.com/r4ss/r4ss/blob/main/R/ss3_data_to_fims.R) will need fixed
* Case studies will need fixed

Note that the 1 to many warning still exists.
Part of #1601

___

## Checklist


R/fimsframe.R
```diff
@@ -1021,17 +1023,17 @@ FIMSFrame <- function(data) {
...
+    summary_by_name <- dplyr::count(missing_ages, .data$fleet, .data$timing) |>
       dplyr::filter(.data$n != n_ages) |>
       dplyr::summarize(
         timings = paste(.data$timing, collapse = ", "),
-        .by = .data$name
+        .by = dplyr::all_of("fleet")
       )
     if (NROW(summary_by_name) > 0) {
       cli::cli_abort(
         "You cannot have missing age values for a given timing and name
```
Bai-Li-NOAA: change to "giving time and fleet combination"

R/fimsframe.R
```diff
@@ -1046,17 +1048,21 @@ FIMSFrame <- function(data) {
...
+    ) |>
       dplyr::filter(.data$n != n_lengths) |>
       dplyr::summarize(
         timings = paste(.data$timing, collapse = ", "),
-        .by = .data$name
+        .by = dplyr::all_of("fleet")
       )
     if (NROW(summary_by_name) > 0) {
       cli::cli_abort(
         "You cannot have missing length values for a given timing and name
```
Bai-Li-NOAA: change to "giving time and fleet combination"

R/fimsframe.R
```diff
@@ -285,7 +285,7 @@ methods::setMethod(
 #' @inheritParams get_data
 #' @param fleet_name A string, or vector of strings, specifying the name of the
```
Bai-Li-NOAA: Not sure if "fleet_name" needs to be "fleet" here.

Bai-Li-NOAA (commented):
Everything looks great to me! I just left a few comments on a couple of areas that might have been missed.

Bai-Li-NOAA (approved):

R/fimsframe.R
kellijohnson-NOAA: done

R/fimsframe.R
kellijohnson-NOAA: done

R/fimsframe.R
kellijohnson-NOAA: Changed, and changed it within all of the accessor functions.

PR #1653, opened 2026-07-22
fix(initialize_fims): log_sd needed filtered by module_type

### LINKED ISSUE #1649: `initialize_fims()` does not work with landings and index data assigned to one fleet

Working with opakapaka in Code Club today, where they are trying to have catch and index data for a single fleet, we ran into errors when running `initialize_fims()` because of setting sd was not the right length. We are NOT filtering by type right now and we need to because index and landings were both using lognormal.

### Additional context

```
    fleet_sd_input <- parameters |>
      dplyr::filter(.data$fleet == .env$fleets[i] & .data$label == "log_sd") |>
      dplyr::mutate(
        label = "sd",
        value = exp(.data$value)
      )
```
in R/initialize_modules, line 1047 is fine but it needs to be filtered for type within each use. This should be fixed in the work that @kellijohnson-NOAA is doing with @Bai-Li-NOAA on the wrapper input functions 🎉.

_No response_

_No response_

kellijohnson-NOAA:
# What is the feature?
* Ensure log_sd is for the correct type
* Bug was found working with opakapaka where there is a fleet with catch and index data
* Removed the warning about the value being too long because this information is actually being changed in a future PR where the sd comes from the data rather than the parameter tibble.

* Pull the parameters for each module rather than pulling the information and then using the same pull for landings and index data. Use dplyr::filter on the module_type

* @awilnoaa once this is in main if you can rebase the growth branch it would be really helpful for Meg and Ian who are going to use the growth branch from now on for the opakapaka case study rather than main.

@MOshima-PIFSC and @iantaylor-NOAA can one of you please check that the fix works for opakapaka? I am not too terribly worried about writing tests for this or the removal of the test because the logic for where this information comes from is changing with a PR that Bai and I have in works. I wanted to get this fix into main though so Adrianne can rebase.

___

## Checklist

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: FIMS
 Title: The Fisheries Integrated Modeling System
-Version: 0.9.3.9000
+Version: 0.9.4
 Authors@R: c(
     person(c("Kelli", "F."), "Johnson", , "kelli.johnson@noaa.gov", role = c("aut", "cre"),
            comment = c(ORCID = "0000-0002-5149-451X")),
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,17 @@
+# FIMS 0.9.4
+
+* Adds likelihood profiles and retrospective diagnostic tools
+* Adds priors on derived quantities
+* Adds tidy output with yardstick
+* Uses .data and .env for dplyr instead of globalVariable()
+* Uses translational units to decrease RAM during compile
+* Expose proportion_female as a scalar that can be set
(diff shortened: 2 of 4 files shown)


MOshima-PIFSC:
This worked for me, I no longer got the error about mismatch in log_sd vector.

MOshima-PIFSC (approved):
Looks good to me!

iantaylor-NOAA (approved):
Changes look good to me. I was going to install and test, but since @MOshima-PIFSC has already done so I think you should just merge.

PR #1678, opened 2026-07-31
Uses distribution in data to set up uncertainty

### LINKED ISSUE #1438: [Refactor]: Centralize hardcoded data-type name strings into a package-level object
### Problem

Data-type name strings (`"landings"`, `"index"`, `"age_comp"`, `"length_comp"`) are hardcoded inline in multiple places, e.g., `distribution_formulas.R:33` has a FIXME asking for this exact change. #1247 standardised the naming but the centralisation was never done.

If a new data type is added or a name changes, every call site has to be found and updated manually.

### Proposal

Define a single authoritative vector (e.g., `fims_data_types`) as internal package data or a package-level constant, and replace all inline occurrences with references to it.

### Related issues

- Follow-up to #1247 (naming was standardized, but not centralized)
- #556 (output label naming conventions)

kellijohnson-NOAA:
# What is the feature?
* No longer passes the uncertainty to parameters. The distributional assumption now lives in the data.

* A formula system was created for the "uncertainty" column of FIMSFrame@data. The column takes the form of "~<distribution>(<central_tendency_label> = <derived_quantity>, <uncertainty_parameter_label> = <user_input>)". For example, "~lognormal(mean = log_landings_expected, sd = 0.001)" would be a viable entry for landings data.
* `parse_data_distribution()` parses this out into a list which is then used to set up the data distributions among other things

* I deleted quite a few tests so we will need to check coverage
* This PR and the parent branch are behind main by a commit and thus still failing R CMD check for linux.

___

## Checklist


data-raw/fims_distributions.R
Bai-Li-NOAA, line 1: non-blocking suggestion: A tabular layout might be easier to read here and would make it straightforward to add new distributions in the future. For example, we could use `tibble::tribble()`:

```r
fims_distributions <- tibble::tribble(
  ~family,       ~mean,  ~sample_function, ~other_parameters,
  "gaussian",    "mean", "rnorm",           list("sd"),
  "lognormal",   "mean", "rlnorm",          list("sd"),
  "multinomial", "prop", "rmultinom",       list("size"),
  "normal",      "mean", "rnorm",           list("sd")
)
```

Gemini: defining a small helper function (e.g., `add_distribution()`) so that adding a new distribution is as simple as adding a single line of code.

vignettes/fims-demo-projections.Rmd
```diff
@@ -100,17 +100,17 @@ data_4_projections <- data_big_with_extra_year |>
...
     uncertainty = ifelse(
       type %in% c("index") & value == -999,
-      0.19804220,
+      "~ lognormal(mean = log_index_expected, sd = 0.19804220)",
       uncertainty
     ),
     uncertainty = ifelse(
       type %in% c("age_comp", "length_comp") & value == -999,
-      0,
+      "~ multinomial(prop = agecomp_proportion, size = 0)",
```
Bai-Li-NOAA: For the fallback uncertainty string used when `type %in% c("age_comp", "length_comp")`, this always sets `prop = agecomp_proportion`. Do we need to link to `lengthcomp_proportion` for `length_comp`?

R/distribution_formulas.R
```diff
@@ -280,60 +179,61 @@ get_expected_name <- function(family, data_type) {
...
-    # is correct, as creating a new VariableVector for log_sd here would
-    # trigger an error in integration tests with wrappers.
-    new_module$log_sd[] <- log(sd[["value"]])
-    new_module$log_sd$set_estimation_types(sd[["estimation_type"]])
+    sd_values <- purrr::map_dbl(
+      uncertainty_split,
+      list("parameters", "sd"),
+      .default = NA_real_
+    )
+    new_module$log_sd[] <- log(sd_values)
```
Bai-Li-NOAA: Do we need to add checks here to ensure `sd_values` are valid? R won't throw an error if ` .default = NA_real_ `. Do we need to check some potential edge cases as well, such as `sd_value` being 0 or negative?

R/parse_data_distribution.R
```diff
@@ -0,0 +1,220 @@
...
+      is_scalar_na(family) ||
+      !nzchar(family)
+  ) {
+    return(NA_character_)
+  }
+
+  fims_distributions <- NULL
+
+  # First, prefer objects already loaded in the package namespace.
+  if (requireNamespace("FIMS", quietly = TRUE)) {
```
Bai-Li-NOAA: non-blocking suggestion: It might be worth creating a new issue to standardize how `fims_distributions` and `fims_input_types` are loaded across FIMS functions, since they currently appear to be loaded differently. I noticed that if a user doesn't call library(FIMS) and instead uses `FIMS::FIMSFrame(data_big)`, they get the following error:

`Error in validityMethod(object) : object 'fims_input_types' not found`

It would be useful to understand when these datasets are loaded into the environment and whether the loading behavior can be made more consistent.

R/parse_data_distribution.R
```diff
@@ -0,0 +1,220 @@
...
+  if (lang[[1]] != quote(`~`)) {
+    cli::cli_abort(
+      "Expression must be a distribution call (e.g., ~normal(0, 1))"
+    )
+  }
+  right_hand <- if (length(lang) == 3) lang[[3]] else lang[[2]]
+
+  # 3. Ensure we are looking at a function call (e.g., normal(0, 1))
+  if (!is.call(right_hand)) {
+    stop("Expression must be a distribution call (e.g., ~normal(0, 1))")
```
Bai-Li-NOAA: could use `cli::cli_abort()` for consistent formatting.

R/parse_data_distribution.R
```diff
@@ -0,0 +1,220 @@
...
+  # We use match.call or just slice the call object
+  # Ensure the arguments retain names if provided, or remain positional
+  # e.g., list(mean = 0, sd = 1) or list(0, 1)
+  args_list <- as.list(right_hand[-1])
+
+  # 6. Extract the left-hand side if it exists
+  if (is.call(lang) && lang[[1]] == quote(`~`)) {
+    # One-sided formula (~expr) has length 2; RHS is at index 2
+    # Two-sided formula (y ~ expr) has length 3; RHS is at index 3
+    right_hand <- if (length(lang) == 3) lang[[3]] else lang[[2]]
```
Bai-Li-NOAA: Do we need to extract `right_hand` again here, since it was already extracted in Step 1 (line 84)?

R/parse_data_distribution.R
```diff
@@ -0,0 +1,220 @@
...
+    cli::cli_abort("You must pass a string call.")
+  }
+  if (lang[[1]] != quote(`~`)) {
+    cli::cli_abort(
+      "Expression must be a distribution call (e.g., ~normal(0, 1))"
+    )
+  }
+  right_hand <- if (length(lang) == 3) lang[[3]] else lang[[2]]
+
+  # 3. Ensure we are looking at a function call (e.g., normal(0, 1))
```
Bai-Li-NOAA: Maybe remove the step numbers? It looks like Step 2 is missing.

R/parse_data_distribution.R
```diff
@@ -0,0 +1,220 @@
+#' Parse a distribution formula into structured components
```
Bai-Li-NOAA, line 1: I'm definitely not the best reviewer for the parsing logic. 😄 Maybe we could have Copilot generate some tests for this function, then add a few edge cases to make sure it behaves as expected.

data-raw/fims_distributions.R
```diff
@@ -0,0 +1,15 @@
+fims_distributions <- data.frame(
+  family = c("gaussian", "lognormal", "multinomial", "normal"),
```
Bai-Li-NOAA: Do we want to expose `normal` to users? I thought the implementation uses `gaussian` rather than `normal`. See `R/distribution_formulas.R`.

tests/testthat/test-distribution-formulas.R
```diff
@@ -199,90 +178,23 @@ test_that("`initialize_process_distribution()` returns correct error messages",
...
-      sd = list(value = fleet_sd, estimation_type = "constant"),
+      uncertainty = "~multinomial(prop = agecomp_proportion, size = 10)",
       data_type = "length_comp"
     ),
     "must be one of"
   )
 
-  #' @description Test that error is thrown when `sd` value and `estimation_type` dimensions do not match.
-  expect_error(
-    initialize_data_distribution(
```
Bai-Li-NOAA, line 259: Some of the existing edge case tests for sd could still be useful. I wonder if we could convert them to the new interface and keep them. For example, tests for missing required sd or sample size, etc.

data-raw/data_big.R
```diff
@@ -261,7 +261,11 @@ landings_data <- data.frame(
   timing = returned_om[["om_input"]][["year"]],
   value = returned_om[["em_input"]][["L.obs"]][[1]],
   unit = "mt", # metric tons
-  uncertainty = cv_2_sd(returned_om[["em_input"]][["cv.L"]][[1]])
+  uncertainty = paste(
+    "~ lognormal(mean = log_landings_expected, sd = ",
```
Bai-Li-NOAA: How are users expected to know what to provide for the distribution mean? It might be helpful to add some documentation explaining the expected input. This could also be a new issue.

Bai-Li-NOAA (approved):
I think I ended up leaving more non-blocking suggestions that could become follow-up issues. 😉 Let me know if you have any questions.

data-raw/data_big.R
kellijohnson-NOAA: This is a really great question and one that I struggled with. I thought that I was using the arguments from the associated distribution in R. For example `rnorm(n, mean = 0, sd = 1)` and `rmultinom(n, prop, size)` but then I noticed that I did not pay attention to the fact that `rlnorm()` uses meanlog and sdlog 🤦‍♀️. So, maybe I should change the arguments for the lognormal? Or, we should follow some other convention?

Bai-Li-NOAA: Oh, I was not aware of the `mean` vs. `meanlog` convention. I thought using the `mean` was fine for a lognormal distribution as long as it was included in the `fims_distribution` lookup table. I was not sure how users would know where to find `log_landings_expected` or `agecomp_proportion` and then include them in the paste() string when setting up FIMS data.

data-raw/fims_distributions.R
kellijohnson-NOAA: Thanks for the suggestion @Bai-Li-NOAA. I changed to using the tribble code you provided. I did not add a helper function to add a distribution because that would just be one more thing for us to maintain and it helps people understand what is going on if they have to add the code themselves.

data-raw/fims_distributions.R
kellijohnson-NOAA: I have since removed normal and guassian after talking to Andrea about this. Instead, I am using the names that are native to R, e.g., dnorm and dlnorm, so R users are familiar with what is allowed.

R/parse_data_distribution.R
kellijohnson-NOAA: Thanks for noticing the use of stop here. I have removed much of this function now and made things simpler so this problem here no longer exists.

R/parse_data_distribution.R
kellijohnson-NOAA: LOL, no we don't need this duplicated code and I have removed it. Thank you.

R/parse_data_distribution.R
kellijohnson-NOAA: This was copilot making things too complicated. Again, I have removed much of this code here so it is no longer a problem.

data-raw/data_big.R
kellijohnson-NOAA: I added documentation to FIMSFrame and also added a link to this documentation in the documentation for data_big$uncertainty.

R/parse_data_distribution.R
kellijohnson-NOAA: Step numbers, i.e., #2, are removed from the comments now.

vignettes/fims-demo-projections.Rmd
kellijohnson-NOAA: Thank you for catching this syntax error with respect to length composition data. I have since fixed it.

R/distribution_formulas.R
kellijohnson-NOAA: Great suggestion @Bai-Li-NOAA . I added one to check that sdlog and sd are always positive values. Are there any other checks that you can think of?

R/parse_data_distribution.R
kellijohnson-NOAA: I used copilot to generate some tests and then added some additional ones myself.

tests/testthat/test-distribution-formulas.R
kellijohnson-NOAA: These are largely still tested for in the new tests that I wrote.

kellijohnson-NOAA:
@Bai-Li-NOAA I did quite a bit of work to this branch since your review. I think that we should merge it into your branch as a single commit and then merge the whole thing into main. I am going to open a PR in r4ss/r4ss and NOAA-FIMS/case-studies to update both of the repos to use this new version and I would like to get approval of those two PRs before we merge this stuff into main. That way people are reviewing the user experience rather than the logic. What do you think?

Issue #1414, opened 2026-04-07
Add Internal Growth Estimation to FIMS


awilnoaa:
### Description

This issue issue is to track the addition of internal growth estimation to FIMS.  Users who want to look at the current progress on this work can check the `dev-vonb-growth` branch.

At this point, this branch already includes:

- a von Bertalanffy growth module
- fixed growth support though the existing growth module pathway
- growth products such as mean length-at-age, SD length-at-age, and weight-at-age
- growth derived ALK/WAA support in catch-at-age

Fixed growth is available through `VonBertalanffy` when growth parameters are treated as constants rather than estimated.

What is still missing is a full **internal growth estimation pathway** inside the integrated model, where growth can be informed directly by growth observations rather than only the fixed inputs. 

### Next Steps

- refactoring the ALK logic into a more modular polymorphic interface
- adding an initial internal growth estimation pathway for one direct growth-data type
- external growth estimation support, where growth is estimated outside of the full CAA model

iantaylor-NOAA:
Sorry I haven't had a chance to try out the growth module yet in spite of being really excited about it. Thank you @awilnoaa for all your work on this. I can't believe that it has come together so quickly!

Since variability in growth was brought up as a topic for the CIE review tomorrow, I took a quick look at the [dev-vonb-growth](https://github.com/NOAA-FIMS/FIMS/compare/main...dev-vonb-growth) branch and see that the it is parametrized via a vector of `length_at_age_sd_at_ref_ages`. I think this is a great setup. It will be good to have a discussion of whether additional options are needed and if so, which ones.

SS3 has 5 options for parameterizing variability in growth, which seems excessive. I think ALL of the options can produce similar results for many stocks and we haven't adequately researched when one might be preferred. Also, SS3 users rarely explore more than one option.

However, in case it's helpful for the discussion, here is the frequency of use of the 5 options in SS3 for 40 U.S. West Coast Groundfish assessments conducted between 2015 and 2023 (models that I had readily available to summarize):
```
       Var1 Freq
  CV=f(LAA)   31
  SD=F(LAA)    4
    CV=F(A)    3
 logSD=f(A)    1
    SD=F(A)    1
```
I honestly have no idea why the CV=f(LAA) is so much more commonly used. Perhaps it's just because it's the first option on the list (option 0). I'm also sure many of us are just copying the same example models and not exploring the other options.

[...]

PR #1697, opened 2026-08-17
Add size and growth infrastructure for VonB growth estimation and growth-derived ALK

### LINKED ISSUE #1415: Refactor ALK Handling into a Polymorphic Interface
### Description
The current ALK logic in `catch_at_age` mixes together:

- fixed age-to-length conversion behavior
- growth-derived ALK behavior
- growth-product lookup
- row-building logic
- fleet/path validity checks

This issue is to refactor that logic into a more modular polymorphic interface so ALK behavior is more consistent with existing FIMS patterns and easier to maintain, extend, and test.

### Proposed Fix
Right now, ALK handling is spread across multiple helper functions and conditional branches in `catch_at_age`. This refactor would introduce an `ALKBase` base class and child classes for different ALK types, for example:

- `FixedMatrixALK`
- `GrowthDerivedALK`

awilnoaa:
# What is the feature?
- Adds the core size and growth infrastructure needed for VonB growth estimation and growth-derived ALK/WAA support in FIMS.

- Added size-grid infrastructure under `inst/include/population_dynamics/size/`.
- Added polymorphic ALK infrastructure under `inst/include/population_dynamics/alk/`.
- Added growth model infrastructure under `inst/include/population_dynamics/growth/`.
- Added the Von Bertalanffy growth functor in `inst/include/population_dynamics/growth/functors/von_bertalanffy.hpp`.
- Updated catch-at-age model wiring in `inst/include/models/functors/catch_at_age.hpp` so fixed ALK and growth-derived ALK paths can share model logic.
- Added R/Rcpp initialization support in `R/initialize_modules.R`, `R/setup_default_parameters.R`, and the Rcpp interface files.
- Added gtests, R integration tests, and vignettes covering the new growth path.

- This affects FIMS model infrastructure, C++/Rcpp wiring, R initialization, tests, and vignettes.
- Closes #1415
- Partially addresses #1414; direct user-input growth observations remain future work.

___

## Checklist


inst/include/models/functors/catch_at_age.hpp
```diff
@@ -626,12 +652,12 @@ class CatchAtAge : public FisheryModelBase<Type> {
...
         this->GetPopulationDerivedQuantities(population->GetId());
 
-    Type phi_0 = CalculateSBPR0(population);
+    Type phi0 = CalculateSBPR0(population);
 
     if (i_dev == population->n_years) {
       dq_["numbers_at_age"][i_age_year] =
           population->recruitment->evaluate_mean(
-              dq_["spawning_biomass"][year - 1], phi_0);
+              dq_["spawning_biomass"][year - 1], phi0);
```
kellijohnson-NOAA: Not sure why this object was renamed here and elsewhere.

tests/gtest/test_growth.cpp
kellijohnson-NOAA, line 1: This file is not registered in CMakeLists.txt so it is not actually being tested.

tests/testthat/test-get_estimates.R
```diff
@@ -25,6 +25,13 @@ expected_colnames <- c(
   "lpdf", "likelihood", "log_sd", "log_like_cv", "gradient"
 )
 
+canonicalize_parameter_ids <- function(x) {
+  ids <- unique(x$parameter_id[!is.na(x$parameter_id)])
+  x$parameter_id <- match(x$parameter_id, ids)
+  x
+}
+
+
```
kellijohnson-NOAA:
```suggestion
```
Please remove this function and its use throughout the tests. We want the snapshots to have the parameter ids from the model and check that parameter id is the same as it was before. This does result in the snapshot being updated quite regularly but it also alerts us if something changes order when we don't want it to.

vignettes/fims-code-club-growth.Rmd
```diff
@@ -0,0 +1,1007 @@
...
+code span.co{
+  color: #8c8c8c;
+}
+</style>
+
+This vignette walks through the new Von Bertalanffy growth pathway in FIMS.
+It is intended to help explain the changes that make growth an active and
+estimable part of the model, run the new pathway using the standard FIMS
+workflow, and inspect the outputs needed to confirm that the branch is working
+as expected.
```
kellijohnson-NOAA: Throughout this file, paragraphs should be a single line without hard returns, which helps us determine changes later on rather than focusing on white-space changes.

vignettes/fims-code-club-growth.Rmd
```diff
@@ -0,0 +1,1007 @@
...
+}
+```
+
+<style type="text/css">
+code span.co{
+  color: #8c8c8c;
+}
+</style>
+
+This vignette walks through the new Von Bertalanffy growth pathway in FIMS.
```
kellijohnson-NOAA: All instances of "Von Bertalanffy" should be "von Bertalanffy"

kellijohnson-NOAA (changes requested):
Some initial thoughts because I haven't finished reviewing 🫤 sorry about that
* In general we are trying to stay away from abbreviations so I suggest the following parameter names: mean_length_young, mean_length_old, and von_Bertalanffy_coefficient. I am not set on the last one but we really want things to be pretty self-explanatory when reading the code. Why are we using ALK rather than age_length_conversion like what we had in the data?
* I think we need to name this version of growth something other than von Bertalanffy, more like Schnute--Fournier or maybe just the Schnute, I am honestly not sure which Gemini told me the former but we want to save von Bertalanffy for when we use the delta method
* I am going through the vignette now but I don't think that we need the extra vignette that was added for code club and we should choose just one example rather than 2 for the main growth vignette
* Andrea is going to review the C++ code

awilnoaa:
Thanks Kelli! I pushed updates for these comments.

Main changes:
- Renamed to `VonB-Schnute` so it is clear this is the Schnute-style two-reference-age parameterization. Let me know if this name works. I was thinking we could use something like `VonB-Traditional` later when that version is added.
- Updated `setup_default_parameters()` so the Growth default is data-aware. If empirical `weight_at_age` rows are present, it defaults to `EWAA`. If empirical `weight_at_age` rows are absent but `length_comp` rows are present, it defaults to `VonB-Schnute`. If neither empirical `weight_at_age` nor `length_comp` rows are present, it now errors instead of silently choosing a Growth setup. 
- Updated `setup_default_Growth()` so users can still explicitly choose the Growth module in a modular workflow.
- Updated the main Growth parameter names to what you suggested: `mean_length_young`, `mean_length_old`, and `von_bertalanffy_coefficient_K`.
- Removed the extra Code Club / exploratory Growth vignettes and kept one focused `fims-growth-vonb-schnute.Rmd` vignette.
- Removed `tests/gtest/test_growth.cpp` because it was not registered in `CMakeLists.txt` and the EWAA evaluate behavior it covered is already tested in the registered `tests/gtest/test_growth_EWAAGrowth_evaluate.cpp` tests (`HandlesCorrectInput` and `HandlesEdgeCase`). 
- Removed the parameter-id canonicalization from the `get_estimates()` tests and updated the snapshots so they keep the model parameter IDs.
- Restored `phi_0` in `catch_at_age.hpp`.

I also changed user-facing wording from “ALK” to “age-to-length conversion” in the R messages, test descriptions, and vignette text. I did not rename the internal backend/API objects like `GrowthDerivedALK`, `ALKBase`, or `growth_derived_alk_used` in this round because that seems like a larger follow-up naming pass. I can do that separately if we want to fully remove ALK terminology from the backend too.

iantaylor-NOAA:
@awilnoaa, thanks for all the good work on this branch. I've been working on the petrale case study and updated to adapt to the current state of the branch. I like the way it automatically detects the need for the growth module through the absence of age_to_length_conversion and weight_at_age data.

Regarding the "ALK" abbreviation, I would support eventually removing that acronym from the backend as well (no rush). That abbreviation is used in the SS3 source code and output but has created a little confusion because the traditional use of "age-length key" was opposite: a conversion from length to age rather than age to length (see Kimura 1977 https://cdnsciencepub.com/doi/10.1139/f77-052 or this tutoral from Giancarlo Correa https://giancarlomcorrea.netlify.app/post/crl-tutorial/). I've heard folks describe conditional age-at-length data as an age-length key, which I think shares more with the original usage of the term than the age-to-length conversion.

One more thought on growth. While working on the case study today I wrote some functions for working with the age-to-length conversion output: `get_realized_age_to_length_conversion_dataframe()` and `plot_age_to_length_conversion()` (currently in files in the `/content/R/` folder of the opakapaka-update branch of case-studies). Running those functions in the in-development [petrale case study](https://github.com/NOAA-FIMS/case-studies/blob/opakapaka-update/content/NWFSC-petrale.qmd) created the plot below. Is this worth incorporating into FIMS and/or the new growth vignette? It can also just live in the case-studies repo until there's time to include ideas like this as part of a larger collection of helper functions or plotting tools in the future.
age_to_length_conversion_plot

awilnoaa:
@iantaylor-NOAA thank you for looking over the branch and giving me some guidance for the ALK renaming. I really appreciate it!

@kellijohnson-NOAA we also talked a bit about the names and before I start making the changes, I wanted to make sure we are on the same page so the updates are consistent across the branch. I remember you mentioned something about being able to search for certain C++ vs R code but I checked my notes and missed writing that down.

I am using the existing FIMS style as my best guess here, but please let me know if this works or needs edits.

My understanding of the current style is:

* **User-facing R functions:** `snake_case`, except established constructors/classes
  Examples: `get_report()`, `fit_fims()`, `setup_default_parameters()`
  Existing exceptions: `FIMSFrame()`, `FIMSFit()`, `setup_default_Growth()`

* **Report names / derived quantity names:** `snake_case`
  Examples: `numbers_at_age`, `spawning_biomass`, `catch_weight_at_age`

* **Parameter labels:** `snake_case`
  Examples: `inflection_point`, `log_Fmort`, `log_init_naa`

* **C++ classes/structs:** `UpperCamelCase`
  Examples: `Information`, `DataObject`, `CatchAtAgeInterface`

* **C++ public/interface methods:** usually `UpperCamelCase`
  Examples: `SetGrowth()`, `SetFleetCatchData()`, `AddPopulation()`, `DoReporting()`

* **C++ lower-level utility methods, members, and local variables:** follow local style, often `snake_case`
  Examples: `clear()`, `resize()`, `mean_weight_at_age`, `spawning_biomass_ratio`

For the age-to-length conversion cleanup, I am thinking the following:

* `ALKBase` → `AgeToLengthConversionBase`
* `GrowthDerivedALK` → `GrowthDerivedAgeToLengthConversion`
* `FixedMatrixALK` → `FixedAgeToLengthConversion`
* `BuildFleetALK()` → `BuildFleetAgeToLengthConversion()`
* `BuildALKRow()` → `BuildAgeToLengthConversionRow()`
* `SupportsGrowthDerivedALK()` → `SupportsGrowthDerivedAgeToLengthConversion()`
* `ReportGrowthDerivedALKTensor()` → `ReportGrowthDerivedAgeToLengthConversionTensor()`
* `growth_derived_alk_used` → `growth_derived_age_to_length_conversion_used`
* `fleet->alk` → `fleet->age_to_length_conversion_model`

For the growth coefficient parameter, I am leaning toward:

* `von_bertalanffy_coefficient_K` → `growth_coefficient`

Instead if `von_bertalanffy_coefficient` or `von_bertalanffy_growth_coefficient`, because the module type already gives the growth-model context. When we add a traditional von Bertalanffy parameterization later, it could also use `growth_coefficient`. But if parameter labels need to be globally unique outside of the module type, then I’m open

Related names would also become:

* `log_sd_growth_coefficient_K` → `log_sd_growth_coefficient`
* `growth_coefficient_k_variance` → `growth_coefficient_variance`

Then there is the VonB-Schunte naming. I looked at both [Schnute (1981)](https://cdnsciencepub.com/doi/10.1139/f81-153) and [Schnute-Fournier (1980)](https://cdnsciencepub.com/doi/10.1139/f80-172)  papers.

My understanding is that Schnute and Fournier (1980) uses a two-reference-age reparameterization of the von Bertalanffy curve, with length at a young age, length at an old age, and a growth coefficient. That is close to what FIMS is using, but the paper is also tied to a broader length-frequency analysis framework, so calling the whole module `SchnuteFournier` may overstate what FIMS implements.

Schnute (1981) defines a broader flexible growth model with additional shape flexibility. FIMS is using Eq. 15, the von Bertalanffy special case, but not the full Schnute growth model, so calling the module just `Schnute` could also be misleading.

I may be in the weeds about this, but at this stage, I am thinking `vonBertalanffySchnute` is the clearest option. It keeps the focus on the von Bertalanffy growth curve, while acknowledging the Schnute-style reference-age parameterization. But I would like to confirm this is best before moving forward with all the changes.

Finally, I also wanted to follow up on Ian’s plotting/helper-function suggestion. The functions `get_realized_age_to_length_conversion_dataframe()` and `plot_age_to_length_conversion()` seem really useful for understanding the realized age-to-length conversion output.

At this stage, I am thinking these could probably stay in the `case-studies` repo for now while the report names and age-to-length conversion naming are still changing. Once the naming is stable, we could either add an example to the new Growth vignette or decide whether FIMS should eventually include a broader set of helper/plotting functions. Let me know your thoughts and thank you again for you guidance!

kellijohnson-NOAA:
Thanks @awilnoaa for providing your plan. Regarding your summary of how things are named, it is pretty spot on. The reason there are exceptions in "User-facing R functions" is that `FIMSFrame` and `FIMSFit` are R class names; `setup_default_Growth()` where Growth is capitalized because it is trying to match the C++ class name. I think that many of the "C++ lower-level utility methods have the snake case construct because they are being exported to R using Rcpp.

For your proposed names, can you please put the adjective at the end, e.g., `AgeToLengthConversionGrowthDerived`, `AgeToLengthConversionFixed`, `SupportsAgeToLengthConversionGrowthDerived`. But now that I am reading these. Do we need "GrowthDerived" or can we just use "Fixed" and "Estimated" instead of "Fixed" and "GrowthDerived" as the adjectives? I think `BuildFleetAgeToLengthConversion` should be `BuildAgeToLengthConversionFleet` but not entirely sure on that one. Then `ReportGrowthDerivedAgeToLengthConversionTensor` would be `ReportAgeToLengthConversionEstimatedTensor`.

I do not think that parameters need to be globally unique because the reporting says which module it is part of. I think if the parameters are comparable it would be great if they are the same name so `growth_coefficient` is good. But regarding related names, should "log_sd" and "variance" come before or after?

I love your suggestion and background for naming the growth module! 🎉 One minor change would to use `VonBertalanffySchnute` to keep with true PascalCase.

Thank you for investigating Ian's functions. I agree that we can keep them in a case study and have it be part of an additional PR to bring those in at a later date. I am trying to get better about keeping PRs realistic 🤦‍♀️ 😉

inst/include/common/information.hpp
```diff
@@ -531,6 +535,47 @@ class Information {
...
+
+    f->age_to_length_conversion_model =
+        fims_popdy::BuildAgeToLengthConversionFleet<Type>(p, f);
+
+    if (f->age_to_length_conversion_model != nullptr &&
+        f->age_to_length_conversion_model->IsActive()) {
+      return;
+    }
+
+    if (f->n_lengths > 0) {
```
Andrea-Havron-NOAA: What about the error where a user supplies the age-to-length-bin conversion path but forgets to provide length comps? I think this would result in `BuildAgeToLengthConversionFleet` returning a `nullptr` but it doesn't throw an error anywhere here in information. Is this type of error caught elsewhere? You could add a `if( n_lengths == 0 } { throw error } just before the return on line 564?

inst/include/models/functors/catch_at_age.hpp
```diff
@@ -938,6 +989,378 @@ class CatchAtAge : public FisheryModelBase<Type> {
...
+  }
+
+  /**
+   * @brief Build growth-derived fleet mean WAA cache for the current
+   * evaluation.
+   *
+   * Only [year, age] mean WAA is cached. The full derived age-to-length
+   * conversion tensor is not materialized in objective-side evaluation.
+   */
+  void RefreshFleetGrowthDerivedMeanWAACache() {
```
Andrea-Havron-NOAA: It appears you are building the ALK matrix multiple times in catch_at_age.hpp. While this makes sense from a storage perspective, it is an expensive calculation (nyears x nages x nlengths) and I think we could simplify this. I explored this more with copilot and got the following suggestion:


> The current RefreshFleetGrowthDerivedMeanWAACache() builds each fleet-specific ALK row to populate a [year, age] cache of fleet mean WAA. GrowthDerivedFleetMeanWeightAA() later retrieves that cache during reporting. However, evaluate_length_comp() independently builds the same (year, age) ALK rows for fleets with length composition data. This means the current approach constructs the ALK at least twice per fleet-year-age evaluation, while only retaining the smaller WAA result.
> 
> growth_derived_mean_WAA is a fleet-level reporting quantity, and its report logic is already limited to isDouble<Type>::value. Consider removing the WAA cache and RefreshFleetGrowthDerivedMeanWAACache() from Evaluate(), then calculate growth_derived_mean_WAA[i_age_year] directly in Report() with MeanWeightFromAgeToLengthConversionDerived(...). This avoids building report-only WAA during AD/objective evaluation and removes the cache machinery.
> 
> Separately, the required per-iteration ALK calculation in evaluate_length_comp() should remain because it depends on the current estimated growth parameters. The static rebinning weights used to map population size bins to fleet length bins are a better caching target: they depend only on the population and fleet bin edges, so they can be built once when those grids are configured and reused across evaluations.


The only incorrect point about this is that we are currently calling Report() at every iteration, but that is a separate issue that will be fixed outside this work. 

This is a non-blocking change request, but I think it should be addressed eventually. If not in this PR, file an issue so that it gets addressed in the future.

inst/include/population_dynamics/size/growth_derived_size_provider.hpp
```diff
@@ -0,0 +1,263 @@
...
+    const Type plus_group_warning_threshold = static_cast<Type>(0.01);
+    bool plus_group_warning_needed = false;
+    std::size_t warning_year_index = 0;
+    std::size_t warning_age_index = 0;
+
+    for (std::size_t year_index = 0; year_index < n_years_; ++year_index) {
+      for (std::size_t age_index = 0; age_index < n_ages_; ++age_index) {
+        fims::Vector<Type> prob_row(population_size_grid_->n_bins);
+        Type row_sum = static_cast<Type>(0.0);
+        const Type negative_tolerance = static_cast<Type>(1e-10);
```
Andrea-Havron-NOAA: remove negative tolerance definition

inst/include/population_dynamics/size/growth_derived_size_provider.hpp
```diff
@@ -0,0 +1,263 @@
...
+
+        for (std::size_t size_bin_index = 0;
+             size_bin_index < population_size_grid_->n_bins; ++size_bin_index) {
+          Type bin_prob = PopulationSizeBinProb(growth_products, year_index,
+                                                age_index, size_bin_index);
```
Andrea-Havron-NOAA: define: `static const Type  minimum_bin_prob = static_cast<Type>(1e-12);` just above he start of the year_index, age_index loop.

and then add minimum_bin_prob to bin_prob: `Type bin_prob = PopulationSizeBinProb(...) + minimum_bin_prob;`

inst/include/population_dynamics/size/growth_derived_size_provider.hpp
```diff
@@ -0,0 +1,263 @@
...
+
+          if (!(bin_prob >= -negative_tolerance)) {
```
Andrea-Havron-NOAA: remove check - this condition is not possible

inst/include/population_dynamics/size/growth_derived_size_provider.hpp
```diff
@@ -0,0 +1,263 @@
...
+            throw std::runtime_error(
+                "GrowthDerivedSizeProvider produced a materially negative "
+                "population size probability");
+          }
+
+          if (bin_prob < static_cast<Type>(0.0)) {
```
Andrea-Havron-NOAA: remove check, this condition is no longer possible

inst/include/population_dynamics/size/growth_derived_size_provider.hpp
```diff
@@ -0,0 +1,263 @@
...
+            bin_prob = static_cast<Type>(0.0);
+          }
+
+          prob_row[size_bin_index] = bin_prob;
+          row_sum += bin_prob;
+        }
+
+        if (!(row_sum > static_cast<Type>(1e-12))) {
```
Andrea-Havron-NOAA: remove check, this condition is no longer possible

inst/include/population_dynamics/size/growth_derived_size_provider.hpp
```diff
@@ -0,0 +1,263 @@
...
+          throw std::runtime_error(
+              "GrowthDerivedSizeProvider produced a zero-mass population "
+              "size row");
+        }
+
+        const Type safe_row_sum =
```
Andrea-Havron-NOAA: `safe_row_sum` is no longer needed and can just be changed to `row_sum` below on line 136. Removing this function here means the size_probability_normalization.hpp file can go away.

inst/include/population_dynamics/age_to_length_conversion/functors/age_to_length_conversion_derived.hpp
```diff
@@ -0,0 +1,366 @@
...
+   */
+  bool TrySanitizeProbabilityRowInPlace(
+      fims::Vector<Type>& row, std::size_t expected_size,
+      const Type& negative_tolerance = static_cast<Type>(1e-12)) const {
+    if (row.size() != expected_size) {
+      return false;
+    }
+
+    for (std::size_t i = 0; i < row.size(); ++i) {
+      if (!(row[i] >= -negative_tolerance)) {
```
Andrea-Havron-NOAA: `row` is dependent on a parameter and cannot be used in a conditional statement.

Because TMB uses a statis tape, this conditional will be set with the initial values and not update as parameters update, effectively "freezing" the conditional at its initial TRUE or FALSE. 

The following lines need to be updated: 204, 208, 237, 265, 271, and 124 in growth_derived_size_provider.hpp

Recommended change: remove this function entirely and add 1e-12 to all row[i] values (see comments in growth_derived_size_provider.hpp::PrepareSizeProducts() and age_to_length_conversion_derived.hpp:: PopulationSizeBinProb(), BuildPopulationProbSizeRow(), and TryFinalizeMappedProbabilityRow()). It is ok to remove the -negative_tolerance check - you've already checked that bins always increase with `HasStrictlyIncreasingEdges` so `normalcdf` will always return a positive value so there should never by an instance where `row[i]` is negative.

inst/include/population_dynamics/size/growth_derived_size_provider.hpp
```diff
@@ -0,0 +1,263 @@
...
+
+    if (size_bin_index == 0) {
+      const Type upper = static_cast<Type>(population_size_grid_->edges[1]);
+      return fims_math::normalcdf(upper, mean_laa, sd_laa);
+    }
+
+    if (size_bin_index + 1 == population_size_grid_->n_bins) {
+      const Type lower = static_cast<Type>(
+          population_size_grid_->edges[population_size_grid_->n_bins - 1]);
+      return static_cast<Type>(1.0) -
```
Andrea-Havron-NOAA: Change to:
```
fims_math::normalcdf(static_cast<Type>(2.0) * mean_laa - lower, mean_laa, sd_laa)
```
This is called an upper tail reflection: P(X > lower) = P(X < 2*mean - lower). The value 1 has less precision than 0 to if normalcdf(lower) is small enough, it will round to 1.0, which would return 0.

inst/include/population_dynamics/age_to_length_conversion/functors/age_to_length_conversion_derived.hpp
```diff
@@ -0,0 +1,366 @@
...
+    }
+
+    const SizeGrid* population_size_grid = size_provider_ptr->TryGetSizeGrid();
+    if (population_size_grid == nullptr ||
+        !population_size_grid->IsConsistent() ||
+        population_size_grid->n_bins == 0) {
+      return fims::Vector<Type>();
+    }
+
+    fims::Vector<Type> prob_size_row(population_size_grid->n_bins);
```
Andrea-Havron-NOAA, line 166: after this add:

```
for(std::size_t i = 0; i < population_size_grid->n_bins; ++i) {
  prob_size_row[i] = size_provider_ptr->ProbSize(year_index, age_index, i);
}

return prob_size_row;
```

The try/catch statement and code from 178-183 can be removed because changes recommended in `PrepareSizeProducts` ensures these values are calculated in a valid way. Also we shouldn't catch runtime_errors and return empty vectors. FIMS should be set up now so that any std::exceptions can get carried through and printed to the screen in R.

inst/include/population_dynamics/age_to_length_conversion/functors/age_to_length_conversion_derived.hpp
```diff
@@ -0,0 +1,366 @@
...
+      fims::Vector<Type>& row, std::size_t expected_size,
+      const Type& minimum_mass = static_cast<Type>(1e-12),
+      const Type& minimum_bin_prob = static_cast<Type>(1e-12)) const {
+    if (!TrySanitizeProbabilityRowInPlace(row, expected_size)) {
+      return false;
+    }
+
+    Type row_sum = static_cast<Type>(0.0);
+    for (std::size_t i = 0; i < row.size(); ++i) {
+      if (row[i] < minimum_bin_prob) {
```
Andrea-Havron-NOAA: remove conditional, just calculate:

```
row[i] = minimum_bin_prob;
row_sum += row[i]
```

inst/include/population_dynamics/age_to_length_conversion/functors/age_to_length_conversion_derived.hpp
```diff
@@ -0,0 +1,366 @@
...
+        row[i] = minimum_bin_prob;
+      }
+      row_sum += row[i];
+    }
+
+    if (!(row_sum > minimum_mass)) {
```
Andrea-Havron-NOAA: this check can be removed

inst/include/population_dynamics/age_to_length_conversion/functors/age_to_length_conversion_derived.hpp
```diff
@@ -0,0 +1,366 @@
...
+
+  /**
+   * @brief Clamp tiny negative roundoff and reject materially invalid
+   * probability rows.
+   * @param row Probability row to sanitize in place.
+   * @param expected_size Expected number of bins.
+   * @param negative_tolerance Smallest allowed negative value before failing.
+   * @return True if the row is usable after sanitization.
+   */
+  bool TrySanitizeProbabilityRowInPlace(
```
Andrea-Havron-NOAA: remove function

inst/include/population_dynamics/age_to_length_conversion/functors/age_to_length_conversion_derived.hpp
```diff
@@ -0,0 +1,366 @@
...
+   * @param row Fleet probability row on observation bins.
+   * @param expected_size Expected number of fleet bins.
+   * @param minimum_mass Minimum acceptable total probability mass.
+   * @return True if the mapped fleet row can be used downstream.
+   */
+  bool TryFinalizeMappedProbabilityRow(
+      fims::Vector<Type>& row, std::size_t expected_size,
+      const Type& minimum_mass = static_cast<Type>(1e-12),
+      const Type& minimum_bin_prob = static_cast<Type>(1e-12)) const {
+    if (!TrySanitizeProbabilityRowInPlace(row, expected_size)) {
```
Andrea-Havron-NOAA: remove check

inst/include/population_dynamics/age_to_length_conversion/functors/age_to_length_conversion_derived.hpp
```diff
@@ -0,0 +1,366 @@
...
+  /**
+   * @brief Validate a prepared population probability row and renormalize small
+   * roundoff.
+   * @param row Prepared population probability row.
+   * @param expected_size Expected number of bins.
+   * @param minimum_mass Minimum acceptable total probability mass.
+   * @param mass_tolerance Allowed deviation from unit mass before failing.
+   * @return True if the row is a usable prepared probability row.
+   */
+  bool TryFinalizePreparedProbabilityRow(
```
Andrea-Havron-NOAA: remove function

inst/include/population_dynamics/age_to_length_conversion/functors/age_to_length_conversion_derived.hpp
```diff
@@ -0,0 +1,366 @@
...
+  /**
+   * @brief Validate and normalize one mapped fleet probability row.
+   * @param row Fleet probability row on observation bins.
+   * @param expected_size Expected number of fleet bins.
+   * @param minimum_mass Minimum acceptable total probability mass.
+   * @return True if the mapped fleet row can be used downstream.
+   */
+  bool TryFinalizeMappedProbabilityRow(
+      fims::Vector<Type>& row, std::size_t expected_size,
+      const Type& minimum_mass = static_cast<Type>(1e-12),
```
Andrea-Havron-NOAA: remove minimum_mass

inst/include/population_dynamics/age_to_length_conversion/functors/age_to_length_conversion_derived.hpp
```diff
@@ -0,0 +1,366 @@
...
+#include <cstddef>
+#include <memory>
+
+#include "../../../common/fims_math.hpp"
+#include "age_to_length_conversion_base.hpp"
+#include "../../fleet/fleet.hpp"
+#include "../../growth/growth_model_adapter.hpp"
+#include "../../size/size_distribution_provider_base.hpp"
+#include "../../size/functors/size_bin_mapping.hpp"
+#include "../../size/functors/size_probability_normalization.hpp"
```
Andrea-Havron-NOAA: remove

inst/include/interface/rcpp/rcpp_objects/rcpp_growth.hpp
```diff
@@ -285,4 +358,742 @@ class EWAAGrowthInterface : public GrowthInterfaceBase {
...
+
+    ConfigureLengthReferenceParameterization<Type>(vb);
+
+    load_transformed_and_register(
+        this->mean_length_young, vb->MeanLengthYoungVector(),
+        "mean_length_young", WorkingScaleInitialLengthAtRefAge1());
+    load_transformed_and_register(this->mean_length_old,
+                                  vb->MeanLengthOldVector(), "mean_length_old",
+                                  WorkingScaleInitialLengthAtRefAge2());
+    load_and_register(this->growth_coefficient, vb->GrowthCoefficientVector(),
```
Andrea-Havron-NOAA, line 1014: I find this a bit confusing. It looks like growth_coefficient gets logged before being passed to the backend. This is a bit problematic because this only happens when setting initial values. None of the interface code is called when TMB is reiterating over different parameter values. So there might be a transformation mismatch between the optimizer and the model. To simplify, remove the `load_transformed_and_register` function and only use the `load_and_register` function. Everything that gets exponentiated in the model should be given a `log_` prefix and passed into the back end as is on its logscale.

inst/include/population_dynamics/growth/functors/vonb_schnute.hpp
```diff
@@ -0,0 +1,248 @@
...
+  virtual ~VonBertalanffySchnuteGrowth() {}
+
+  /**
+   * @brief Validate the VonBertalanffySchnute parameters used by length-at-age
+   * calculations.
+   *
+   * Throws when the stored growth parameter values are incompatible with the
+   * raw VonBertalanffySchnute length-at-age calculation.
+   */
+  void ValidateLengthParameters() const {
```
Andrea-Havron-NOAA: Delete this function. First it uses parameter values as conditional tests and second, since these values are exponentiated, they can't be <=0. Anything that must be positive is handled as a logged parameter with the optimizer and then exponentiated in the model. This guarantees values stay > 0.

inst/include/population_dynamics/growth/functors/vonb_schnute.hpp
```diff
@@ -0,0 +1,248 @@
...
+          "VonBertalanffySchnuteGrowth mean_length_old must be > "
+          "mean_length_young");
+    }
+  }
+
+  /**
+   * @brief Validate the length-weight parameters used by weight-at-age
+   * calculations.
+   */
+  void ValidateWeightParameters() const {
```
Andrea-Havron-NOAA: delete per same reason above

inst/include/population_dynamics/growth/functors/vonb_schnute.hpp
```diff
@@ -0,0 +1,248 @@
...
+    }
+  }
+
+  /**
+   * @brief Evaluate mean length at age.
+   * @param age Age on the natural scale.
+   * @return Mean length at the requested age.
+   */
+  Type length_at_age(const Type& age) const {
+    ValidateLengthParameters();
```
Andrea-Havron-NOAA: remove

inst/include/population_dynamics/growth/functors/vonb_schnute.hpp
```diff
@@ -0,0 +1,248 @@
...
+   * mean_length_young.
+   * @param d_log_laa_d_l2 Output derivative with respect to
+   * mean_length_old.
+   * @param d_log_laa_d_k Output derivative with respect to
+   * growth_coefficient.
+   */
+  void log_length_at_age_gradient(const Type& age, Type& d_log_laa_d_l1,
+                                  Type& d_log_laa_d_l2,
+                                  Type& d_log_laa_d_k) const {
+    ValidateLengthParameters();
```
Andrea-Havron-NOAA: remove

inst/include/population_dynamics/growth/functors/vonb_schnute.hpp
```diff
@@ -0,0 +1,248 @@
...
+    d_log_laa_d_log_k = d_log_laa_d_k * growth_coefficient;
+  }
+
+  /**
+   * @brief Evaluate mean weight at age via the length-weight relationship.
+   * @param age Age on the natural scale.
+   * @return Mean weight at the requested age.
+   */
+  Type weight_at_age(const Type& age) const {
+    ValidateWeightParameters();
```
Andrea-Havron-NOAA: remove

Andrea-Havron-NOAA (changes requested):
First off, nice work! This was a massive project that you've managed very well and delivered an excellent product! 

One change that is required before merging into main is that there are multiple conditionals that depend on parameters. Since TMB uses a static tape, these conditionals essentially become frozen at the boolean set by the initial values and do not update as parameter values change. There are some work arounds I added, but a more stable change is to reparametrize the calculation of proportions in log-space and use helper function like `log-sum-exp`, `log_diff_exp`, `log1p` and `expm` to prevent under/overflow, which is definitely beyond the scope of this PR and can be filed as a new issue for future improvement. 

I also picked up on a transformation issue in the interface and some more clean-up around naming parameters log_* to keep track of what is in logscale. 

I would also like to add that there are a lot of conditional statements in this PR, so I had to use AI to find the ones that use conditional tests that are based on parameters. AI reported there roughly twice the number of conditionals used in the growth PR compared to main. This is super impressive because conditionals are challenging to write and track the logic correctly, so well done! We should, however, file an issue that aims to move some of the conditional branching in growth to polymorphic design patterns (e.g., subclassing/strategy patterns) to make the growth framework easier to extend and maintain.

I still have a bit more code to review, so I will add more comments if needed.

inst/include/common/information.hpp
iantaylor-NOAA: I don't understand all the path options but want to make sure that parametric growth doesn't depend on the presence of length comp data. I think from the vignette in this PR it looks like you can supply data with `type = "length_bin"`, which hopefully would be enough to enable the growth options. 

We've used fixed von Bertalanffy growth parameters with no age or length comps for production assessments in the U.S. west coast (for data-moderate stocks like English sole, report at https://www.pcouncil.org/documents/2015/01/data-moderate-stock-assessments-for-brown-china-copper-sharpchin-stripetail-and-yellowtail-rockfishes-and-english-and-rex-soles-in-2013-published-january-2015.pdf/).

kellijohnson-NOAA:
@awilnoaa where is documentation for the length_bin data type?

awilnoaa:
> @awilnoaa where is documentation for the length_bin data type?

Right now, length_bin is only documented indirectly. It is used in the `vonb-schnute vignette` (which I need to rename) and listed as an allowed FIMSFrame input type, but there is not a dedicated section explaining what `length_bin` is or when users should provide it. I could add a short `type == "length_bin"` section to `vignettes/fims-input-data.Rmd`. Is that what you had in mind?

kellijohnson-NOAA:
The data vignette would be a great place and the .R file that describes the data, i.e., `data_.R`.

inst/include/population_dynamics/size/growth_derived_size_provider.hpp
awilnoaa: Addressed. Removed the `negative_tolerance` definition and replaced the old negative-probability checks with a small `minimum_bin_prob` floor before normalization.

inst/include/population_dynamics/size/growth_derived_size_provider.hpp
awilnoaa: removed

inst/include/population_dynamics/size/growth_derived_size_provider.hpp
awilnoaa: removed

inst/include/population_dynamics/size/growth_derived_size_provider.hpp
awilnoaa: removed

inst/include/population_dynamics/size/growth_derived_size_provider.hpp
awilnoaa: Addressed. Changed normalization to use `row_sum` directly and removed the now-unused `size_probability_normalization.hpp` helper.

inst/include/population_dynamics/size/growth_derived_size_provider.hpp
awilnoaa: Addressed, changed to your suggested code block. Thank you!

inst/include/population_dynamics/age_to_length_conversion/functors/age_to_length_conversion_derived.hpp
awilnoaa: Addressed. I used the same direct ProbSize(...) loop and return pattern and kept the existing size_bin_index variable name

inst/include/population_dynamics/age_to_length_conversion/functors/age_to_length_conversion_derived.hpp
awilnoaa: Addressed. I removed the parameter-dependent probability-row checks and helper finalization functions, added the small `1e-12` floor before row normalization, and updated `BuildPopulationProbSizeRow()` so it returns the prepared `ProbSize(...)` row directly. One small difference from the suggested code: I used `row[i] += minimum_bin_prob` instead of `row[i] = minimum_bin_prob`. Please let me know if that is correct.

inst/include/population_dynamics/age_to_length_conversion/functors/age_to_length_conversion_derived.hpp
awilnoaa: Addressed. Removed the function.

inst/include/population_dynamics/age_to_length_conversion/functors/age_to_length_conversion_derived.hpp
awilnoaa: Addressed. Removed the function.

inst/include/population_dynamics/age_to_length_conversion/functors/age_to_length_conversion_derived.hpp
awilnoaa: Addressed with a small adjustment. I removed the conditional and used `row[i] += minimum_bin_prob` instead of `row[i] = minimum_bin_prob` so the calculated probability pattern is preserved before normalization.

inst/include/population_dynamics/age_to_length_conversion/functors/age_to_length_conversion_derived.hpp
awilnoaa: Addressed. Removed the check.

inst/include/population_dynamics/age_to_length_conversion/functors/age_to_length_conversion_derived.hpp
awilnoaa: Addressed. Removed the check.

inst/include/population_dynamics/age_to_length_conversion/functors/age_to_length_conversion_derived.hpp
awilnoaa: Addressed. Removed `minimum_mass`.

inst/include/interface/rcpp/rcpp_objects/rcpp_growth.hpp
awilnoaa: Thank you, I removed `load_transformed_and_register()` and the now-unused working-scale initial-value helpers. `mean_length_young` and `mean_length_old` now use `load_and_register(..., true)`

inst/include/population_dynamics/growth/functors/vonb_schnute.hpp
awilnoaa: Deleted

inst/include/population_dynamics/growth/functors/vonb_schnute.hpp
awilnoaa: done

inst/include/population_dynamics/growth/functors/vonb_schnute.hpp
awilnoaa: done

inst/include/population_dynamics/growth/functors/vonb_schnute.hpp
awilnoaa: removed

inst/include/population_dynamics/growth/functors/vonb_schnute.hpp
awilnoaa: done

inst/include/models/functors/catch_at_age.hpp
awilnoaa: Fixed

vignettes/fims-code-club-growth.Rmd
awilnoaa: updated

vignettes/fims-code-club-growth.Rmd
awilnoaa: updated

inst/include/common/information.hpp
awilnoaa: Thanks for catching that, Ian and prvding the paper. I fixed this so `VonBertalanffySchnute` growth does not soley require `length_comp` rows. The growth-derived age-to-length path can now also use `length_bin` rows as the fleet length-bin geometry, even when there are no `length_comp` rows.

The default Growth choice was unchanged: `length_bin` rows by themselves do not auto-select `VonBertalanffySchnute`. But if a user explicitly chooses that Growth type, the derived age-to-length path can now initialize from `length_bin` rows only.

Thanks Andrea, I also cleaned up the related initialization checks so the missing-bin and missing-conversion-path cases fail with clearer messages.

inst/include/population_dynamics/size/growth_derived_size_provider.hpp
awilnoaa: Addressed. I added the small probability floor before summing and normalizing the population size-bin probabilities.

I used a local `const Type` rather than `static const Type`.Please let me know if you would still prefer the `static const Type` pattern here.

inst/include/population_dynamics/age_to_length_conversion/functors/age_to_length_conversion_derived.hpp
awilnoaa: Removed

tests/gtest/test_growth.cpp
awilnoaa: Addressed. Removed the `test_growth.cpp` file. The remaining growth gtests are split into `growth_vonb_schnute`, `growth_model`, and `growth_model_adapter`.

tests/testthat/test-get_estimates.R
awilnoaa: Removed

inst/include/models/functors/catch_at_age.hpp
awilnoaa: Thank you Andrea, I went ahead and cleaned this up now. I removed the growth-derived fleet mean WAA cache and the related refresh call from `Evaluate()`.

`growth_derived_mean_WAA` is now calculated directly in `Report()` using `MeanWeightFromAgeToLengthConversionDerived(...)`, so that report-only WAA calculation is no longer being built during objective evaluation.

I left the length-composition age-to-length conversion calculation in `Evaluate()` unchanged because it still needs to use the current estimated growth parameters during each evaluation.

kellijohnson-NOAA:
I did a lot of renaming and Matthew helped me fix a 🐛 I introduced 🤦‍♀️. Once this passes all tests we can merge it into main.

PR #1706, opened 2026-08-24
Support TMB parameter maps

msupernaw:
# What is the feature?

- Adds support for TMB parameter maps in FIMS. Users can fix parameters with
  `NA` map entries or make parameters share an estimated value by assigning
  them the same factor level.

- Added internal handling that reduces the full FIMS parameter names to match
  the mapped optimization vector returned by `TMB::MakeADFun()`.
- Updated `fit_fims()` to expand the optimized vector through TMB's `parList()`
  before copying fixed and shared parameter values back into FIMS.
- Updated estimate reshaping so initial and optimized values remain aligned
  with the reduced TMB parameter indexing.
- Added validation for map type, map length, and the resulting number of mapped
  parameter names.
- Added tests covering fixed parameters, shared factor levels, mapped names,
  expansion back into FIMS, and aligned initial and fitted estimates.

- No other repository is affected. Within FIMS, the change is limited to TMB
  model fitting, parameter naming, and estimate reshaping. Existing workflows
  without a parameter map retain their current behavior.

___

## Checklist


R/fimsfit.R
kellijohnson-NOAA, line 1: I am getting errors from `FIMSFit()`
```
Error in `FIMSFit()`:
! TMB and FIMS produced different numbers of fixed parameters.
i TMB produced 96; FIMS produced 171 mapped names.
```
where I have 171 parameters and the map passed to TMB has the correct element "p", which is of length 171 but it has 96 levels. So, either I am doing something wrong or something is not working. This is using the Pacific hake example in the case studies.

msupernaw: > I am getting errors from `FIMSFit()`
> 
> ```
> Error in `FIMSFit()`:
> ! TMB and FIMS produced different numbers of fixed parameters.
> i TMB produced 96; FIMS produced 171 mapped names.
> ```
> 
> where I have 171 parameters and the map passed to TMB has the correct element "p", which is of length 171 but it has 96 levels. So, either I am doing something wrong or something is not working. This is using the Pacific hake example in the case studies.

@kellijohnson-NOAA if the map is passed into MakeADFun and not stored in input, FIMS fit has a null map. An immediate fix would look something like:

```{r}
input$map <- map  # map must be list(p = factor(...))

obj <- TMB::MakeADFun(
  data = list(),
  parameters = input$parameters,
  map = input$map,
  random = "re",
  DLL = "FIMS",
  silent = TRUE
)

fit <- FIMSFit(
  input = input,
  obj = obj,
  opt = opt,
  sdreport = sdreport,
  run_time = run_time
)
```

R/fimsfit.R
```diff
@@ -449,10 +477,25 @@ FIMSFit <- function(
...
+    cli::cli_abort(c(
+      "TMB and FIMS produced different numbers of fixed parameters.",
+      "i" = paste0(
+        "TMB produced {length(obj[[\"par\"]])}; FIMS produced ",
+        "{length(parameter_names)} mapped names."
+      )
+    ))
+  }
   names(obj[["par"]]) <- parameter_names
-  random_effects_names <- names(get_random_names(obj[["env"]]$parList()[["re"]]))
```
kellijohnson-NOAA, line 455: @msupernaw why did you remove this line?

msupernaw: @kellijohnson-NOAA It created an unused local variable—nothing read it or assigned those names to the fit.

R/fimsfit.R
kellijohnson-NOAA: I found what I was doing wrong, I was not labeling the map list entry, i.e. Map <- list(factor(...)), instead of map <- list(p = ...)).

R/fimsfit.R
```diff
@@ -637,7 +680,10 @@ fit_fims <- function(input,
 
   check_mle_convergence(input, obj, opt, maxgrad)
 
-  FIMS::set_fixed(opt[["par"]])
+  # Expand TMB's reduced optimization vector back to the full FIMS parameter
+  # vector before copying values into the C++ model. This is a no-op when no
+  # TMB map is present and restores fixed/shared entries when one is present.
+  FIMS::set_fixed(obj[["env"]]$parList(opt[["par"]])[["p"]])
```
kellijohnson-NOAA: @Andrea-Havron-NOAA can you confirm that this change is okay?

R/fimsfit.R
kellijohnson-NOAA: Perfect, thank you for the clean up.

R/reshape_output.R
kellijohnson-NOAA: @Andrea-Havron-NOAA can you review the changes to this file?

kellijohnson-NOAA:
I am pretty happy with how this branch worked for cobia, where I wanted to create "time blocks" but we don't have that functionality within FIMS. So, I mapped off the time series of selectivity parameters into two blocks. In the future we might think about creating some helper functions for people. @Bai-Li-NOAA we might even think about adding a column to the input tibble that supplies the factor-level entry if people want to map parameters together. Maybe we should make this a separate GitHub Discussion but I wanted to bring up the utility of this branch! I have a few small changes I am going to push that will fix up the code in the messages printed to the screen because cli has glue in it so you do not need to use paste.

PR #1724, opened 2026-09-01
add more exploration and explanation of default parameters to demo

iantaylor-NOAA:
# What is the feature?
* While chatting with @e-perl-NOAA and @nathanvaughan-NOAA after a meeting wrapped up earlier today, Elizabeth noted that the vignettes and case studies don't do a great job of explaining which of the default parameters you might want to update or why you would want to do so.

* I've added some explorations and explanations to the demo vignette to help users understand what the default parameters are and which you might want to modify in a real-world application. The modifications we make below this in the vignette are a good demo of the process even if they aren't things that you would necessarily focus on in a real-world application (e.g. values for the the `log_Fmort` parameters which should be easy enough to estimate based on the input catch values).

* No, I don't think so. Maybe similar additions could be made to the case studies to add narrative around why the parameters are being changed.

Vignette as rendered on my computer is attached: [fims-demo.html](https://github.com/user-attachments/files/31715686/fims-demo.html)

___

## Checklist

### DIFF
--- a/inst/WORDLIST
+++ b/inst/WORDLIST
@@ -63,6 +63,7 @@ fimsfit
 fimsframe
 fimslog
 firstname
+fishprior
 fishres
 Fmort
 fsel
--- a/vignettes/fims-demo.Rmd
+++ b/vignettes/fims-demo.Rmd
@@ -106,7 +106,6 @@ You can use the base R function `plot(data_4_model)` to see the data types (e.g.
 plot(data_4_model)
 ```
 
-
 ## Parameters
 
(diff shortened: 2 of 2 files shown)


e-perl-NOAA:
@iantaylor-NOAA I made only a couple of changes based on things that I didn't necessarily think were super clear. Would love to get thoughts from other though! Feel free to revert or remove any changes you don't agree with.

kellijohnson-NOAA:
I would like others to review these changes besides myself. Perhaps @MOshima-PIFSC or @dgoethel-noaa?

kellijohnson-NOAA (approved):
This is great! Thank you for taking the time to make this PR. I added some text and also provided an example of using `FIMS:::setup_default_log_init_naa()`. Once this passes the vignette-building stage of R CMD check I will bring it into main.

PR #1735, opened 2026-09-11
docs: plot age-to-length distributions across modeled ages

awilnoaa:
# What is the feature?
*  Replaces the single-age realized age-to-length conversion plot in the Growth vignette with a diagnostic figure showing length distributions across all modeled ages.

* Reshaped the reported growth-derived age-to-length conversion tensor into a tidy data frame and plotted the final modeled year with `ggridges::geom_density_ridges()`.
* This follows Ian’s suggested realized age-to-length conversion diagnostic figure from his PR comment: https://github.com/NOAA-FIMS/FIMS/pull/1697#issuecomment-5375371362
* The plotting approach is based on Ian’s case-studies helper: https://github.com/NOAA-FIMS/case-studies/blob/opakapaka-update/content/R/plot_age_to_length_conversion.R
* Added `ggridges` to `Suggests` because it is only used for the vignette plot.

* No expected impact outside the Growth vignette. No exported helper functions or model code were changed.

___

## Checklist


awilnoaa:
I also looked into the two helper functions @iantaylor-NOAA  mentioned: get_realized_age_to_length_conversion_dataframe() and plot_age_to_length_conversion() from the case-studies repo to try to figure out where those should live.

@kellijohnson-NOAA  recently showed me [r4ss::SS_plots() ](https://github.com/r4ss/r4ss/blob/main/R/SS_plots.R)as an example of a larger collection of model diagnostic plots. Like Ian mentioned in his previous comment, maybe these functions could eventually be part of a larger collection of FIMS helper functions or plotting tools. I’m not sure they fit in stockplotr if that is mainly intended for report-level plots and these are more useful for model diagnostics and checking outputs.

I’m interested in your thoughts and am also happy to make a separate discussion post about where these types of functions should live in the future.

awilnoaa:
I added `ggridges` to `Suggests`  inside `DESCRIPTION`  for the new age-to-length conversion plot, but it looks like that increased the install time enough to trigger the CI check. I was able to make a similar violin plot using only ggplot2 instead. I included that version here. Let me know which approach you think would be best.

Screenshot 2026-09-11 143213

iantaylor-NOAA (approved):
@awilnoaa, all the changes look good to me.
Regarding the plot type, I like the steps which help the user see the discrete length bins, but for purposes of this vignette, the violin plot is fine and better if it avoids slowing down the integration tests.

kellijohnson-NOAA:
Thank you @awilnoaa for these, I personally love ggridges and think it would be useful for other figures so I think we should put it into suggests as you first did. What do you think about making a function for this plot rather than providing the code in the vignette? I think that we can develop figures in FIMS and if they prove to be universal we can move them to stockplotr.

awilnoaa:
That makes sense to me. @iantaylor-NOAA already had a version of this as plot_age_to_length_conversion() in the case-studies repo, and I had kept it inline here because we were not sure yet where these helper plots should live.

If the preferred version is the ggridges version, should I move the plotting code into its own file, similar to R/plot_likelihood.R, for example R/plot_age_to_length_conversion.R?

kellijohnson-NOAA:
Yes, having the code in its own file would be great.

vignettes/fims-growth-vonb-schnute.Rmd
```diff
@@ -695,39 +695,32 @@ ggplot2::ggplot(residual_plot_data, ggplot2::aes(x = age, y = residual)) +
...
-
-# Pick a representative age/year slice so we can show one realized age-to-length conversion row.
-# `ceiling(length(demo$ages) / 2)` picks the middle modeled age, rounding up
-# when there are an even number of ages.
-representative_age <- demo$ages[ceiling(length(demo$ages) / 2)]
-representative_year <- demo$years[1]
+final_year <- max(demo$years)
 
 # Pull the full realized Growth-derived age-to-length conversion tensor from the report.
 realized_age_to_length_conversion <- report[["age_to_length_conversion_derived"]][[1]]
```
kellijohnson-NOAA: Why isn't this information available in the json reporting instead of in TMB report?

awilnoaa: I initially only added this to the TMB report because it can sometimes be a large derived output, built from the final Growth parameters and fleet-specific length-bin layout. At the time, I was treating it mostly as a debugging/diagnostic object rather than a standard user-facing output.

After working with Ian, it does seem like an output that users may want to access more readily for plotting etc. Let me know if you think I should add it to the JSON report in this PR or handle it in a separate follow-up PR.

iantaylor-NOAA: I have no thoughts on which PR should include this addition to the JSON report, but here are a few comments on why it's helpful to have the age-to-length conversion in the standard output:
1. even if you can calculate the growth curve and the variability around it from the estimated parameters, I think it's really helpful to see how it get allocated to the length bin structure
2. there are additional diagnostic plots which I think would be helpful to add to FIMS/stockplotr in the future which depend on having access to that conversion matrix, such as these two long-standing issues from r4ss: https://github.com/r4ss/r4ss/issues/19, https://github.com/r4ss/r4ss/issues/274

awilnoaa: Sounds good! I can add it in

awilnoaa:
I also noticed that Codecov reports no direct coverage for the new plotting helper. I don't see a targeted test for the existing plot_likelihood() helper, but I can add a small focused test for plot_age_to_length_conversion(). Let me know your thoughts

vignettes/fims-growth-vonb-schnute.Rmd
kellijohnson-NOAA: And, having it in the json allows users to have access to the information with all of the metadata rather than having to personally attach it as you do now in the vignette. I think it should be part of this PR.

kellijohnson-NOAA:
I have some local changes, where I removed the year input and faceted the plot by "timing" if users pass information with more than one "year". I also modified the code to allow for plotting input data as well as model output, i.e., the ability to pass a FIMSFrame and plot the age_to_length conversion input information. Let me know if you want me to push these changes or just provide more detailed comments on what I did. I think having the ability to also plot the input data will allow for tests on the input data rather than tests after running a model. We can also mock up model output and test that.

awilnoaa:
> I have some local changes, where I removed the year input and faceted the plot by "timing" if users pass information with more than one "year". I also modified the code to allow for plotting input data as well as model output, i.e., the ability to pass a FIMSFrame and plot the age_to_length conversion input information. Let me know if you want me to push these changes or just provide more detailed comments on what I did. I think having the ability to also plot the input data will allow for tests on the input data rather than tests after running a model. We can also mock up model output and test that.

Yes, please push your changes. I’ve already started working on the JSON/tidy output changes, so I can look at everything together once your changes are up. Thank You!

awilnoaa:
Thanks @kellijohnson-NOAA , I looked over your changes and I think I understand the direction you are headed. I just added the realized age-to-length conversion tensor to the JSON/tidy output, but I want to make sure I handle the coordinate metadata the way you intended. Right now my question is:

Should the JSON/tidy output include the actual `timing`, `age`, and fleet length-bin values directly, or should the plotting helper attach those from the original input/model object?

Thanks!

kellijohnson-NOAA:
The json parser takes information from dimensionality to add year, age, and length in long format, e.g., AgeComp
```
    ss << " \"dimensionality\": {\n";
    ss << "  \"header\": [" << "\"n_ages\", \"n_years\"" << "],\n";
    ss << "  \"dimensions\": [" << amax << ", " << ymax << "]\n},\n";
```
@msupernaw is working on a branch that adds real information to the json regarding year, age, and length rather than year_i, age_i, length_i that we have now. That is, information only stores the index of year not the actual year value. But, regardless of Matthew's edits or not, you should not have to put the vector of years, lengths, and ages, into the json, you just need to specify the dimensions the way that other modules do and then they will be parsed out when the json is converted to a long data frame.

